PR #24, opened 2023-10-05
Fix failure to create pull requests in pull-subtrees

pietroalbini:
When no previous commit created by the automation was found, and the subtree pull didn't result in any change, the automation would fail as it'd try to create a GitHub PR without any diff. This changes the automation to detect whether a subtree pull was actually committed.


tshepang (approved):
bors merge

PR #32, opened 2023-10-06
add missing option to config.example.toml

tshepang:


config.example.toml
```diff
@@ -875,3 +875,6 @@ changelog-seen = 2
 
 # Name of the AWS S3 bucket containing the document signatures file.
 #document-signatures-s3-bucket = "ferrocene-document-signatures"
+
+# Avoid "Unsigned draft" warning in Qualification docs, when they are not signed
```
pietroalbini, line 879: This also avoids erroring out if signatures are present, and you don't have access to the private signature files (stored in the S3 bucket).

Dajamante (approved):
bors merge

PR #40, opened 2023-10-09
x86_64 is also a tested target

tshepang:


ferrocene/doc/qualification-report/src/tests.rst
```diff
@@ -10,18 +10,21 @@ Ferrocene for ARMv8-A bare metal and x86-64 Linux
...
-For this qualification, testing is restricted to the following environment:
+For this qualification, testing is restricted to the following environments:
 
 .. list-table:: 
    :align: left
    :stub-columns: 1
 
    * - Host
      - x86_64 Linux
+     - x86_64 Linux
```
Dajamante, line 21: Is it a matrix? Why the repetition?

Dajamante: rst syntax!

PR #53, opened 2023-10-17
Automate dependency updates with dependabot

pietroalbini:
This PR configures dependabot to update the dependencies in the monorepo automatically, both for the majority of crates and for bootstrap. This is configured to open a weekly PR on Monday morning, and to group all updates in a single PR.


Veykril (approved):

Veykril:
bors merge

PR #59, opened 2023-10-18
Improve the workflow for upstream pulls

pietroalbini:
This PR improves the workflow for upstream pulls, in the hope of reducing the work for the person in the PR rotation, and to make review of upstream pulls easier.

With this change, the automation will open PRs even when there are merge conflicts, committing the conflict and the relevant conflict markers. It will be up to the person on the PR rotation to then fetch the branch, make a new commit fixing the conflict, and pushing it. Being a separate commit, reviewing the change will be easier, and the person on PR rotation won't have to generate the PR description anymore.

This also includes other miscellaneous changes:

* The support for different merge strategies has been removed: it was never used, and the need for it is reduced even further by this PR.
* There was a weird bug with the `Cargo.lock` autofix that I managed to fix. See its commit message for a more detailed description of the problem.


pietroalbini:
Good point, I'll investigate how deleted files are handled.

pietroalbini:
Changed the script to emit custom markers for deleted files (as there doesn't seem to be any builtin conflict markers for deleted files), and changed the script to look for them. Also, I changed the behavior of the script when run locally to create commits with conflicts in them by default, and using an environment variable to opt out.

japaric (approved):

pietroalbini:
bors merge

PR #68, opened 2023-10-26
Fix inconsistent behavior from the release scheduler

pietroalbini:
I noticed a corner case in the release scheduler that could lead to unpredictable commits being releases. More details are available in the comments.


skade (approved):

pietroalbini:
bors merge

PR #92, opened 2023-11-07
remove backlinks from automated PRs

pietroalbini:
Before this PR, every pull request mentioned in an automated pull from upstream or a subtree would include a link to the pulled PR in the form of `repo#number`. GitHub then used that to include a link to the PR, and automatically insert the PR title.

Unfortunately, doing so also resulted in a "backlink" event being included in the target PR. That is indeed annoying, and caused a fair amount of spam in `rust-lang/rust`. We asked GitHub Support, but there isn't an officially supported way to prevent those backlinks from being generated.

To solve the problem, this PR implements custom code to generate the links in automated PRs, crucially using `www.github.com` rather than `github.com` as the domain name. It turns out that adding www to the domain prevents backlinks from being created, which is perfect for us. This updates both pulls from upstream, and pulls from subtrees. Example rendered version:

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/ferrocene/tools/common/pr_links.py
+++ b/ferrocene/tools/common/pr_links.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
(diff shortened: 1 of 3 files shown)


skade (approved):
LGTM

pietroalbini:
bors merge

PR #95, opened 2023-11-08
Prevent `ferrocene/ci/configure.sh` from being invoked locally

pietroalbini:
Using the configure script to setup a local development environment is usually not a good idea, as that is meant to be used as part of our CI, and some of its settings hamper local development (for example, it rebuilds everything as soon as a new git commit is made).

To guide developers in not using the script, running it outside CI will now error out, pointing people to the internal procedures doc. There is an escape hatch if someone really needs it, which is pointed out in the error message.

This also updates the suggested `config.toml` with the new change tracking introduced upstream.

### DIFF
--- a/ferrocene/ci/configure.sh
+++ b/ferrocene/ci/configure.sh
@@ -16,6 +16,20 @@ add() {
     done
 }
 
+if [[ -z "${CI+x}" ]]; then
+    echo "error: this doesn't look like it's being executed in CI."
+    echo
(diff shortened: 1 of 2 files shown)


jonathanpallant (approved):

jonathanpallant:
Is it possible to add a link to the "Internal Procedures"?

pietroalbini:
Added the link.

jonathanpallant (approved):

jonathanpallant:
bonza

PR #99, opened 2023-11-09
Add infra to include the technical report in the built docs

pietroalbini:
This PR adds the infrastructure needed to include the technical report from TÜV into the generated documentation, by setting a key in `config.toml` pointing to the file in S3. Right now that is not configured in CI, as it's supposed to be used only on stable releases.

Two changes to the rest of Ferrocene were needed to make this work:

* There is now a way to hide some elements from the index based on the subsets the customer has access to. This is accomplished by hiding all elements with a `data-subset` HTML attribute by default, and having per-subset CSS files that show the subset's elements.
* There is now a way to include individual files in a subset by creating a file called `$name.ferrocene-subset`. The file behaves like a `ferrocene-subset` file inside a directory, but it only applies to that file rather than the whole directory.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -886,3 +886,7 @@ change-id = 117435
 # - Avoids an error that occurs if signatures are present,
 #   but you don't have access to the private signature files
 #ignore-document-signatures = false
+
+# Download URL for the qualification technical report. If omitted, the
+# technical report will not be included in the documentation.
+#technical-report-url = <none> (url)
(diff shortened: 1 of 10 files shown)


config.example.toml
```diff
@@ -886,3 +886,7 @@ change-id = 117435
 # - Avoids an error that occurs if signatures are present,
 #   but you don't have access to the private signature files
 #ignore-document-signatures = false
+
+# Download URL for the qualification technical report. If omitted, the
+# technical rpeort will not be included in the documentation.
```
tshepang, line 891:
```suggestion
# technical report will not be included in the documentation.
```

tshepang (approved):
bors merge

PR #100, opened 2023-11-09
Draft: Use LLD instead of LD for our three targets

jonathanpallant:
1) Switch x86_64-linux-gnu to use lld.

   Using the bundled `rust-lld` as the linker (via its `ld.lld` wrapper) means all users always use the same linker. This helps with the qualification efforts, rather than users using random versions of `ld`.

   However, we still drive `rust-lld` (or actually `ld.lld`) via the system C compiler, because the C compiler knows where the system libraries live, and Rust does not.

2) Switch the aarch64-unknown-none target to use rust-lld.

   Using the bundled `rust-lld` as the linker (via its `ld.lld` wrapper) means all users always use the same linker. This helps with the qualification efforts, rather than users using random versions of `ld`.

   We don't go via the system C compiler on `aarch64-unknown-none` because we do not need system C libraries. We do go via the system C compiler on `aarch64-unknown-ferrocenecoretest` because that needs `libstd` to build the test runner.

### DIFF
--- a/compiler/rustc_target/src/spec/mod.rs
+++ b/compiler/rustc_target/src/spec/mod.rs
@@ -589,6 +589,12 @@ impl LinkSelfContainedDefault {
             _ => "crt-objects-fallback",
         }
     }
+
+    /// Creates a `LinkSelfContained` enabling the self-contained linker for target specs (the
+    /// equivalent of `-Clink-self-contained=+linker` on the CLI).
+    pub fn with_linker() -> LinkSelfContainedDefault {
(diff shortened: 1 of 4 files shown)


pietroalbini:
Changes look good, but GitHub is reporting conflicts? I'm quite puzzled.

jonathanpallant:
I just rebased on main and they went away.

tshepang:
> Changes look good, but GitHub is reporting conflicts? I'm quite puzzled.

may due to [moved files](https://github.com/rust-lang/rust/pull/117702)

pietroalbini:
Good catch!

PR #112, opened 2023-11-13
Automatically update Cargo.lock during upstream pulls

pietroalbini:
In https://github.com/ferrocene/ferrocene/pull/109 we noticed another way automated pulls could fail: the `Cargo.lock` requiring to be regenerated due to the custom Ferrocene tools we have in the repository. This PR prevents these problems from showing up in the future by regenerating the lockfile every single pull, if needed.


tshepang (approved):
bors merge

pietroalbini:
Turns out that I forgot `--manifest-path` was a thing. Rather than using an unstable feature (`-C`) I pushed a commit to switch to it. `RUSTC_BOOTSTRAP=1` is still needed as the rustc `Cargo.toml` requires unstable features to be parsed.

tshepang (approved):
bors merge

PR #115, opened 2023-11-13
Fix PR message not being generated during some upstream pulls

pietroalbini:
There was a long standing bug in the script used to generate upstream pulls' PR bodies: if followup commits were present on the branch, the script would not output anything. Up until today this only happened if the script was ran locally after committing extra changes, but since #112 the automation itself can generate followup commits.

This PR fixes the bug by walking back the commit graph until a merge commit is found, rather than the current behavior of always using the last commit even if it's not a merge commit.

This PR also speeds up the execution of the automation in CI, as most of the time right now is spent cloning LLVM. The CI script to download a shallow clone of a subset of it is now used instead.

### DIFF
--- a/ferrocene/tools/pull-upstream/generate_pr_body.py
+++ b/ferrocene/tools/pull-upstream/generate_pr_body.py
@@ -84,6 +84,18 @@ def render_changes(origin, base_branch, new_branch):
     if last is None:
         return "**Nothing**"
 
+    # When merge conflicts are fixes with separate commits, the latest commit
+    # on the branch will not be the merge commit from upstream. In those cases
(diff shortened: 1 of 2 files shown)


ferrocene/tools/pull-upstream/generate_pr_body.py
```diff
@@ -84,6 +84,18 @@ def render_changes(origin, base_branch, new_branch):
     if last is None:
         return "**Nothing**"
 
+    # When merge conflicts are fixes with separate commits, the latest commit
```
tshepang, line 87:
```suggestion
    # When merge conflicts are fixed with separate commits, the latest commit
```

PR #116, opened 2023-11-13
Teaching ferrocene-self-test how to check the C compiler

jonathanpallant:
* Only some targets now need a C compiler, the others use LLD directly
* We need to know what the C compiler is called (`cc`, `gcc`, `clang`, etc)
* We need to check the C compiler supports `-fuse-ld=lld`
* We need to check the C compiler supports `-B/path/to/tools`
* We check if the linker is being given LTO related arguments by the C compiler and, if so, turn that off.
* We need to tell the user what linker args we used to build their binary

### DIFF
--- a/ferrocene/tools/self-test/src/linkers.rs
+++ b/ferrocene/tools/self-test/src/linkers.rs
@@ -1,71 +1,225 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: The Ferrocene Developers
 
-use crate::error::Error;
-use crate::report::Reporter;
-use crate::targets::Target;
-use crate::utils::{find_binary_in_path, run_command, DisplayList};
-use crate::Environment;
-use std::collections::BTreeMap;
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub(crate) enum Linker {
+use crate::error::Error;
+use crate::report::Reporter;
+use crate::targets::Target;
+use crate::utils::{find_binary_in_path, run_command};
+use crate::Environment;
+
+/// What kind of C compiler does a target require
+#[derive(Debug)]
+pub enum Linker {
+    /// No C compiler required
     BundledLld,
-    GccUbuntu18 { target: &'static str, mode: GccMode },
+    /// The system's native C compiler is required
+    HostCC,
+    /// Some kind of cross compiler, with one of the given target prefixes
+    CrossCC(&'static [&'static str]),
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub(crate) enum GccMode {
-    Normal,
-    #[cfg_attr(not(test), allow(unused))]
-    BareMetal,
-}
-
+/// We need to find a system C compiler and determine what flags should be
+/// added when calling `rustc`.
+///
+/// These flags will include `-Clinker=/path/to/cc`.
 pub(crate) fn check_and_add_rustflags(
     reporter: &dyn Reporter,
     environment: &Environment,
     sysroot: &Path,
     targets: &mut [Target],
 ) -> Result<(), Error> {
-    // Multiple installed targets might use the same linker, deduplicate them.
-    let mut linkers = BTreeMap::new();
-    for target in targets {
-        linkers.entry(target.linker).or_insert_with(Vec::new).push(target);
-    }
+    // Step 1. Check we have ld.lld available
+    let _rust_lld_path = find_bundled_lld(reporter, sysroot)?;
+    let lld_bin = find_bundled_lld_wrapper(reporter, sysroot)?;
+    let lld_dir = lld_bin.parent().expect("ld.lld should be a in a directory");
 
-    for (linker, targets) in linkers {
-        let triples = targets.iter().map(|t| t.triple.to_string()).collect::<Vec<_>>();
-        let (bin, flavor, extra_flags): (_, _, &[&str]) = match linker {
-            Linker::BundledLld => (check_bundled_lld(reporter, sysroot, &triples)?, "ld.lld", &[]),
-            Linker::GccUbuntu18 { target: gcc_target, mode } => (
-                check_gcc(reporter, environment, &triples, gcc_target, mode, [7, 5])?,
-                "gcc",
-                match mode {
-                    GccMode::Normal => &[],
-                    GccMode::BareMetal => &["-ffreestanding", "-nostdlib"],
-                },
-            ),
+    'target_loop: for target in targets {
+        let prefix_list: &[&str] = match target.linker {
+            Linker::BundledLld => {
+                continue 'target_loop;
+            }
+            Linker::HostCC => &[""],
+            Linker::CrossCC(list) => list,
         };
-        let bin = bin.to_str().ok_or_else(|| Error::NonUtf8Path { path: bin.clone() })?;
-        for target in targets {
-            target.rustflags.push(format!("-Clinker={bin}"));
-            target.rustflags.push(format!("-Clinker-flavor={flavor}"));
-            for flag in extra_flags {
-                target.rustflags.push(format!("-Clink-arg={flag}"));
+        reporter.note(&format!(
+            "Target {:?} requires a C compiler. Searching for one...",
+            target.triple
+        ));
+        for cc_prefix in prefix_list {
+            for compiler_kind in ["cc", "gcc", "clang"] {
+                let temp_dir = tempfile::tempdir().expect("making temp dir");
+                let compiler_name = format!("{cc_prefix}{compiler_kind}");
+                let cc_result = check_system_compiler(
+                    reporter,
+                    environment,
+                    &compiler_name,
+                    lld_dir,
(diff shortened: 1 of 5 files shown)


pietroalbini:
From a first pass the code seems good, but I have a few pieces of feedback on the functionality:

* We should make the log more concise: the only needed log lines is that we did find the C compiler "foo" for target "bar", or an error that we couldn't find any C compiler for target "bar". There shouldn't be a need to print every compiler we're attempting, or the flags it passes to LLD.

* Regarding flags, it would be nice to have more automation and stricter checks about the flags passed to LLD. Right now the only check is denying the LTO plugin flag, while we should move to a model where we error out if any of the flags is not explicitly allowed. Then we can document in the safety manual that the C compiler needs to directly invoke the linker and only provide that set of flags. This would also catch the LTO case, even though in that case it's probably worth it to special-case it (adding `-fno-lto` to the list of CLI flags and re-running the check).

The second point is going to require a fair amount of code (we at least need to lex the CLI flags to split flags from flag values), so it's fine to do it in a followup PR and keep the current behavior as-is for this PR.

jonathanpallant:
```text
    Info: using sysroot /home/ubuntu/ferrocene
 Skipped: optional binary cargo (not present)
 Success: target installed correctly: x86_64-unknown-linux-gnu
 Success: target installed correctly: aarch64-unknown-linux-gnu
 Success: target installed correctly: aarch64-unknown-none
 Success: bundled linker detected
 Success: bundled linker-wrapper detected
    Info: C compiler cc detected, testing supported arguments
 Success: Found C compiler `cc` for target `x86_64-unknown-linux-gnu`
    Info: C compiler aarch64-linux-gnu-gcc detected, testing supported arguments
 Success: Found C compiler `aarch64-linux-gnu-gcc` for target `aarch64-unknown-linux-gnu`
 Skipped: Target `aarch64-unknown-none` does not require a C compiler
 Success: compiled sample program `addition.rs` for target x86_64-unknown-linux-gnu
 Success: compiled sample program `subtraction.rs` for target x86_64-unknown-linux-gnu
 Success: compiled sample program `subtraction-sys.rs` for target x86_64-unknown-linux-gnu
 Success: compiled sample program `assertion.rs` for target x86_64-unknown-linux-gnu
 Success: compiled sample program `addition.rs` for target aarch64-unknown-linux-gnu
 Success: compiled sample program `subtraction.rs` for target aarch64-unknown-linux-gnu
 Success: compiled sample program `subtraction-sys.rs` for target aarch64-unknown-linux-gnu
 Success: compiled sample program `assertion.rs` for target aarch64-unknown-linux-gnu
 Success: compiled sample program `addition.rs` for target aarch64-unknown-none
 Success: compiled sample program `subtraction.rs` for target aarch64-unknown-none
 Success: compiled sample program `subtraction-sys.rs` for target aarch64-unknown-none
 Success: compiled sample program `assertion.rs` for target aarch64-unknown-none
    Info: Target 'x86_64-unknown-linux-gnu' requires special linker flags:
    Info: 	-Clinker=cc
    Info: 	-Clink-arg=-fno-lto
    Info: Target 'aarch64-unknown-linux-gnu' requires special linker flags:
    Info: 	-Clinker=aarch64-linux-gnu-gcc
    Info: 	-Clink-arg=-fno-lto
    Info: Target 'aarch64-unknown-none' requires no special linker flags
 Success: Ferrocene self-check completed!
 ```

File: ferrocene/tools/self-test/src/linkers.rs [hunk 1]
```diff
@@ -1,71 +1,257 @@
...
         };
-        let bin = bin.to_str().ok_or_else(|| Error::NonUtf8Path { path: bin.clone() })?;
-        for target in targets {
-            target.rustflags.push(format!("-Clinker={bin}"));
-            target.rustflags.push(format!("-Clinker-flavor={flavor}"));
-            for flag in extra_flags {
-                target.rustflags.push(format!("-Clink-arg={flag}"));
+        for cc_prefix in prefix_list {
+            for compiler_kind in ["cc", "gcc", "clang"] {
+                let temp_dir = tempfile::tempdir().expect("making temp dir");
```
pietroalbini, line 53: There is an error variant for this already, `Error::TemporaryCompilationDirectoryCreationFailed`. We also create a temporary directory in `compile.rs`, it might be worth to extract that to `main()` and pass the path to the temporary directory to all functions requiring one?

ferrocene/tools/self-test/src/linkers.rs
```diff
@@ -1,71 +1,257 @@
...
+                let compiler_name = format!("{cc_prefix}{compiler_kind}");
+                let cc_result = check_system_compiler(
+                    environment,
+                    target.triple,
+                    &compiler_name,
+                    lld_dir,
+                    temp_dir.path(),
+                );
+                temp_dir.close().expect("clean up temp");
```
pietroalbini, line 62: Any reason why we're explicitly closing this?

ferrocene/tools/self-test/src/linkers.rs
```diff
@@ -1,71 +1,257 @@
...
+
+/// Check if the given system C compiler works.
+///
+/// We are given a path to the real `ld.lld`.
+///
+/// We are also given a path to a fresh temporary directory we can put source
+/// code in.
+///
+/// Returns the path to the C compiler.
+fn compile_c_program(cc_path: &Path, lld_dir: &Path, temp_dir: &Path) -> Result<(), Error> {
```
pietroalbini, line 142: I think this is redundant with `make_fake_linker`? That also compiles a C program, and we can pass `-fuse-ld`/`-B` to it as well.

ferrocene/tools/self-test/src/linkers.rs
```diff
@@ -1,71 +1,257 @@
...
+    let args_file = temp_dir.join("_fst_args_capture");
+
+    // Concatentation, using byte strings
+    let mut c_source = C_SOURCE.to_owned();
+    c_source.extend(args_file.as_os_str().as_encoded_bytes());
+    c_source.extend(C_SOURCE2);
+
+    let source_file = temp_dir.join("ldlld.c");
+    let object_file = temp_dir.join("ld.lld");
+    std::fs::write(&source_file, &c_source).expect("saving source code to temp file");
```
pietroalbini, line 204: There is the `Error::WritingSampleProgramFailed` for this case.

ferrocene/tools/self-test/src/linkers.rs
```diff
@@ -1,71 +1,257 @@
...
+///
+/// Returns the path to the C compiler.
+fn compile_c_program(cc_path: &Path, lld_dir: &Path, temp_dir: &Path) -> Result<(), Error> {
+    // We need some C source code,
+    let c_source = r#"int main(void) { return 0; }"#;
+
+    // We need a temp directory we can save the output file to
+    let source_file = temp_dir.join("input.c");
+    let object_file = temp_dir.join("output.bin");
+    std::fs::write(&source_file, c_source.as_bytes()).expect("saving source code to temp file");
```
pietroalbini, line 149: There is the `Error::WritingSampleProgramFailed` for this case.

ferrocene/tools/self-test/src/linkers.rs
```diff
@@ -1,71 +1,257 @@
...
+        lld_dir.as_os_str().to_owned(),
+        source_file.as_os_str().to_owned(),
+        "-o".into(),
+        object_file.as_os_str().to_owned(),
+    ];
+    let mut cc_child = Command::new(cc_path);
+    cc_child.args(&args);
+
+    let _output = run_command(&mut cc_child).map_err(|error| {
+        let cc_name = cc_path.file_name().unwrap().to_str().unwrap();
```
pietroalbini, line 164: There is `Error::NonUtf8Path` for this.

pietroalbini (commented):
Did the thorough review on this, and left some pieces of feedback. Also, it seems like there are no tests for the fake linker creation and gathering the linker args?

ferrocene/tools/self-test/src/linkers.rs
jonathanpallant: I wanted each compiler test to have its own temp directory, so we don't have to worry about output files lying around from the previous test.

But I will use the correct error code.

ferrocene/tools/self-test/src/linkers.rs
jonathanpallant: I actually thought the same thing, but then I realised `make_fake_linker` needs to make a host binary, but `compile_c_program` needs to do a cross-compilation. I should clarify this in the comments.

ferrocene/tools/self-test/src/linkers.rs
jonathanpallant: Returning a non-UTF-8 error here would hide the fact the actual problem was that "`run_command` failed". I'll unwrap_or in a replacement string for the unlikely event the C compiler path is not UTF-8.

ferrocene/tools/self-test/src/linkers.rs
pietroalbini: You're right, good catch.

ferrocene/tools/self-test/src/linkers.rs
jonathanpallant: no.

jonathanpallant:
OK, I cleaned that up a bit. Thank you for the detailed review.

pietroalbini (approved):
bors merge

PR #118, opened 2023-11-13
Actually run tests for `generate-tarball`

pietroalbini:
While preparing the open sourcing of Ferrocene 23.06, I stumbled upon ferrocene/ferrocene-prehistory#1454. As part of the fix, that PR accidentally switched the step executed when requesting the `generated-tarball` tests. This PR fixes that.


tshepang (approved):
bors merge

Issue #10, opened 2023-08-10
Switch the linker to LLD


pietroalbini:
Right now, all targets require the use of GCC 7.5.x as the linker, which is suboptimal for everyone not using Ubuntu 18.04 LTS. This was done because we need to test the exact linker version customers will use for qualification, and we had to pick one.

The solution we should move towards is using the LLD bundled with Ferrocene as the linker. This means updating the target specs, updating the self-testing tool to check for the right linker, and updating the target documentation.

For bare metal targets, this is as simple as changing the linker, as LLD can work with them by default (and that's actually upstream's behavior, we had to override it). Then the self-test tool can be updated to verify the presence of LLD (the code should already be there, just an enum change should be needed).

For Linux targets though we cannot invoke LLD directly, as LLD does not know where the system libraries are located. Thus, we need to use the system GCC or clang as the default linker, passing `-fuse-ld=lld` to it. That flag seems to work from GCC 9 onwards. This requires wider changes:

* Change the target spec to always use LLD.
* Change the self-testing tool to check for the presence of LLD, *and* the system linker (either gcc or clang).
* Add a new check in the self-test tool, that when compiling something, GCC or clang end up invoking LLD with the correct flags. This can be implemented by creating a fake LLD binary that checks the flags are correct: GCC/clang can then be invoked with some flags pointing to the fake LLD, which will error out if the flags are not passed through and the system flags are not present.
* In the safety manual, add a requirement for how GCC/clang should behave, and mention the self-test tool performs that check automatically.

jonathanpallant:
Completed in #100, however we still need to update the documentation (#106) and the self-test program (#107)

pietroalbini:
Let's track them separately, and close this issue.

jonathanpallant:
Also, I want to clarify here some changes to our thinking. Rather than try and produce a binary called `ld` (which runs the risk of clashing with the system provided `ld`), we instead decided to check if the system supplied C compiler supports `-fuse-ld=lld` and if not, check to see if a suitable `clang` was installed. Clang is available on Ubuntu 18.04, and also knows where the system libraries live, so works as a system C compiler for our purposes.

You can also verify which linker produced any given ELF file by running:

```console
$ readelf --string-dump .comment /path/to/elf
String dump of section '.comment':
  [     1]  rustc version 1.75.0-dev
  [    1a]  Linker: LLD 17.0.4
```

PR #133, opened 2023-11-20
Update target docs.

jonathanpallant:
Updates the docs for `aarch64-unknown-none` and `x86_64-unknown-linux-gnu` to reflect our new choices around which linker we use (see #10)

### DIFF
--- a/ferrocene/doc/user-manual/src/targets/x86_64-unknown-linux-gnu.rst
+++ b/ferrocene/doc/user-manual/src/targets/x86_64-unknown-linux-gnu.rst
@@ -12,7 +12,7 @@ or higher on x86_64 using glibc 2.17 or higher.
 Prerequisites
 -------------
 
-This target uses the LLVM ``ld.lld`` linker and is qualfied for use with the
+This target uses the LLVM ``ld.lld`` linker and is qualified for use with the
 version of ``ld.lld`` bundled with Ferrocene. To locate the system C libraries
 required to link a functional Linux binary, this target drives the ``ld.lld``
 linker using your system's C compiler as a linker-driver.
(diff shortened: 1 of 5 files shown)


jonathanpallant:
We discussed this, and

* most of these changes should be in the Safety Manual and not the target pages in the User Manual
* The support Linux platform details should match Ubuntu 18.04 as that's the build and test target
* The Safety Manual must call out the supported linker arguments specifically and cannot rely on users running `ferrocene-self-test`, which is not qualified

ferrocene/doc/user-manual/src/targets/x86_64-unknown-linux-gnu.rst
```diff
@@ -6,26 +6,28 @@
...
-On Ubuntu 18.04 LTS you can install it with:
+You must have a C compiler which:
 
-.. code-block::
-
-   $ sudo apt install gcc
+* Supports ``-fuse-ld=ld.lld`` option to select ``ld.lld`` as the linker
+* Supports ``-B`` option to modify the tool search path so it can find Ferrocene's
+  copy of ``ld.lld``
+* Does not load plugins into the linker (e.g. the GCC LTO plugin)
```
pietroalbini, line 24: I'd add a line "only passes the command line flags in [link to safety manual] to ld.lld".

pietroalbini (commented):
Looks great! Left a minor fix, but otherwise we can land this.

ferrocene/doc/user-manual/src/targets/x86_64-unknown-linux-gnu.rst
jonathanpallant: Added

pietroalbini (approved):
bors merge

PR #139, opened 2023-11-27
Limit how many PRs can be pulled in a single automated pull

pietroalbini:
Before this PR, the pull-upstream automation would include all pending upstream PRs into the automated pull PR, regardless of how many changes were merged upstream since the last automated pull. That behavior could result in huge PRs being created, which would make fixing conflicts in them needlessly painful.

This PR changes the code to allow at most 30 PRs to be merged in a single pull, requiring multiple pulls to be performed if there are more pending PRs to be pulled in. Note that this will often result in more than 30 PRs being merged, as rollup PRs are counted as one PR (even though a rollup itself contains more than one merge).


Dajamante (approved):
bors merge

PR #142, opened 2023-11-28
remove reference to non-existent repo

tshepang:
The subtree was removed, but the rustfmt instruction was not.


Dajamante (approved):
bors merge

PR #36, opened 2023-10-09
Test and fix CI for the 1.73 branch

pietroalbini:
I just squashed the 1.73 branch in `release/1.73`, this is the equivalent of https://github.com/ferrocene/ferrocene/pull/1 and https://github.com/ferrocene/ferrocene/pull/30 for it.

PR #147, opened 2023-11-30
Test and fix CI for the 1.68 branch (23.06)

pietroalbini:
This PR is the equivalent of #36 for the 1.68 branch, which contains the source code of Ferrocene 23.06.x.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -670,7 +670,7 @@ commands:
         default: false
     steps:
       - ferrocene-git-shallow-clone:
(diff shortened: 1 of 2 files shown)


pietroalbini:
Ok there was one last set of files that had to be deleted. This time I verified there are no extra files in this repo compared to the squashed one with:

```
diff -u <(git -C ferrocene-prehistory ls-files) <(git -C ferrocene ls-files)
```

tshepang (approved):
bors merge

PR #156, opened 2023-12-05
Ferrocene Self Test: Check C compiler passes extra args.

jonathanpallant:
We now give the C compiler a `-Wl,--foobar` argument, and test that the linker receives `--foobar` as an argument (but using a long random string instead of `foobar`).

### DIFF
--- a/ferrocene/tools/self-test/src/compile.rs
+++ b/ferrocene/tools/self-test/src/compile.rs
@@ -16,18 +16,21 @@ static SAMPLE_PROGRAMS: &[SampleProgram] = &[
         contents: include_bytes!("../sample-programs/addition.rs"),
         rustflags: &["--crate-type", "lib", "--edition", "2021"],
         expected_artifacts: &["libaddition.rlib"],
+        executable_output: None,
     },
     SampleProgram {
         name: "subtraction.rs",
         contents: include_bytes!("../sample-programs/subtraction.rs"),
         rustflags: &["--crate-type", "staticlib", "--edition", "2021"],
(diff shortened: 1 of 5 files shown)


jonathanpallant:
Now we execute sample programs (if they are executables, and we're on the host target) so now we check the object files are linked correctly.

jonathanpallant:
bors try

ferrocene/tools/self-test/src/compile.rs
```diff
@@ -119,12 +135,35 @@ fn compile(ctx: &Context<'_>, program: &SampleProgram) -> Result<(), Error> {
...
+                name: program.name.into(),
+                error,
+            })?;
+        if output.stdout != expected_output {
+            // TODO change for SampleProgramOutputWrong
+            return Err(Error::SampleProgramOutputWrong {
+                name: program.name.into(),
+                expected: expected_output.to_vec(),
+                found: output.stdout,
+            });
```
pietroalbini, line 163: Why the TODOs here?

pietroalbini (commented):
Code changes look good, only left a small comment on some leftover TODOs. Also, we should add a requirement for this in the safety manual (https://public-docs.ferrocene.dev/main/safety-manual/options.html#linker-options)

ferrocene/tools/self-test/src/compile.rs
jonathanpallant: ah, because I fixed the TODO and didn't delete the note I'd left myself

jonathanpallant:
Added a para to the safety manual about what the C compiler must do.

PR #180, opened 2024-01-02
Code Coverage Support for Core Library

Ax9DTW:
This PR adds code coverage support for the `core` library using rustc's existing support for instrumentation using the [`-Cinstrument-coverage`](https://doc.rust-lang.org/rustc/instrument-coverage.html) flag.

## Usage 

To generate the coverage data run the command:
`./x.py test --coverage library/core --no-doc`

1. Currently coverage doesn't work with doc tests, hence the `--no-doc` flag has to be passed.
2. The `.profraw` files are created in the `build/coverage` directory.

## The Problem
When the `-Cinstrument-coverage` flag is set during compilation, llvm's profiler runtime is injected into the binary. This runtime is provided by the `profiler_builtins` crate and this crate is linked by making it an implicit dependency of the current crate being compiled.

This creates a problem when we want to instrument the `core` library, as the `profiler_builtins` crate has a dependency on core and hence can't be built it, creating a cyclic dependency.

> Note: `profiler_builtins` doesn't need to be dependent on core, the dependency is there to define a specific build order.

## Proposed Solution

Our approach is to make the dependency on `core` library optional in `profiler_builtins`. This allows to build and use a different version of the `profiler_builtins` crate, without the `core` dependency when instrumenting the `core` library. 

## Notes

1. Removing `core` dependency in `profiler_builtins` has certain limitations as discussed in [this PR](https://github.com/rust-lang/rust/pull/101009).

2. Another approach to resolve the cylic dependency is to remove the `profiler_builtins` crate entirely and build the profiler runtime as a separate build step similar to sanitizers. Though this works it has certain issues as discussed in [this PR](https://github.com/rust-lang/rust/pull/118305).

## Implementation Details

1. The dependency on core is made optional and is enabled by default by setting a feature flag called `core` in `profiler_builtins`.

2. In bootstrap, a custom step has been added to build `profiler_builtins` separately under a different name with `--no-default-features` and hence without a dependency on `core`.

4.  When gathering code coverage  we link against this custom built `profiler_builtins`.

@pietroalbini @tw-amit @gnana-ganesh-tw

### DIFF
--- a/compiler/rustc_metadata/src/creader.rs
+++ b/compiler/rustc_metadata/src/creader.rs
@@ -779,7 +779,7 @@ impl<'a, 'tcx> CrateLoader<'a, 'tcx> {
         self.inject_dependency_if(cnum, "a panic runtime", &|data| data.needs_panic_runtime());
     }
 
-    fn inject_profiler_runtime(&mut self, krate: &ast::Crate) {
+    fn inject_profiler_runtime(&mut self) {
         if self.sess.opts.unstable_opts.no_profiler_runtime
             || !(self.sess.instrument_coverage()
                 || self.sess.opts.unstable_opts.profile
@@ -791,9 +791,9 @@ impl<'a, 'tcx> CrateLoader<'a, 'tcx> {
         info!("loading profiler");
 
         let name = Symbol::intern(&self.sess.opts.unstable_opts.profiler_runtime);
-        if name == sym::profiler_builtins && attr::contains_name(&krate.attrs, sym::no_core) {
-            self.sess.emit_err(errors::ProfilerBuiltinsNeedsCore);
-        }
+        // if name == sym::profiler_builtins && attr::contains_name(&krate.attrs, sym::no_core) {
+        //     self.sess.emit_err(errors::ProfilerBuiltinsNeedsCore);
+        // }
 
         let Some(cnum) = self.resolve_crate(name, DUMMY_SP, CrateDepKind::Implicit) else {
             return;
@@ -991,7 +991,7 @@ impl<'a, 'tcx> CrateLoader<'a, 'tcx> {
 
     pub fn postprocess(&mut self, krate: &ast::Crate) {
         self.inject_forced_externs();
-        self.inject_profiler_runtime(krate);
+        self.inject_profiler_runtime();
         self.inject_allocator_crate(krate);
         self.inject_panic_runtime(krate);
 
--- a/compiler/rustc_metadata/src/creader.rs
+++ b/compiler/rustc_metadata/src/creader.rs
@@ -791,9 +791,6 @@ impl<'a, 'tcx> CrateLoader<'a, 'tcx> {
         info!("loading profiler");
 
         let name = Symbol::intern(&self.sess.opts.unstable_opts.profiler_runtime);
-        // if name == sym::profiler_builtins && attr::contains_name(&krate.attrs, sym::no_core) {
-        //     self.sess.emit_err(errors::ProfilerBuiltinsNeedsCore);
-        // }
 
         let Some(cnum) = self.resolve_crate(name, DUMMY_SP, CrateDepKind::Implicit) else {
             return;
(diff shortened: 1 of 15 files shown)


compiler/rustc_metadata/src/creader.rs
```diff
@@ -791,9 +791,6 @@ impl<'a, 'tcx> CrateLoader<'a, 'tcx> {
         info!("loading profiler");
 
         let name = Symbol::intern(&self.sess.opts.unstable_opts.profiler_runtime);
-        if name == sym::profiler_builtins && attr::contains_name(&krate.attrs, sym::no_core) {
-            self.sess.emit_err(errors::ProfilerBuiltinsNeedsCore);
-        }
```
pietroalbini, line 796: I don't think we can remove this check.

I did some spelunking on why the check was introduced, [and this comment explains why](https://github.com/rust-lang/rust/pull/79958#issuecomment-743958675): when `profiler_builtins` is injected, it brings its dependencies into the build, which include the `core` crate. Now, if you link that into a crate that defines its own lang items, like a `#![no_core]` crate or `core` itself (like we're trying to do), you will see the duplicate lang item errors we've been seeing.

It's correct that this error doesn't apply to us anymore, since the `profiler_builtins` we're injecting doesn't depend on core, but users passing `-C instrument-coverage` on `#![no_core]` crates still need to see this error.

src/bootstrap/src/core/config/config.rs
```diff
@@ -347,6 +347,7 @@ pub struct Config {
     pub ferrocene_document_signatures_s3_bucket: String,
     pub ferrocene_ignore_document_signatures: bool,
     pub ferrocene_technical_report_url: Option<String>,
+    pub ferrocene_code_coverage: bool,
```
pietroalbini, line 350: In the rest of the bootstrap codebase, the value of command line flags is not added to the `Config` struct, but as a method of the `Subcommand` struct (in `flags.rs`). See for example `Subcommand::bless()`.

pietroalbini (changes requested):
Thanks for all the work on this! Left a few small comments on how the bootstrap codebase is usually structured, plus a major comment on a removed check.

src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -2547,6 +2569,19 @@ impl Step for Crate {
             target,
             builder,
         );
+
+        if collect_profraw {
```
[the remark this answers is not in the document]
Ax9DTW: My rationale was that if the coverage command somehow fails, the old coverage data would be lost, that's why we write to temp and then rename. 

That being said std::fs::rename still can't be used to do an atomic move replace.

compiler/rustc_metadata/src/creader.rs
Ax9DTW: What if we set a different name like `-Zprofiler-runtime=profiler_runtime_no_core` in the build step? This way the check gets bypassed because the name doesn't match. This would also not affect the regular code path, when the user passes `-Cinstrument-coverage`.

gnana-ganesh-tw:
> Thanks for all the work on this! Left a few small comments on how the bootstrap codebase is usually structured, plus a major comment on a removed check.

We have updated the PR as per the review comments.

compiler/rustc_metadata/src/creader.rs
pietroalbini: That works too!

src/bootstrap/src/core/build_steps/test.rs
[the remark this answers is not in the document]
pietroalbini: Makes sense. Maybe add an inline comment explaining this?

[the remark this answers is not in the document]
Ax9DTW: Sure, I've left a comment

pietroalbini (approved):
Thanks!

bors merge

pietroalbini:
@gnana-ganesh-tw could you force-push the branch again / push an empty commit (`git commit --allow-empty -m "empty commit"`)? CI was not executed on the PR as it was a draft, and without the PR CI passing bors will not enqueue the PR for merging.

Ax9DTW:
Did a force push

PR #189, opened 2024-01-08
Fix conflict in bootstrap's `Cargo.lock` not being resolved

pietroalbini:
One of the goals of the pull-upstream automation is to automatically resolve `Cargo.lock` conflicts. This is done in two steps:

1. If there is a merge conflict in the lockfile, we automatically resolve that by reverting to our version and then invoking Cargo to update the lockfile. This is committed as part of the merge commit.
2. We then attempt to update the lockfile with Cargo and commit it separately if that results in changes, to handle the case where the merge produced a lockfile with no conflict but that would fail `cargo build --locked`.

In theory this should avoid any lockfile conflict from ever showing up. Unfortunately, before this PR step 1 was done only for `Cargo.lock`, while step 2 was done for both `Cargo.lock` and `src/bootstrap/Cargo.lock`. This means that if a merge conflict happened in `src/bootstrap/Cargo.lock`, we'd attempt to invoke Cargo with a lockfile containing merge conflict markers.

This PR ensures step 1 is executed on both lockfiles, and creates a single list of lockfiles to update so that the two steps always act on the same lockfiles.

### DIFF
--- a/ferrocene/tools/pull-upstream/pull.sh
+++ b/ferrocene/tools/pull-upstream/pull.sh
@@ -8,6 +8,7 @@ IFS=$'\n\t'
 UPSTREAM_REPO="https://github.com/rust-lang/rust"
 TEMP_BRANCH="pull-upstream-temp--do-not-use-for-real-code"
 MAX_MERGES_PER_PR=30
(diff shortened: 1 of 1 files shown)


tshepang (approved):
bors merge

PR #212, opened 2024-01-22
Generate the internal `ferrocene-docs-doctrees` tarball

pietroalbini:
This PR adds a new dist step that generates an *internal* `ferrocene-docs-doctrees` tarball, which contains the XML serialization of Sphinx's AST (the docutils doctrees). There is no use for the tarball right now, but a thing I'd like to build in the near future is a tool to generate a semantic diff of the documentation between two versions (so that we can easily give a changelog to TÜV), and that requires having the doctrees of old versions.

The tarball is internal not because it contains sensitive info (it's a different representation of the same info after all), but because showing it in the download pages to users would be super confusing.

This also changes some Sphinx extensions to output their files only when building HTML.


japaric (approved):
bors merge

PR #213, opened 2024-01-22
add test for Cortex-A53 errata 843419 mitigation

### LINKED ISSUE #125: Investigate `-fix-cortex-a53-843419`
In #123 we discovered that GCC passes the `-fix-cortex-a53-843419` flag to linkers to address an errata in Cortex-A53 processors. This means that by default GCC seems to pass that, but we need to do some work around it:

```[tasklist]
### Tasks
- [ ] Check whether bare LLD also implements the mitigation, or if we need to change the target spec or  mandate users pass `-C link-args` in the safety manual.
- [ ] Include in the safety manual that the mitigation must be enabled.
- [ ] Add a test to ensure the mitigation works.
- [ ] Include detection in FST to require the flag to be passed if CC doesn't include it.
```

japaric:
closes #125

### DIFF
--- a/tests/run-make/fix-cortex-a53-843419/Makefile
+++ b/tests/run-make/fix-cortex-a53-843419/Makefile
@@ -1,0 +1,17 @@
+# only-aarch64-unknown-none
+
+# regression test for rust-lang/rust#118095
+
+# run-make sets `cc` as the linker; override that default
+RUSTC_LINKER=rust-lld
+include ../tools.mk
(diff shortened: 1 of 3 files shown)


tests/run-make/fix-cortex-a53-843419/Makefile
```diff
@@ -0,0 +1,17 @@
+# only-aarch64-unknown-none
```
pietroalbini, line 1: Should we run this on all aarch64 platforms?

japaric:
bors try

pietroalbini (approved):
bors merge

PR #215, opened 2024-01-23
Enable the new `rust.use-lld=self-contained`

pietroalbini:
This flag will instruct bootstrap to use our own LLD when linking rustc itself, rather than using the system linker. Introduced in https://github.com/rust-lang/rust/pull/116278.


pietroalbini:
bors try

pietroalbini:
bors try

PR #219, opened 2024-01-24
Add cross-linking between the safety manual and the evaluation report

pietroalbini:
The "Constraints" page of the safety manual needs to refer to requirements defined in the evaluation report. Before the 23.06 release, there was a time where the qualification documents were supposed to be shipped separately, so to prevent broken links those requirements were duplicated in the safety manual.

Now that qualification documents are always shipped, there is no need to duplicate those requirements, and we can just have the safety manual link to them in the evaluation report. This commit does that.


japaric (approved):

japaric:
bors merge

PR #230, opened 2024-01-29
Attempt to automatically update `src/etc/completions/` on upstream pulls

Veykril:


ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -229,6 +230,16 @@ for prefix in "${DIRECTORIES_CONTAINING_LOCKFILES[@]}"; do
     fi
 done
 
+# We expose additional commands for `x.py` which affects the completions file generation,
+# so we just run the command to regenerate those in case they need updating as this usually
+# does not need manual intervention.
+echo "pull-upstream: checking whether ${GENERATED_COMPLETIONS_DIR} needs to be updated..."
+./x.py run generate-completions >/dev/null
+if git status --porcelain=v1 | grep "^ M ${GENERATED_COMPLETIONS_DIR}*" >/dev/null; then
```
pietroalbini, line 238:
```suggestion
if git status --porcelain=v1 | grep "^ M ${GENERATED_COMPLETIONS_DIR}" >/dev/null; then
```

I don't think the `*` would work well with the regex? It shouldn't be needed anyway, as we're matching from the start of the line up to the `/`, so files inside of the directory are matched without the `*` already.

Veykril: It does work (I tested this manually), but you are right it is pointless in this case since we don't match up to the line end explicitly

pietroalbini (approved):
bors merge

PR #232, opened 2024-01-29
Split and refactor the test outcomes page

pietroalbini:
This PR changes how the test outcomes page's content is structured, splitting it into one page per target. This has two purposes:

* Reduce the clutter on the page, which would otherwise mix information of all qualified targets.
* Provide an accurate list of ignored tests, as each target now has the list of tests it ignored rather than just displaying a list of tests ignored by *all* targets.

To simplify the generation of the separate pages, I changed how the `ferrocene_test_outcomes` extension works. Rather than having `.. ignored-tests::` and `.. suite-summary::` directives that are rendered in Python code, the extension now gathers and organizes the data, and defers the rendering of the information to a Jinja2 rST template.

This results in the whole test results page being rendered by the template, which eases the maintenance of those pages. For example, a note that should be displayed only on cross-compiled targets can now be gated behind `{% if host != target %}`.

Finally, while refactoring I removed the whole parsing code of the debug representation of steps, replacing it with the [structured test metadata](https://github.com/rust-lang/rust/pull/111936) I added a while back. This will increase the robustness of the tool.

There are still some open issues on this I'll address in a future PR:

* The list of crates for the bootstrap test suite is empty.
* The note for doc-tests not being executed is not present on aarch64.
* I'd like to add references to the test suite definitions in the evaluation plan.
* I'd like to see if I can make the information displayed in the page more concise.


Veykril (approved):

Veykril:
bors merge

pietroalbini:
Had to force push to fix the merge conflict.

Also, the last commit was added to fix the test failure. For some context there, other parts of the documentation were linking to the test results, but there was no single page to link to anymore. To address that, I added a very simple landing page that I plan to expand in a future PR. I didn't do the full implementation now to avoid adding even more complexity to this PR.

Veykril (approved):

Veykril:
bors merge

PR #233, opened 2024-01-29
Suggestions for docs from Pietro

amanjeev:
Adds sections to `upstread-pulls` and added two new files to note 

- Common failures and their fix procedures
- Useful commands and tools


pietroalbini:
Could you try putting the new sections all in the upstream pull page, rather than creating more pages? Also, most of the content of the current page is out of date and should be removed/updated :slightly_smiling_face:

amanjeev:
Closing. will open another one when ready.

PR #239, opened 2024-01-31
Add patching process to the qualification plan

pietroalbini:
This PR adds a new page to the qualification plan called "Patching process", which describes the process we must follow to publish bug fix releases without going through full requalification.


pietroalbini:
bors merge

Issue #224, opened 2024-01-26
Automatically inject the channel name in Sphinx


pietroalbini:
The channel name is currently hardcoded in `ferrocene/doc/sphinx-substitutions.toml`, while it should be dynamically calculated by the build system and injected there.

tshepang:
does https://github.com/ferrocene/ferrocene/pull/240 close this

PR #240, opened 2024-01-31
Move generate-build-metadata.sh functionality into a bootstrap dist step

Veykril:
cc https://github.com/ferrocene/ferrocene/issues/224


.circleci/workflows.yml
```diff
@@ -292,7 +292,7 @@ jobs:
           depth: 1
       - run:
           name: Generate build metadata
-          command: ferrocene/ci/scripts/generate-build-metadata.sh
+          command: ./x.py --stage 2 dist ferrocene-build-metadata
```
pietroalbini, line 295: Why `--stage 2`?

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -323,3 +325,107 @@ impl Step for TestOutcomes {
...
+    "ferrocene_version": "{ferrocene_version}",
+    "ferrocene_channel": "{ferrocene_channel}",
+    "release_channel": "{release_channel}",
+    "sha1_full": "{sha1_full}",
+    "sha1_short": "{sha1_short}"
+}}
+"#,
+            sha1_full = sha1_full.trim(),
+            sha1_short = sha1_short.trim(),
+        );
```
Veykril, line 421: We could go with a proper struct + serde here but I'm not sure whether that's worth it

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -323,3 +325,107 @@ impl Step for TestOutcomes {
...
+        // accordingly. Note that new releases *won't* be made until publish-release and
+        // this use the same version number.
+        let data = format!(
+            r#"{{
+    "metadata_version": 3,
+    "rust_version": "{src_version}",
+    "rust_channel": "{src_channel}",
+    "ferrocene_version": "{ferrocene_version}",
+    "ferrocene_channel": "{ferrocene_channel}",
+    "release_channel": "{release_channel}",
```
Veykril, line 414: Taking offers for a better name here

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -323,3 +325,107 @@ impl Step for TestOutcomes {
...
+    fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
+        run.alias("ferrocene-build-metadata")
+    }
+
+    fn make_run(run: RunConfig<'_>) {
+        run.builder.ensure(GenerateBuildMetadata);
+    }
+
+    fn run(self, builder: &Builder<'_>) -> Self::Output {
+        let dist_dir = "build/dist";
```
pietroalbini, line 346: To avoid problems we should add at the top of the file:

```rust
if builder.config.dry_run() {
   return;
}
```

src/bootstrap/src/ferrocene/dist.rs
pietroalbini: Just "channel" is also fine, that is what `publish-release` uses.

src/bootstrap/src/ferrocene/dist.rs
pietroalbini: IMO I'd use the `json!()` macro of `serde_json`.

.circleci/workflows.yml
Veykril: I copied what the other ci steps did when invoking dist, that's why

src/bootstrap/src/ferrocene/dist.rs
pietroalbini: The main concern I have around dry run is not the `--dry-run` flag (I haven't seen it much used in the wild), but that bootstrap will run steps twice, first with dry run and then without it. This means you're executing the `git` commands twice.

.circleci/workflows.yml
pietroalbini: Oh ok! The others use `--stage 2` to make sure they don't use the stage1 compiler, but it's not really needed for this specific dist step.

src/bootstrap/src/ferrocene/dist.rs
Veykril: Aight, changed it

pietroalbini (approved):
bors merge

pietroalbini:
Note that this will need to be backported to release/1.68, release/1.75 and release/1.76.

PR #245, opened 2024-02-01
Inject channel dynamically into sphinx substitutions

Veykril:


Veykril:
We could have the doc step run the dist step to generate a typed build metadata object without writing it to file and fetch the relevant fields from that as well. That would look cleaner at the expense of making the dist step a bit more complicated. Happy to do either.

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_qualification/__init__.py
```diff
@@ -20,6 +20,7 @@ def setup(app):
     app.add_config_value("ferrocene_private_signature_files_dir", None, "env", [str])
     app.add_config_value("ferrocene_version", None, "env", [str])
     app.add_config_value("rust_version", None, "env", [str])
+    app.add_config_value("channel", None, "env", [str])
```
pietroalbini, line 23: This seems to be missing the change in `substitutions.py`

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -431,3 +405,39 @@ impl Step for GenerateBuildMetadata {
         builder.copy_to_folder("ferrocene/packages.toml".as_ref(), dist_dir.as_ref());
     }
 }
+
+impl GenerateBuildMetadata {
+    pub(super) fn channel(
```
pietroalbini, line 410: Nit: I'd move this to a standalone function `ferrocene_channel` in `src/bootstrap/src/ferrocene/mod.rs`

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -431,3 +405,39 @@ impl Step for GenerateBuildMetadata {
...
+        rust_channel: &str,
```
pietroalbini, line 411: This can be accessed with `builder.config.channel`. I'd accept a builder instance instead of this argument.

pietroalbini: Yeah that is the expected value though, locally you're using the `dev` channel. `src/ci/channel` is never read by bootstrap, it is configured by CI [by setting the `rust.channel` key in `config.toml` to the value of that file](https://github.com/ferrocene/ferrocene/blob/bc59de8a10dacc2b649df88f521cb51ecbcf58d9/ferrocene/ci/configure.sh#L233-L240), and upstream CI does the same.

If we were to read `src/ci/channel` ourselves we would stop respecting `rust.channel`, which would be super confusing. I'd prefer if we'd access it and then treat `dev` as `nightly` in the match above.

pietroalbini (approved):
bors merge

pietroalbini (approved):
bors merge

PR #246, opened 2024-02-01
Add support for metadata version 3 in calculate-release-job-matrix

Veykril:


ferrocene/ci/scripts/calculate-release-job-matrix.py
```diff
@@ -207,13 +194,42 @@ def build_metadata(ctx, commit):
             # the "rolling" channel, so we hardcode it.
             ferrocene_channel="rolling",
             ferrocene_version="rolling",
+            channel=channel
         )
     elif metadata["metadata_version"] == 2:
+        rust_channel=metadata["rust_channel"]
```
pietroalbini, line 200: It looks to me that https://github.com/ferrocene/publish-release/pull/43 rejects any metadata version that is not `3`. In that case we can stop handling version 1 and version 2, since they'd fail anyway as soon as the release job starts.

pietroalbini (commented):
We can probably remove some code here! Also, there seems to be a spurious submodule update.

ferrocene/ci/scripts/calculate-release-job-matrix.py
Veykril: I think we need to keep version 2 for a bit? As running the script locally fails with a version of 2

Veykril: Unless you mean silently ignoring version 1 and 2 instead of raising an error

Veykril: For context, `RuntimeError: unexpected ferrocene-ci-metadata.json version `2` (for commit `9a68cf18621d6d96c7d96a62f0acf70629d88b8c`)`

Veykril: Oh, I guess this is might be due to the backports missing?

pietroalbini (approved):
bors merge

PR #258, opened 2024-02-05
Generate coverage report for core library

gnana-ganesh-tw:
This PR adds functionality to generate code coverage reports from the coverage data. The coverage data is generated by following the implementation of [this PR](https://github.com/ferrocene/ferrocene/pull/180).

## Usage
To generate the coverage report: 
```bash
./x.py run tools/generate-coverage-report
```

It is generated using the `grcov` utility.  So you might have to install `grcov` before running this command.

```bash
cargo install grcov
```

The `grcov` utility is needed to convert the raw coverage data to a human readable report. An `html` report is created in the following path `build/coverage_report`.

## Implementation

A new tool named `generate-coverage-report` is added in the `ferrocene/tools`. This tool internally uses `grcov` to generate the report.

`GenerateCoverageReport` build step is added in the `src/bootstrap/ferrocene/run.rs`. This checks if the `grcov` is already installed on the system and sets the environment variables with the required values to pass it to the `generate-coverage-report` tool.

`LLVM_PROFILE_FILE` environment variable can be set with the path of the `.profraw` files when running `generate-coverage-report` tool. This is useful when the coverage data is generated in a different directory. 

```bash
LLVM_PROFILE_FILE= path/to/profraw/files ./x.py run tools/generate-coverage-report
```

## Quirks

- The report is generated based only on the tests in the `library/core` excluding `benches` and `doc` tests for now.
 
- In the report, certain code region are incorrectly marked as uncovered. We believe this is due to an issue with `grcov`. So we might have to use `llvm-cov` to generate the report in the future.

- There are certain lines in the source code that should not be reachable and cannot be tested. Hence it cannot be covered in the report. To address this, the tool looks for "COVR_EXCL_LINE" comment in the source code and ignores these lines in the report.

For example, in the given code snippet the line is never reached, In this case, by adding the comment it is ignored in the coverage report.

```rust
None => unsafe { hint::unreachable_unchecked() }, // COVR_EXCL_LINE
```

Edit:
1. Added note about using `LLVM_PROFILE_FILE`

@pietroalbini @tw-amit  @Ax9DTW


ferrocene/tools/generate-coverage-report/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
pietroalbini, line 1: Are there plans to add more functionality in the `generate-coverage-report` tool? Otherwise we could inline its functionality in bootstrap itself.

src/bootstrap/src/ferrocene/run.rs
```diff
@@ -99,3 +100,139 @@ impl Step for TraceabilityMatrix {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub(crate) struct GenerateCoverageReport;
+
+impl Step for GenerateCoverageReport {
+    type Output = ();
+    const ONLY_HOSTS: bool = true;
+    const DEFAULT: bool = true;
+
+    fn run(self, builder: &Builder<'_>) -> Self::Output {
+        builder.info(&format!("Generating coverage report"));
```
pietroalbini, line 113: Nit: redundant call to `format!()`.

src/bootstrap/src/ferrocene/run.rs
```diff
@@ -99,3 +100,139 @@ impl Step for TraceabilityMatrix {
...
+
+        let coverage_report_data_dir =
+            env_llvm_profile_data_dir().unwrap_or_else(|| builder.out.join("coverage"));
+
+        let coverage_report_out_dir = builder.out.join("coverage_report");
+
+        let coverage_src_path = builder.out.join("../library/core").canonicalize().unwrap();
```
pietroalbini, line 120: `builder.src` points to the root of the repository, rather than going back with `..`.

src/bootstrap/src/ferrocene/run.rs
```diff
@@ -99,3 +100,139 @@ impl Step for TraceabilityMatrix {
...
+    let cargo_toml_str =
+        std::fs::read_to_string("library/core/Cargo.toml").expect("Failed to read Cargo.toml");
+    let cargo_toml_data: toml::Value =
+        toml::from_str(&cargo_toml_str).expect("Failed to parse Cargo.toml");
+
+    let tests = cargo_toml_data.get("test").expect("Failed to get tests");
+    let tests_array = tests.as_array();
+    let tests_array = tests_array.unwrap();
+    let test = &tests_array[0];
+    let test = test.as_table().unwrap();
```
pietroalbini, line 196: Can this be fetched from `builder.crates`? The source code that populates it is in https://github.com/ferrocene/ferrocene/blob/main/src/bootstrap/src/core/metadata.rs.

src/bootstrap/src/ferrocene/run.rs
```diff
@@ -99,3 +100,139 @@ impl Step for TraceabilityMatrix {
...
+        if !grcov_installed {
+            panic!("grcov not installed, to install it run:\n cargo install grcov");
+        }
+
+        let mut cmd = builder.tool_cmd(Tool::FerroceneGenerateCoverageReport);
+        cmd.env("COVERAGE_REPORT_DATA_DIR", coverage_report_data_dir);
+        cmd.env("COVERAGE_REPORT_OUT_DIR", coverage_report_out_dir);
+        cmd.env("COVERAGE_REPORT_SRC_PATH", coverage_src_path);
+        cmd.env("COVERAGE_REPORT_BIN_PATH", core_tests_binary_path);
+        cmd.env("RUSTC", rustc_path);
```
pietroalbini, line 162: This environment variable doesn't seem to be used by the tool?

ferrocene/tools/generate-coverage-report/Cargo.toml
gnana-ganesh-tw: We can plan to add `llvm-cov` to generate report in the future using this tool. In that case, we have to handle the parameters to be passed between `grcov` and `llvm-cov`, this can be done using this tool.

So, we think it is better to have it as a tool.

src/bootstrap/src/ferrocene/run.rs
gnana-ganesh-tw: `RUSTC` environment variable is required by grcov to locate the right version of the `llvm-cov`. If the variable is not set `grcov` uses the the `rustc` in the host. This version mismatch of `llvm-cov` leads to empty report.  So, we need the `stage1/rustc` to create the report.

`grcov` uses `cargo-binutils` to invoke `llvm-cov`. `cargo-binutils` looks for `RUSTC` env variable to get the path for `llvm-cov`.

https://github.com/rust-embedded/cargo-binutils/blob/5c38490e1abf91af51d0a345bb581e37facd28ff/src/rustc.rs#L8.

gnana-ganesh-tw:
We have implemented the changes mentioned in the comments.

src/bootstrap/src/ferrocene/run.rs
pietroalbini: Right, missed it!

src/bootstrap/src/ferrocene/run.rs
pietroalbini: The `crates` field is populated in https://github.com/ferrocene/ferrocene/blob/main/src/bootstrap/src/core/metadata.rs. That files parses the output of `cargo metadata` ([fields reference](https://doc.rust-lang.org/nightly/cargo/commands/cargo-metadata.html)) and stores it in the field.

If the information you need is present in `cargo metadata` you can add the relevant serde structs/fields in the file to parse it! Looking at the fields reference, it should provide all the information needed for all the test binaries it runs.

ferrocene/tools/generate-coverage-report/Cargo.toml
pietroalbini: Hmm, I think it'd be better to pick between grcov and llvm-cov and stick to just one of them. This will be used to generate the reports to include in our documentation, and we're going to include a single copy anyway.

For the time being I'd still prefer if we remove the dedicate tool and invoke grcov/llvm-cov from bootstrap directly. Right now it just adds a level of indirection, and if in the future we'll reach a point where we need more complex logic we can revisit whether to have a separate tool.

src/bootstrap/src/ferrocene/run.rs
gnana-ganesh-tw: Since we have only one test binary name for now, can we just hardcode the binary name as "coretests" ?
Given that we are using the code coverage for core library only at this point in time.

ferrocene/tools/generate-coverage-report/Cargo.toml
gnana-ganesh-tw: @pietroalbini We need your input for choosing between llvm-cov and grcov.

llvm-cov , produces more accurate data, but it does not have a way to skip lines/branch/region from the coverage report.

grcov, produces less accurate report (sometimes accounting the lines as not covered), but it has a way to skip lines from the coverage.

src/bootstrap/src/ferrocene/run.rs
pietroalbini: Yeah that's fine.

ferrocene/tools/generate-coverage-report/Cargo.toml
pietroalbini: I'd prefer llvm-cov then. Let's merge this PR and switch to llvm-cov in a future one.

Ax9DTW:
Is bors stuck again 😢

tshepang:
bors merge

tshepang (approved):
bors merge

tshepang:
looks like it needs an empty commit

Hoverbear:
This seems to have gotten stuck again?

Ax9DTW:
Needs a bors merge I think

tshepang (approved):
bors merge

PR #259, opened 2024-02-06
Remove more boilerplate

pietroalbini:
This PR further reduces the boilerplate present in our qualification documents.

**Purpose page:**  
The purpose page was present in all documents, and contained an overview of the structure of the Ferrocene toolchain. This isn't actually needed in most documents, so it has been removed. The only exceptions is the Evaluation Plan, where we need to describe what we are evaluating. Also, note that in the Qualification Plan the page was redundant, as a more detailed description is present in the "Ferrocene Details" page.

**Associated documents page:**  
The associated documents page was redundant as it merely referenced either standards (that are listed in the Document List) or other documents (that are part of the same package). It has been removed, replaced with a link to the document list in the index.


Urhengulas (approved):

pietroalbini:
bors merge

PR #266, opened 2024-02-08
Update target-names.toml

jonathanpallant:
Renamed the target strings to match Arm preferences.

See for example https://developer.arm.com/documentation/ddi0403/ee/?lang=en

> An Armv7-M implementation that includes the DSP extension is called an Armv7E-M implementation


pietroalbini (approved):
bors merge

PR #225, opened 2024-01-26
Create central table of human-readable target names

pietroalbini:
This PR adds a new role to the `ferrocene_qualification` extension: `:target:`. The role converts the target triple into the human-readable name based on the new `ferrocene/doc/target-names.toml`. The purpose of this is twofold:

* It increases consistency across the documents, as we have a single place where they are defined.
* It helps with a WIP PR I'm working on to split the [test results](https://public-docs.ferrocene.dev/main/qualification/report/tests.html) page into per-target pages, as it means the rest of the automation can fetch the human-readable target name programmatically.

PR #282, opened 2024-02-14
Add `:target-with-triple:` and document it

pietroalbini:
This PR adds `:target-with-triple:`, in addition to the existing `:target:` added in #225, as per Jonathan's suggestion. It also adds documentation for both `:target:` and `:target-with-triple:` in the IP.


Veykril:
bors merge

PR #283, opened 2024-02-14
fix segfaulting rustc when `download-ci-llvm = true`

japaric:
- the segfault occurred in the second / third C++ frame after `rustc` called into LLVM using the `PassWrapper.cpp` glue
- the segfault occurred due a ABI mismatch between those two frames
- both frames corresponded to LLVM C++ functions. (the first frame was the `PassWrapper.cpp` glue)
- one of the frames (functions) came from the LLVM library built in CI; the other frame was codegen'd locally
- when comparing to a rust-lang/rust build, it was found that both frames involved in the segfault were present in the LLVM library built in rust-lang/rust CI. the rust-lang/rust artifact did not result in a segfault
- the differences between the rust-lang/rust and ferrocene build settings are: (a) rust-lang/rust builds llvm as shared library; ferrocene builds them as static libraries. (b) rust-lang/rust uses `clang` and a newer version of `gcc` / `g++`
- building LLVM as shared libraries in ferrocene did not fix the segfault but additionally bumping the `gcc` / `g++` version (by way of using a newer base docker image) did

### DIFF
--- a/ferrocene/ci/configure.sh
+++ b/ferrocene/ci/configure.sh
@@ -66,6 +66,21 @@ add --enable-cargo-native-static
 # might require libstdc++ to be installed on the user's system.
 add --enable-llvm-static-stdcpp
 
+# Build LLVM as shared libraries
+#
+# `llvm.static-stdcpp` requires dynamic LLVM libraries to work properly. A
+# dynamic library (`.so`) can hold a copy of `libstdc++.a` whereas static
+# libraries (`.a`) will not include a copy of `libstdc++.a`.
+#
+# Building `rustc` against static LLVM libraries will use the local version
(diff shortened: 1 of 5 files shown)


japaric:
bors try

japaric:
bors try

japaric:
bors try

japaric:
bors try

japaric:
bors try

tshepang (approved):
bors merge

tshepang:
we can get the directory (ubuntu-18) renamed (to ubuntu-20) in a future PR

tshepang:
bors retry

pietroalbini:
Superseded by https://github.com/ferrocene/ferrocene/pull/376.

PR #286, opened 2024-02-14
Rework the standards traceability section

pietroalbini:
Before this PR, the "Validation process" page of the QP had a [Traceability Matrix](https://public-docs.ferrocene.dev/main/qualification/plan/validation.html#traceability-matrix) section mapping the standard clauses with our documents. That was hard to find and kinda misplaced in "validation process", so this PR moves this to its own top-level section, with each standard having its page:

image

While I was doing this, I also made some improvements:

* I converted the table from `csv-table` to a `list-table`, which is way easier to read and edit. This was done automatically, the commit message includes the command I used.

* I swapped columns so that the link to our own page is now the second column rather than the third, meaning you don't have to scroll to get to that link:

  image

* I tweaked the IEC 61508 docs to have one document per line, reducing the overall table width.

### DIFF
--- a/ferrocene/doc/qualification-plan/src/index.rst
+++ b/ferrocene/doc/qualification-plan/src/index.rst
@@ -27,6 +27,13 @@ Ferrocene in accordance to the standards above.
    validation
    patching
 
+.. toctree::
+   :caption: Traceability of standard requirements:
+   :numbered:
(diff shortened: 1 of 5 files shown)


tshepang (approved):
bors merge

PR #288, opened 2024-02-15
Add x86_64 to supported environment targets

Veykril:


pietroalbini:
Can we switch to a "normal" table where we have a row for each host/target pair? Like we currently do in the QR: https://public-docs.ferrocene.dev/main/qualification/report/tests/index.html. This is also because the set of libraries is different between targets. We might also want to reword the libs column to "uncertified libraries".

Veykril:
image

pietroalbini:
Awesome! A few last changes:

* Can you use inline code for `core` and `alloc`?
* x86-64 linux target has all the libraries available, not just the no-std ones

pietroalbini (approved):
bors merge

PR #290, opened 2024-02-15
Bug fixes for `--serve`

pietroalbini:
This PR makes two fixes to the `--serve` command:

* Occasionally rebuilding the documentation would fail due to the `ferrocene_toctrees` extension failing to disable the builtin toctree collector. Now the extension will attempt disabling it only when it's not already disabled.
* `--fresh --serve` would not work as expected, as it'd do an initial fresh build and then incremental builds. Now all followup builds will be fresh too.


Veykril (approved):

Veykril:
bors merge

PR #291, opened 2024-02-15
Rework the validation process page

pietroalbini:
When working with https://app.clickup.com/t/8693u2ge5 I ended up rewriting most of the validation process page... so here we are! Most of the content is new, and in general clarifies the steps that are automated while reducing the manual checks when not strictly needed.


Veykril:
bors merge

PR #295, opened 2024-02-16
Allow building all Sphinx docs at once, and error out if `--serve` is misused

pietroalbini:
This PR makes a few quality of life changes to the tooling used to build the documentation:

* There is now a new `./x doc ferrocene/doc`, which builds all the Sphinx documents at once. This is easier than manually listing all documents in the single `./x` invocation, and it's way faster than invoking `./x doc` multiple times[^1]. Bikeshedding welcome on the name, I'm not too happy about it (the only constraint is having the `ferrocene-` prefix).

* Passing `--serve` when multiple documents are requested will now print an error and exit. It was already unsupported to pass `--serve` with multiple documents, as it would only start the server for the first one, but it was a silent and unintuitive.

[^1]: When invoking `./x` to build any document, the build system first gathers the intersphinx artifacts from all the documents, and only then actually builds the requested document(s). Gathering the intersphinx artifacts takes a few seconds. If you build multiple documents in the same invocation, the intersphinx artifacts will be reused, but if you invoke `./x` multiple times they will be regenerated every single time.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/doc-procedures.rst
+++ b/ferrocene/doc/internal-procedures/src/doc-procedures.rst
@@ -50,7 +50,16 @@ paths:
 
    ./x doc ferrocene/doc/$foo ferrocene/doc/$bar ferrocene/doc/$baz
 
-If you want to build all of the available documentation, you can omit the path:
+If you want to build all the documentation generated with Sphinx (for example
+if you need to ensure cross-document links work, or you need to test an
+extension change across all documents) you can run:
+
+.. code-block:: text
+
+   ./x doc ferrocene-all-sphinx
+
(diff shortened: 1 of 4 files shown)


Veykril (approved):
Some ideas: ` ferrocene-sphinx-complete`, `ferrocene-sphinx-full`, `ferrocene-docs-full`, `ferrocene-docs-complete`, `ferrocene-docs-all`

amanjeev:
`ferrocene-docs-all` is my vote.

pietroalbini:
Renamed to `ferrocene/doc` after internal discussion.

tshepang (approved):
bors merge

PR #296, opened 2024-02-16
Prevent glossary terms to be matched in the middle of a word

pietroalbini:
This PR changes the `ferrocene_autoglossary` extension to only match terms if they are not in the middle of a word, by adding a filtering steps after a sentence is lexed. This results in the following diff across the document (with matched terms within `[]`):

```diff
-Bri[git]te Markmann and Leslie Glass
+Brigitte Markmann and Leslie Glass

-All the qualification documents we send to TUV have to be di[git]ally signed by the responsible parties, to attest they were reviewed and to prevent accidental changes to the documents (which would require TUV to review the documents again).
+All the qualification documents we send to TUV have to be digitally signed by the responsible parties, to attest they were reviewed and to prevent accidental changes to the documents (which would require TUV to review the documents again).

-The Ferrocene [QMS] documentation also needs to be di[git]ally signed by responsible parties to attest that their contents are up to date and represent the latest applicable procedures to be applied throughout the organization.
+The Ferrocene [QMS] documentation also needs to be digitally signed by responsible parties to attest that their contents are up to date and represent the latest applicable procedures to be applied throughout the organization.

-Invoke cosign to di[git]ally sign the document metadata gathered earlier.
+Invoke cosign to digitally sign the document metadata gathered earlier.

-by 1. Note that you might need to remove some di[git]al signatures when you increment the version number.
+by 1. Note that you might need to remove some digital signatures when you increment the version number.
```

This also adds a debug helper to the extension that, when an environment variable is set, writes to a file all the nodes that contain at least one matched node. I used this while developing this PR to generate the diff above and ensure I was not breaking anything by accident, and I figured out it would be better to keep it committed but disabled by default for the next time we need to change the extension.


ferrocene/doc/sphinx-shared-resources/exts/ferrocene_autoglossary/debug.py
```diff
@@ -0,0 +1,78 @@
...
+        and getattr(node, "line", None) is not None
+    ):
+        return f"{node.source}:{node.line}"
+    elif node.parent is not None:
+        return node_location(node.parent)
+    else:
+        return "<unknown>"
+
+
+def debug_file():
```
tshepang, line 73: is this used

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_autoglossary/lexer.py
```diff
@@ -0,0 +1,131 @@
...
+def _filter_matches(matches):
+    previous_token_allows_match = True
+
+    for token in matches:
+        # Convert a match into a token if the previous token doesn't allow the
+        # following token to be a match.
+        if type(token) == MatchedTerm and not previous_token_allows_match:
+            token = token.text
+
+        # Only allow the next token to be a match if this is is a text token
```
tshepang, line 62: might as well...
```suggestion
        # Only allow the next token to be a match if this is a text token
```

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_autoglossary/debug.py
pietroalbini: Nope, leftover from a refactor. Good catch :tada:

PR #307, opened 2024-02-20
dist "all" tools

tshepang:


ferrocene/packages.toml
```diff
@@ -45,12 +45,38 @@ subset = "default"
...
+name = "rust-analyzer"
+subset = "default"
+
+[[groups.hosts.packages]]
+name = "clippy"
+subset = "default"
+
+[[groups.hosts.packages]]
+name = "miri"
+subset = "default"
```
pietroalbini, line 62: Miri is nightly-only, we can avoid shipping it.

pietroalbini (changes requested):
Two things also need change:

* In the evaluation plan we list tools that are out of scope: https://public-docs.ferrocene.dev/main/qualification/evaluation-plan/qualification-scope.html
* We should do a try build to see how long this makes the dist builder. If it's too long, we might want to split it into a `dist-tools` builder to avoid increasing CI duration.

tshepang:
bors try

PR #320, opened 2024-02-22
re-use the same blurb spread in glossary

tshepang:
noticed that Internal Procedures had a slight difference from the rest, so decided to have this shared to avoid such in future


pietroalbini (approved):
bors merge

PR #322, opened 2024-02-22
Update setup-local-env.rst

jonathanpallant:
Add `lld = true` and `download-ci-llvm = false` options to the example config.toml file because most developers will need those.


pietroalbini (approved):
bors merge

PR #323, opened 2024-02-22
Configure the ferrocene channel through config.toml

pietroalbini:
The Ferrocene channel is used to determine the overall channel of the release, by combining it with the Rust channel. Not all combinations are valid though, and some result in a bootstrap error.

Before this commit, the Ferrocene channel was read directly from the source (ferrocene/ci/channel), while the Rust channel was configured through config.toml and defaulted to "dev". This caused problems when the Ferrocene channel was not "rolling" (like in beta-24.05), as only rolling is compatible with the Rust "dev".

To address the problem, this commit moves the configuration of the Ferrocene channel to config.toml too, with a default of "rolling". This way, either the user configures both the Rust and Ferrocene channel together, or they don't configure either.

This should fix the failure in https://github.com/ferrocene/ferrocene/pull/317

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -887,6 +887,12 @@
 # =============================================================================
 [ferrocene]
 
+# The channel for the Ferrocene build to produce. This is merged with Rust's
+# channel to determine the full channel name.
+#
+# Can be "rolling", "beta" or "stable".
+#channel = "rolling"
+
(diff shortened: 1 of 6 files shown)


tshepang (approved):
bors merge

tshepang:
bors retry

PR #324, opened 2024-02-22
various text fixes

tshepang:


ferrocene/doc/user-manual/src/cli.rst
```diff
@@ -166,8 +166,9 @@ Command-Line Interface
 
       .. caution::
 
-         Only the arguments listed in section Compilation targets are within
-         the scope of the Ferrocene |iso_ref| qualification.
+         This code generation option is outside the scope of the Ferrocene
```
pietroalbini, line 169: This is not correct, and the old text should stay: this means, you *can* use `-C link-arg`, but only with the values we describe in each target's documentation.

Besides, we should remove |iso_ref| from all of these cautions.

pietroalbini: This needs to be applied to the other similar changes too, it's not resolved.

ferrocene/doc/user-manual/src/cli.rst
```diff
@@ -241,11 +241,6 @@ Command-Line Interface
 
    .. cli:option:: -C linker=<path>
 
-      .. caution::
-
-         Only the paths listed in section Compilation targets are within the
-         scope of the Ferrocene |iso_ref| qualification.
```
pietroalbini, line 247: We should say this that it is possible to use it only for targets that require a linker driver, and in that case the linker driver must respect the option.

tshepang: an attempt https://github.com/ferrocene/ferrocene/pull/324/commits/74af1d982dd352c1ebc3f35521dd94df8f99f791

ferrocene/doc/user-manual/src/cli.rst
```diff
@@ -241,6 +241,11 @@ Command-Line Interface
 
    .. cli:option:: -C linker=<path>
 
+      .. caution::
+
+         This can only be used for targets that require a linker driver,
+         in which case the linker driver must respect the option.
```
pietroalbini, line 247: Can this link to the safety manual?

pietroalbini (approved):
bors merge

PR #328, opened 2024-02-23
we do not (yet) define the "authorized" concept anywhere

tshepang:
Also, the public can also view and review PRs... the paragraph is stale.


pietroalbini (approved):
bors merge

PR #329, opened 2024-02-23
Add backports documentation

pietroalbini:
@tshepang pointed out that our backport procedures were not documented, so this adds a page to the IP to explain how backporting works.


tshepang (approved):
bors merge

PR #334, opened 2024-02-27
Update linker requirements

Urhengulas:
Fixes https://ferroussystems.clickup.com/t/8693tyea2.

Is `ref:` the correct way to link to a section on a page?


pietroalbini:
> Is ref: the correct way to link to a section on a page?

Have you tested locally whether it renders correctly? Not sure if that exact syntax will work.

pietroalbini (approved):
bors merge

PR #337, opened 2024-02-27
Convert csv to list tables

Urhengulas:
Fixes https://app.clickup.com/t/8693umm77.

### DIFF
--- a/ferrocene/doc/evaluation-plan/src/method.rst
+++ b/ferrocene/doc/evaluation-plan/src/method.rst
@@ -21,15 +21,19 @@ Through this analysis, the following potential error are identified:
 Errors, with the possibility of detection and mitigation, are added in
 the :doc:`safety-manual:index` with the following form:
 
-.. csv-table:: Safety Manual  potential error template
+.. list-table:: Safety Manual  potential error template
    :name: Safety Manual potential error template
-   :delim: !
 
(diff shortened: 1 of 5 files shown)


ferrocene/doc/evaluation-plan/src/method.rst
```diff
@@ -21,15 +21,19 @@ Through this analysis, the following potential error are identified:
 Errors, with the possibility of detection and mitigation, are added in
 the :doc:`safety-manual:index` with the following form:
 
-.. csv-table:: Safety Manual  potential error template
+.. list-table:: Safety Manual  potential error template
    :name: Safety Manual potential error template
```
Urhengulas, line 25: Does the name need to be duplicate?

pietroalbini: I mean, I'd personally remove table names completely. It doesn't need to be duplicated though.

Urhengulas: Removed them

pietroalbini: There are still some table names in the diff (they're not duplicated though).

Urhengulas: Removed them as well

pietroalbini (approved):
bors merge

PR #342, opened 2024-02-27
Use the new tarball naming scheme

pietroalbini:
This PR implements the new tarball naming scheme agreed in https://designs.ferrocene.dev/tarball-names.html. It also bumps the metadata version from 3 to 4, so that the release process can identify which naming scheme to use.

### DIFF
--- a/ferrocene/ci/scripts/calculate-release-job-matrix.py
+++ b/ferrocene/ci/scripts/calculate-release-job-matrix.py
@@ -211,7 +211,9 @@ def build_metadata(ctx, commit):
             ferrocene_version=ferrocene_version,
             channel=channel
         )
-    elif metadata["metadata_version"] == 3:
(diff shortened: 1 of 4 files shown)


tshepang (approved):
bors merge

pietroalbini:
bors cancel

Found a problem :sweat_smile:

pietroalbini:
Changed the code that generates the metadata to use the build system's methods to retrieve the commit hash rather than shelling out to git. This is actually a behavior change, since the build system always generates a short hash 9 chars long (compared to the dynamic length returned by git).

This was needed because we are generating the commit hash in the tarball name using the build system's methods, and that would've resulted in the metadata not matching the tarballs we generate.

tshepang (approved):
bors merge

tshepang (approved):
bors merge

PR #348, opened 2024-02-28
Pull new changes from `ferrocene/ferrocene` into `ferrocene/sphinx-shared-resources/` every day

Urhengulas:
Fixes https://ferroussystems.clickup.com/t/8693twnfz.

The last step is not ideal. The workflow should not create a new (empty) commit if there are no changes. This is achieved by this solution. But it will also mark the workflow as failed, which is not true, because it is okay if there are no new changes.


pietroalbini:
Can you make the whole mirroring process a single (Python/Bash) script that CI then executes? This allows you to test and run the changes locally.

Urhengulas:
> Can you make the whole mirroring process a single (Python/Bash) script that CI then executes? This allows you to test and run the changes locally (pointing it to a different destination repository).

Sure. Can I also write it in Rust, as a xtask?

Veykril:
That probably overcomplicates things, a python script for this should be rather simple. Take a look at one of the `ferrocene/tools` ones like https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/update-lockfiles/update.py

pietroalbini:
> Sure. Can I also write it in Rust, as a xtask?

I'd rather avoid that: all of our infra automation is in Python/Bash, and Rust mostly gets in your way when your only task is executing a few git commands. Let's keep it in py/sh for consistency.

Urhengulas:
I moved the logic into 'ferrocene/ci/scripts/mirror-sphinx-shared-resources.py`.

I also changed the direction of the mirroring. Instead of `ferrocene/sphinx-shared-resources` pulling from `ferrocene/ferrocene`, `ferrocene/ferrocene` is now pushing to `ferrocene/sphinx-shared-resources`. I mostly did that because otherwise the script could not live in `ferrocene/ci/scripts/`. Therefore the workflow is now executed in `ferrocene/ferrocene`.

One open question is if the workflow should open a PR (like it is done for the other automations), or if it should directly push to the `main` branch. Directly pushing would require circumventing the branch protection rules. Potentially it makes sense to change the mirroring direction back to pull instead of push, because I assume that a workflow can create commits in its own repository more easily compared to creating commits in another repository.

pietroalbini:
> I also changed the direction of the mirroring. Instead of ferrocene/sphinx-shared-resources pulling from ferrocene/ferrocene, ferrocene/ferrocene is now pushing to ferrocene/sphinx-shared-resources. I mostly did that because otherwise the script could not live in ferrocene/ci/scripts/. Therefore the workflow is now executed in ferrocene/ferrocene.
>
> Potentially it makes sense to change the mirroring direction back to pull instead of push, because I assume that a workflow can create commits in its own repository more easily compared to creating commits in another repository.

I agree that pulling is better, because you can run it in `ferrocene/sphinx-shared-resources`'s GitHub Actions and automatically get a token with write perms without having to maintain a separate GitHub App / PAT. I think it's fine for the script to live in the `ferrocene/sphinx-shared-resources` repository.

> One open question is if the workflow should open a PR (like it is done for the other automations), or if it should directly push to the main branch. Directly pushing would require circumventing the branch protection rules. 

It would be better for the script to push directly to main, adding a PR step doesn't add any value (as all changes are always required to be reviewed in `ferrocene/ferrocene`). Branch protection rules can be adjusted accordingly.

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if not new_commit:
+        print("No changes")
+        return
+
+    # push the commit
+    run(f"git push -u origin {branch_name}")
+
+    # create pull request
+    run(f"gh pr create --body {""} --title \"{COMMIT_MSG}\" --head \"{branch_name}\" --base main")
```
pietroalbini, line 55: I would not push and create the PR in the script itself, as that hinders testing stuff locally. If the script finishes at creating the commit you can run it however many times locally while testing without worrying it messes up the remote repository :slightly_smiling_face:

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def run(cmd: str, check=True) -> subprocess.CompletedProcess:
+    print(f"🏃 {cmd}")
+    args = shlex.split(cmd)
```
pietroalbini, line 60: I would not use `shlex.split` here, but instead accept the array in the `run` function signature. For example, it'd turn

```python
run(f'git commit -m "{COMMIT_MSG}"', check=False)
```

...into:

```python
run(["git", "commit", "-m", COMMIT_MSG])
```

The split function doesn't do any quoting, so the code you wrote above would break if the commit message for example contained `"`s.

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
```diff
@@ -0,0 +1,71 @@
...
+    run(f"git clone {MIRROR_REPO} .")
+
+    # create branch
+    rand_id = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
+    branch_name = f"automation/mirror-sphinx-shared-resources/{rand_id}"
+    run(f"git switch -c {branch_name}")
+
+    # copy contents of mirror directory into mirror repository
+    # `-T` is responsible for only copying the content
+    run(f"cp -frT {cwd}/{MIRROR_DIR} .")
```
pietroalbini, line 40: Copying the files over is probably not the behavior we want, as if a file is deleted in the monorepo it would not be deleted in the mirrored repository.

Given that it'd be better to move this script to the mirrored repo, and we'll want some non-mirrored files there (like a README explaining why there is a mirror, or license files), it would probably be better to keep a list of files to preserve, then delete all the other files, and copy the new files over.

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
Urhengulas: I think `shlex.split` does work:

```python
>>> import shlex
>>> msg = 'Does "this" work?'
>>> shlex.split(f'git commit -m "{msg}"')
['git', 'commit', '-m', 'Does this work?']
```

The `str.split` would not work though:
```python
>>> f'git commit -m "{msg}"'.split(" ")
['git', 'commit', '-m', '"Does', '"this"', 'work?"']
```

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
Urhengulas: > Copying the files over is probably not the behavior we want, as if a file is deleted in the monorepo it would not be deleted in the mirrored repository.

True. I did not consider deletion.

Urhengulas: > Given that it'd be better to move this script to the mirrored repo, and we'll want some non-mirrored files there (like a README explaining why there is a mirror, or license files), it would probably be better to keep a list of files to preserve, then delete all the other files, and copy the new files over.

Would it be a problem to put those files into the subfolder in `ferrocene/ferrocene`?

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
pietroalbini: In that case it works by accident, but for example in this case it produces the wrong outcome:

```python
>>> msg = 'foo" bar "baz'
>>> shlex.split(f'git commit -m "{msg}"')
['git', 'commit', '-m', 'foo', 'bar', 'baz']
```

I'd prefer to pass the array to the function rather than using shlex here.

Urhengulas:
I was just thinking that we could also run the workflow every time we make changes to `ferrocene/doc/sphinx-shared-resources`, instead of running it periodically. We do not make that many changes to it and then we could be certain that when the action runs that there are actually changes as well.

```yml
on:
  push:
    branches:
      - "main"
    paths:
      - "ferrocene/doc/sphinx-shared-resources/*"

```

pietroalbini:
> I was just thinking that we could also run the workflow every time we make changes to ferrocene/doc/sphinx-shared-resources, instead of running it periodically. We do not make that many changes to it and then we could be certain that when the action runs that there are actually changes as well.
>
> If we do this we can just take the last commit (the merge commit) and apply the changes from it to the mirror.

That's a great idea! I'll work and setup credentials.

pietroalbini:
Configured the credentials.

You will need to use the `mirror-sphinx-shared-resources` environment (see in the other workflows how to configure it), and add this step to fetch the relevant token:

```yaml
- name: Authenticate with GitHub
  run: ferrocene/ci/github-app-auth.py
  env:
    # https://github.com/organizations/ferrocene/settings/apps/ferrocene-mirror-shared-resources
    APP_ID: 852121
    APP_PRIVATE_KEY: "${{ secrets.MIRROR_APP_PRIVATE_KEY }}"
  id: github_app_auth
```

Then in following steps you can get the token with `${{ steps.github_app_auth.outputs.token }}` rather than `${{ secrets.GITHUB_TOKEN }}`.

Urhengulas:
Thank you Pietro! 🙇🏾

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
Urhengulas: okeee

Urhengulas: Fixed it

Urhengulas:
Hey @pietroalbini, I struggle to test the workflow.

> Branch "jh/mirror-sphinx-shared-resources" is not allowed to deploy to mirror-sphinx-shared-resources due to environment protection rules.

I assume the environment is configured to only allow deployments from `main`. Can we relax this for testing the workflow?

.github/workflows/mirror-sphinx-shared-resources.yml
```diff
@@ -0,0 +1,47 @@
...
+
+# This workflow runs when a PR gets merged into the main branch and includes
+# changes to any files in ferrocene/doc/sphinx-shared-resources/.
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - "ferrocene/doc/sphinx-shared-resources/**"
+  pull_request:
```
pietroalbini, line 15: I know this is used for testing, leaving a comment as a reminder to remove this line before we merge.

pietroalbini:
Updated branch protection rules.

Urhengulas:
We should also run the formatting and style checks in the monorepo, because it is the source of truth.

Urhengulas:
I added the changes to fix the CI of https://github.com/ferrocene/sphinx-shared-resources/pull/39 here as well. The reason is that the repositories need to be put in sync manually once, which will be the case when both PRs are merged. 

PS. Let's hope we do not create a race condition with some other change.

pietroalbini (commented):
Awesome, the current implementation looks great!

The only remaining concern I have is, if this misses commits for whatever reason (like a GitHub Actions outage) then it'll require manual action to re-sync, as this only supports mirroring a single commit.

I think addressing that won't take too much time. We can change the commit message to also include the commit hash it mirrored (taken from `git rev-parse HEAD`):

```
Automatically push changes from ferrocene/ferrocene

mirrored-commit: 04d6237702837e2a1708eecfd609b33f9ca7575d
```

Then the script can search all commit messages (`git log "--format=%B"`) in the mirror, pick the first line starting with `mirrored-commit:`, and use it rather than `HEAD~1` as the diff base.

This would require the first manual commit in the repo to include the `mirrored-commit` line, but otherwise it would be more resilient than the current solution.

.github/workflows/mirror-sphinx-shared-resources.yml
```diff
@@ -0,0 +1,60 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+---
+name: Mirror sphinx-shared-resources
+
+# This workflow runs when a PR gets merged into the main branch and includes
+# changes to any files in ferrocene/doc/sphinx-shared-resources/.
+on:
+  push:
```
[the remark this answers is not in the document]
Urhengulas: Added it

Urhengulas:
I changed it so it considers all changes since the last mirror event. The only disadvantage is that we now need to fetch the full history of ferrocene/ferrocene (which is big), because the last mirror event might be some time ago.

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
```diff
@@ -27,40 +29,63 @@ def main():
...
+def get_commit_msg(origin_repo_path: str) -> str:
+    # get the hash of the commit which triggered the workflow ...
+    origin_commit_hash = run(["git", "rev-parse", "HEAD"], origin_repo_path).stdout
+    # ... and construct the commit message with it
+    return COMMIT_MSG_START + origin_commit_hash
+
+
+def get_last_mirrored_commit(mirror_repo_path: str) -> str:
+    commit_message = run(
+        ["git", "log", "--format=%B", "-n", "1", "HEAD"], mirror_repo_path
```
pietroalbini, line 68:
```suggestion
        ["git", "log", "--format=%B", "-n", "1", "HEAD"], mirror_repo_path
```

Checking only the first commit (`-n 1`) will break if we send PRs to the mirrored repo only. While *in general* we shouldn't be doing it, I can see us maintaining a separate README or license files in the mirrored repo.

I would remove the `-n 1` and get concatenated commit bodies of all the commits in the history. Then we can iterate line by line and break at the first line beginning with `mirrored-commit: `.

pietroalbini (commented):
Last round of changes :D

ferrocene/ci/scripts/mirror-sphinx-shared-resources.py
Urhengulas: Changed it

Urhengulas:
@pietroalbini Should be ready now. Can you please also take a brief look at https://github.com/ferrocene/sphinx-shared-resources/pull/39. It needs to be merged before the action is executed for the first time.

PR #365, opened 2024-03-04
Remove the record-branch-commit job

Hoverbear:
Speaking with @pietroalbini today we discussed that this task should be unnecessary now that the Ferrocene repository is public.

I suspect this PR will be broken as I am still learning how it works.


.circleci/workflows.yml
```diff
@@ -509,18 +493,6 @@ workflows:
             - x86_64-linux-self-test
             - wasm-dist-oxidos
 
-
-  # Record the branch commit for all the jobs bors merges commit into.
-  record-branch-commit:
-    when:
-      or:
-        - equal: [<< pipeline.git.branch >>, "main"]
```
pietroalbini, line 517: We need to move these conditions to the `skip` job, otherwise CircleCI will fail on every `main` commit.

pietroalbini (commented):
After this change, the release tooling will need to be updated not to rely on these files anymore.

Hoverbear:
I was looking for references to the created files and found this,
https://github.com/ferrocene/publish-release/blob/ba33d5c03ae7f5e8a4876af8d1c61aebe9dcc00c/tests/integration/harness/local_env.rs#L146-L202. 

As far as I can see that's the only reference?

pietroalbini (approved):
Changes look good! We can merge this when we remove the places in the infra that rely on the data uploaded by this job.

Hoverbear:
`publish-release` side changes should be good to go, I'll mark this ready after they merge.

pietroalbini:
bors merge

PR #371, opened 2024-03-06
Properly exclude `build-metrics.json` from Tar

Hoverbear:
Plucked from https://github.com/ferrocene/ferrocene/pull/283, this fixes the "Invocation Stacking" issue in the generated reports from CI.


tshepang (approved):
bors merge

PR #362, opened 2024-03-04
bump dockerfile os

tshepang:

PR #376, opened 2024-03-08
Update to Ubuntu 20.04 and fix downloading LLVM from CI

pietroalbini:
This PR is based on https://github.com/ferrocene/ferrocene/pull/283 and https://github.com/ferrocene/ferrocene/pull/362: it updates CI to run on Ubuntu 20.04 (fixing the failure in https://github.com/ferrocene/ferrocene/pull/305), and changes how we build LLVM to hopefully fix `download-ci-llvm`. Notably, this bumps the minimum glibc version requirement to 2.31.

The hard part of this PR was fixing the broken CI for Aarch64. It turns out that the failure was caused by a breaking change in `binfmt-support` package, which now requires manually enabling new binfmts after importing them (if `/proc/sys/fs/binfmt_misc` is not mounted, like in CI).


Hoverbear (approved):

pvdrz:
bors merge

PR #383, opened 2024-03-11
Fix downloading tarballs from CI after the renaming

pietroalbini:
In #342 we switched to a new tarball naming scheme, but we didn't update the bootstrap code that downloads them to use the new naming. This PR fixes that.

Also, with this and #376 downloading LLVM from CI works again! I thus updated the docs to remove the need to disable CI-downloaded LLVMs.


Veykril:
bors merge

PR #389, opened 2024-03-13
convert table to sections

tshepang:
It is easy to miss that there is another column, because second column is much wider than a standard screen.

### DIFF
--- a/ferrocene/doc/safety-manual/src/constraints.rst
+++ b/ferrocene/doc/safety-manual/src/constraints.rst
@@ -11,66 +11,66 @@ Constraints
    :header-rows: 1
 
    * - Requirement ID
-     - Requirement Definition
      - Associated Requirement ID
+     - Requirement Definition
    * - .. id:: CSTR_0010_INSTALL
+     - :id:`AVD_CHECK_INSTALL_001`
      - Before using the Ferrocene toolchain, the user shall ensure that the
        appropriate tool package has been installed and installation verified
        according to the verification procedure described in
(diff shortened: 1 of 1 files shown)


Urhengulas (approved):
bors merge

pietroalbini:
bors merge-

I wonder whether we should switch from a table to just having sections and paragraphs? For example:

```rst
Installation of Ferrocene
-------------------------

.. id:: CSTR_0010_INSTALL

Associated requirements: :id:`AVD_CHECK_INSTALL_001`

Before using the Ferrocene toolchain, the user shall ensure that the
appropriate tool package has been installed and installation verified
according to the verification procedure described in
:doc:`Installation Procedures <safety-manual:procedures>`.
```

Urhengulas:
@pietroalbini said:
> I wonder whether we should switch from a table to just having sections and paragraphs?

I do like the table. But I agree it is hard to read. Can we configure it to wrap the text?

Screenshot from 2024-03-13 12-43-32

By wrap I mean that it behaves like columns A and not like column B.

pietroalbini:
@Urhengulas so, in general, I'd like this page to move away from tables because tables force the description to be just a sentence. If we had a section for each constraint we'd be able to expand the text in the future (for example adding how to check the version of tools in `CSTR_0030_BUILD_MONITORING`).

Regarding wrapping tables... it's complex :sweat_smile:. While it would work ok for small tables, it produces fairly hard to read layout in large tables. Let's move the conversation on that to Zulip.

Urhengulas (approved):
bors merge

PR #392, opened 2024-03-13
Understand `ferrocene-self-test` by refactoring it

Urhengulas:
I go through the code, try to document it and clean it up, and hope to understand it better at the end.

### DIFF
--- a/ferrocene/tools/self-test/src/binaries.rs
+++ b/ferrocene/tools/self-test/src/binaries.rs
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: The Ferrocene Developers
 
-use crate::error::Error;
-use crate::report::Reporter;
-use crate::utils::run_command;
 use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use std::process::Command;
+
+use crate::error::Error;
+use crate::report::Reporter;
+use crate::utils::run_command;
+use crate::{CFG_RELEASE, SELFTEST_CARGO_HASH, SELFTEST_RUST_HASH, SELFTEST_TARGET};
 
 /// Check that the executables exist in the expected path, and that they have the correct permissions.
 ///
@@ -92,8 +94,8 @@ fn parse_version_output(output: &str) -> Option<VersionOutput> {
 
 fn check_version(version: VersionOutput, hash: CommitHashOf, name: &str) -> Result<(), Error> {
     for (field, expected, found) in [
-        ("host", env!("SELFTEST_TARGET"), version.host),
-        ("release", env!("CFG_RELEASE"), version.release),
+        ("host", SELFTEST_TARGET, version.host),
+        ("release", CFG_RELEASE, version.release),
         ("commit hash", hash.fetch().unwrap_or("unknown"), version.commit_hash),
     ] {
         if expected != found {
@@ -138,8 +140,8 @@ enum CommitHashOf {
 impl CommitHashOf {
     fn fetch(&self) -> Option<&'static str> {
         match self {
-            CommitHashOf::Rust => option_env!("SELFTEST_RUST_HASH"),
-            CommitHashOf::Cargo => option_env!("SELFTEST_CARGO_HASH"),
+            CommitHashOf::Rust => SELFTEST_RUST_HASH,
+            CommitHashOf::Cargo => SELFTEST_CARGO_HASH,
         }
     }
 }
--- a/ferrocene/tools/self-test/src/binaries.rs
+++ b/ferrocene/tools/self-test/src/binaries.rs
@@ -46,17 +46,17 @@ fn check_file(bin: &Path, bin_dir: &Path, name: &str) -> Result<(), Error> {
     match std::fs::metadata(bin) {
         Ok(metadata) => {
             if !metadata.is_file() {
-                Err(Error::MissingBinary { directory: bin_dir.to_owned(), name: name.to_owned() })
+                Err(Error::MissingBinary { directory: bin_dir.into(), name: name.into() })
             } else if metadata.permissions().mode() & MODE != MODE {
-                Err(Error::WrongBinaryPermissions { path: bin.to_owned() })
+                Err(Error::WrongBinaryPermissions { path: bin.into() })
             } else {
                 Ok(())
             }
         }
         Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
-            Err(Error::MissingBinary { directory: bin_dir.to_owned(), name: name.to_owned() })
+            Err(Error::MissingBinary { directory: bin_dir.into(), name: name.into() })
         }
-        Err(err) => Err(Error::MetadataFetchFailed { path: bin.to_owned(), error: err }),
+        Err(err) => Err(Error::MetadataFetchFailed { path: bin.into(), error: err }),
     }
 }
 
@@ -80,9 +80,9 @@ fn parse_version_output(output: &str) -> Option<VersionOutput> {
 
         let Some((key, value)) = line.split_once(": ") else { continue };
         match key {
-            "host" => host = Some(value.to_string()),
-            "commit-hash" => commit_hash = Some(value.to_string()),
-            "release" => release = Some(value.to_string()),
+            "host" => host = Some(value.into()),
+            "commit-hash" => commit_hash = Some(value.into()),
+            "release" => release = Some(value.into()),
             _ => {}
         }
     }
--- a/ferrocene/tools/self-test/src/binaries.rs
+++ b/ferrocene/tools/self-test/src/binaries.rs
@@ -28,47 +28,47 @@ fn check_binary(
     let bin_dir = sysroot.join("bin");
     let bin = bin_dir.join(name);
 
-    match std::fs::metadata(&bin) {
-        Ok(metadata) => {
-            if !metadata.is_file() {
-                return Err(Error::MissingBinary { directory: bin_dir, name: name.into() });
-            }
-            if metadata.permissions().mode() & 0o555 != 0o555 {
-                return Err(Error::WrongBinaryPermissions { path: bin });
-            }
-        }
-        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
-            return Err(Error::MissingBinary { directory: bin_dir, name: name.into() });
-        }
-        Err(err) => return Err(Error::MetadataFetchFailed { path: bin, error: err }),
-    }
-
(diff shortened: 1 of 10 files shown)


ferrocene/tools/self-test/src/compile.rs
```diff
@@ -191,23 +193,25 @@ impl ExpectedFiles {
...
+
         let mut currently_expected = currently_expected.into_iter().collect::<Vec<_>>();
         currently_expected.sort();
-        for missing_file in currently_expected {
-            return Err(Error::MissingCompilationArtifact {
-                name: missing_file.into(),
+
+        if let Some(missing_file) = currently_expected.first() {
+            Err(Error::MissingCompilationArtifact {
+                name: missing_file.to_string(),
```
tshepang, line 209: did `.into()` fail

ferrocene/tools/self-test/src/error.rs
```diff
@@ -17,12 +17,13 @@ pub(crate) enum LinkerArgsErrorKind {
...
 
 #[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
 pub(crate) enum Error {
     NoSysroot,
     MissingBinary { directory: PathBuf, name: String },
     WrongBinaryPermissions { path: PathBuf },
     MetadataFetchFailed { path: PathBuf, error: std::io::Error },
-    VersionFetchFailed { binary: String, error: CommandError },
+    VersionFetchFailed { binary: String, error: Box<CommandError> },
```
tshepang, line 26: why is clippy suggesting this

ferrocene/tools/self-test/src/error.rs
```diff
@@ -17,12 +17,13 @@ pub(crate) enum LinkerArgsErrorKind {
 }
 
 #[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
```
tshepang, line 20: what do you think about renaming LinkerArgsError to just LinkerArgs, to avoid this allow

ferrocene/tools/self-test/src/main.rs
```diff
@@ -42,20 +48,23 @@ fn main_inner(reporter: &dyn Reporter) -> Result<(), Error> {
     let mut targets = targets::check(reporter, &sysroot)?;
     linkers::check_and_add_rustflags(reporter, &environment, &sysroot, &mut targets)?;
     compile::check(reporter, &sysroot, &targets)?;
+    report_linker_flags(reporter, &targets);
+
+    reporter.success("Ferrocene self-check completed!");
+    Ok(())
+}
 
+fn report_linker_flags(reporter: &dyn Reporter, targets: &[Target]) {
```
tshepang, line 57: was this just to improve readability of main_inner

Urhengulas: Yes. I was thinking if we can move it into the `linkers::check_and_add_rustflags`? It feels like it should go in there.

ferrocene/tools/self-test/src/compile.rs
```diff
@@ -51,6 +54,10 @@ static SAMPLE_PROGRAMS: &[SampleProgram] = &[
     },
 ];
 
+/// Compile the [`SAMPLE_PROGRAMS`] for all `targets`.
```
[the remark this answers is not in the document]
Urhengulas: Maybe:
```suggestion
/// Compile the [`SAMPLE_PROGRAMS`] for all `targets` and execute it for host target.
```

ferrocene/tools/self-test/src/compile.rs
Urhengulas: It told me I needed to specify the type for some reason

ferrocene/tools/self-test/src/error.rs
Urhengulas: It's the `result_large_err` lint: https://rust-lang.github.io/rust-clippy/master/index.html#/result_large_err

ferrocene/tools/self-test/src/compile.rs
Urhengulas: But in general I am unsure about using `into`, `to_owned` or `to_string`. And I am definitely not consistent. Probably `into` is best, since it is shortest.

tshepang: I prefer `to_owned` over `to_string`, since it is more clear what it does... the original data is already a string type.

This is not a suggestion to change this, just a nit you can ignore.

ferrocene/tools/self-test/src/main.rs
tshepang: seems like a perfect place to do it

ferrocene/tools/self-test/src/binaries.rs
```diff
@@ -25,15 +25,18 @@ fn check_binary(
...
     let bin = bin_dir.join(name);
 
+    // check that `bin` is a file and has the correct permissions
     match std::fs::metadata(&bin) {
         Ok(metadata) => {
             if !metadata.is_file() {
                 return Err(Error::MissingBinary { directory: bin_dir, name: name.into() });
-            }
-            if metadata.permissions().mode() & 0o555 != 0o555 {
+            } else if metadata.permissions().mode() & MODE != MODE {
```
tshepang, line 48: I would not else here, since the if block above returns from the function

ferrocene/tools/self-test/src/binaries.rs
```diff
@@ -25,15 +25,18 @@ fn check_binary(
     name: &str,
     hash: CommitHashOf,
 ) -> Result<(), Error> {
+    /// File permission the binary should have. The symbolic value is `r-xr-xr-x`.`
```
tshepang: not exactly correct, since these are just the minimum required permissions... `0o777` would also work

ferrocene/tools/self-test/src/utils.rs
```diff
@@ -61,28 +61,6 @@ pub(crate) fn find_binary_in_path(
     Err(FindBinaryInPathError::MissingBinary { name: name.into() })
 }
 
-pub(crate) struct DisplayList<'a, T: Display>(pub(crate) &'a [T]);
```
tshepang, line 64: looks like it's used in tests

ferrocene/tools/self-test/src/binaries.rs
```diff
@@ -83,9 +87,9 @@ fn parse_version_output(output: &str) -> Option<VersionOutput<'_>> {
 
         let Some((key, value)) = line.split_once(": ") else { continue };
         match key {
-            "host" => host = Some(value),
-            "commit-hash" => commit_hash = Some(value),
-            "release" => release = Some(value),
+            "host" => host = Some(value.to_string()),
+            "commit-hash" => commit_hash = Some(value.to_string()),
+            "release" => release = Some(value.to_string()),
```
tshepang, line 83: nit: `.into()` should work here, I think

ferrocene/tools/self-test/src/binaries.rs
Urhengulas: Why? Both blocks do return from the function. It can only ever be one of them, never both.

ferrocene/tools/self-test/src/binaries.rs
Urhengulas: To be honest I do not really understand how the check works.

```rust
metadata.permissions().mode() & MODE != MODE
```

Can you explain?

ferrocene/tools/self-test/src/binaries.rs
tshepang: the `else` is redundant because it does not change the logic, since a `return` means the execution cannot continue anyways

ferrocene/tools/self-test/src/binaries.rs
tshepang: the check ensures that at least the file is both readable and executable...
```
0o5 = 0b101 (r-x)
         ^ we do not care whether or not this, the write bit, is set
0b101 & 0b111 = 0b101
0b101 & 0b101 = 0b101
0b101 & 0b110 = 0b100 (fail, not executable)
0b101 & 0b011 = 0b001 (fail, not readable)

File: ferrocene/tools/self-test/src/binaries.rs (hunk 6 above)
Comment (Urhengulas) [reply ferrocene/tools/self-test/src/binaries.rs:48]: I know. But the else makes it more compact imo. And also it enables us to remove the `return` because we also move it into a function (in a later commit).

File: ferrocene/tools/self-test/src/binaries.rs (hunk 7 above)
Comment (Urhengulas) [reply ferrocene/tools/self-test/src/binaries.rs]: Thank you! (I also changed the doc-comment).

File: ferrocene/tools/self-test/src/binaries.rs (hunk 9 above)
Comment (Urhengulas) [reply ferrocene/tools/self-test/src/binaries.rs]: I used `into` whereever possible

File: ferrocene/tools/self-test/src/binaries.rs (hunk 6 above)
Comment (tshepang) [reply ferrocene/tools/self-test/src/binaries.rs:50]: me fan of removing the `return`, so 👍

File: ferrocene/tools/self-test/src/test_utils.rs [hunk 10]
```diff
@@ -233,15 +233,10 @@ impl Default for CliVersionContent<'_> {
...
-        buffer.push_str(self.release);
-        buffer.push_str("\nhost: ");
-        buffer.push_str(self.host);
-        buffer.push_str("\ncommit-hash: ");
-        buffer.push_str(self.commit_hash);
-        buffer.push('\n');
-        buffer
+        format!(
+            "release: {}\nhost: {}\ncommit-hash: {}\n",
+            self.release, self.host, self.commit_hash
```
tshepang, line 238: would be nice to get this looking close to how it will look when executed (like was done in [parent commit](https://github.com/ferrocene/ferrocene/pull/392/commits/feb1dd1a897252d39fb45df7a1072cc3e4023d80#diff-76f740a27d65cacdf8abe8a57b7ede0e21ea3482062d08c992aa787b2298b5b1R129-R132))

tshepang:
I think that the last commit is not worth it
- it's a trivial alias, so not such a useful indirection
- `Error` is brought into the namespace everywhere, so it's good to be consistent and make it visible in `Result` type as well

ferrocene/tools/self-test/src/test_utils.rs
Urhengulas: Changed it.

Urhengulas:
> I think that the last commit is not worth it
> 
>     * it's a trivial alias, so not such a useful indirection
> 
>     * `Error` is brought into the namespace everywhere, so it's good to be consistent and make it visible in `Result` type as well

Dropped it.

tshepang (approved):
bors merge

PR #403, opened 2024-03-15
Prevent the release job from failing when no release should be published

pietroalbini:
This should remove the last reason why the release job is red with no actual failure. It currently happens on weekends when there are no new releases to publish.


Hoverbear (approved):

Hoverbear:
bors merge

PR #404, opened 2024-03-15
Remove channel sphinx substitution

pietroalbini:
Turns out we can't inject the channel name in our documentation! Our document signature scheme depends on the content of the signature not to change between the beta and stable branch (as we need to keep the same document IDs and signatures between the two), while injecting the channel implies the signatures would be invalidated.

Figured this out while signing documents with Florian.


Veykril (approved):

Veykril:
bors merge

PR #405, opened 2024-03-16
be consistent with other text

tshepang:
This is how it's spelled in the landing page of Ferrocene project.


pietroalbini (approved):
bors merge

PR #419, opened 2024-03-19
properly detect when bootstrap can't be invoked in pull-upstream

pietroalbini:
Should fix the current pulls from failing (including beta).


ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -235,9 +235,17 @@ for prefix in "${DIRECTORIES_CONTAINING_LOCKFILES[@]}"; do
     fi
 done
 
+# Check whether we can compile bootstrap successfully, which will be used to
+# gate on the next few steps.
+echo "pull-upstream: checking whether bootstrap can be invoked safely..."
+if ./x.py; then
```
tshepang, line 241: this always fails, because it's not a valid way of invoking `./x.py`

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -235,9 +235,17 @@ for prefix in "${DIRECTORIES_CONTAINING_LOCKFILES[@]}"; do
...
+    can_invoke_bootstrap=true
+else
+    can_invoke_bootstrap=false
+fi
+
 # Only run these steps when there are no conflicts in bootstrap, otherwise
```
tshepang, line 247: this comment must also go, or be updated

pietroalbini:
Addressed review comments.

tshepang (approved):
bors merge

PR #426, opened 2024-03-20
doc fixes

tshepang:


pietroalbini:
Let's not backport this, it's mostly typos/wording and we should only make actual content changes on a signed release.

PR #428, opened 2024-03-20
Bump to cosign 2.2.3

pietroalbini:
Sigstore updated their TUF root in a way that is not compatible with cosign 2.2.3 anymore. See https://blog.sigstore.dev/tuf-root-update/ for more information. This PR bumps the pinned version of cosign we use.


tshepang (approved):
bors merge

PR #429, opened 2024-03-20
Fix typo in github actions workflow

pietroalbini:
Since https://github.com/ferrocene/ferrocene/pull/403 the release workflow started failing with an internal GitHub error. Looking at the change, I noticed a typo in the configuration. Hopefully this will fix it, otherwise we'll have to contact GitHub support.


tshepang (approved):
bors merge

PR #431, opened 2024-03-20
Distribute `flip-link` with Ferrocene

Urhengulas:
The end goal of this PR is to distribute `flip-link` (https://github.com/knurling-rs/flip-link) with Ferrocene releases.

Part of https://ferroussystems.clickup.com/t/86943jgtj.

### Steps to get there:
- [x] `./x build flip-link`
- [x] `./x dist flip-link`
- [x] Add to CI
- [x] Sign flip-link --> that seems to happen automatically
- [x] Test it with a manual release


Urhengulas:
"Perform Licensing Checks" fails for flip-link:
Screenshot from 2024-03-20 23-46-55

Urhengulas:
bors try

Urhengulas:
bors try

tshepang:
can we clean the history... self-test appears in a few commits for example

tshepang:
am also thinking we should use a git subtree (for fliplink repo), instead of a git submodule... I hear they are more easy to work with, and it would be more consistent with our other repos (like ferrocene/specification)

Urhengulas:
> am also thinking we should use a git subtree (for fliplink repo), instead of a git submodule... I hear they are more easy to work with, and it would be more consistent with our other repos (like ferrocene/specification)

As far as I understand subtrees have the advantage that it is easier to make changes on both sides, while for submodules you can only really pull in new changes, but don’t push them.

But we actually do not want to make changes to flip-link from the ferrocene repo, so I think submodules are fine.

tshepang:
when doing a history re-write, rather not rebase against main

- makes re-reviewing quicker, since one can just click on Compare, instead of doing the whole review again
- bors does a rebase automatically before merging

for the decision on subree vs submodule, would be good to have @Hoverbear or @pietroalbini comment

src/bootstrap/src/ferrocene/tool/flip_link.rs
tshepang, line 1: I see other Ferrocene tools are in tool.rs... why is this one separate

src/bootstrap/src/ferrocene/dist/flip_link.rs
tshepang, line 1: I see other Ferrocene dist things are in dist.rs... why is this one separate

pietroalbini:
I think it's fine to use a submodule for now, as I don't think we want to do Ferrocene-specific changes to it yet. We can always turn it into a subtree if the need arises. The only concern I have is, we don't have infrastructure to keep it up to date yet. Dependabot should do the trick, [like we did in the spec repo](https://github.com/ferrocene/specification/blob/main/.github/dependabot.yml), but we'll need to [filter it to only update that submodule (and not the ones managed by upstream)](https://stackoverflow.com/a/75078947).

Keeping the dist steps in their own module is actually great, I wanted to split the huge `test.rs`/`dist.rs`/etc into separate files for ages but never had the time to do it.

pietroalbini:
Also, I see you need to update `ferrocene/packages.toml` too.

Urhengulas:
@pietroalbini said:
> I wanted to split the huge `test.rs`/`dist.rs`/etc into separate files for ages but never had the time to do it.

I will do that in a follow-up PR. Didn't want to clutter this one with unrelated changes.

src/bootstrap/src/ferrocene/dist/flip_link.rs
Urhengulas: Same as above.

Urhengulas:
> when doing a history re-write, rather not rebase against main
> 
>     * makes re-reviewing quicker, since one can just click on Compare, instead of doing the whole review again
> 
>     * bors does a rebase automatically before merging

Sorryy. I usually do `git rebase -i main` because it's easy for me. Otherwise I'd need to look up the commit the PR branched off from the main branch. But I see why it is annoying for reviwer.

Urhengulas:
@pietroalbini said:
> The only concern I have is, we don't have infrastructure to keep it up to date yet. Dependabot should do the trick, [like we did in the spec repo](https://github.com/ferrocene/specification/blob/main/.github/dependabot.yml), but we'll need to [filter it to only update that submodule (and not the ones managed by upstream)](https://stackoverflow.com/a/75078947).

Can I do this in a follow-up PR?

Urhengulas:
bors try

pietroalbini:
> Can I do this in a follow-up PR?

Sure!

Urhengulas:
The manual release worked!

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -9,11 +17,8 @@ use crate::ferrocene::doc::ensure_all_xml_doctrees;
 use crate::ferrocene::test_outcomes::TestOutcomesDir;
 use crate::t;
 use crate::utils::tarball::{GeneratedTarball, Tarball};
-use std::collections::BTreeMap;
-use std::fs;
-use std::path::{Path, PathBuf};
-use std::process::Command;
-use std::rc::Rc;
+
+pub(crate) use self::flip_link::FlipLink;
```
pietroalbini, line 21: Nit: rather than re-exporting I'd make the module `pub(crate)`. That way it's less maintenance when we add a new item to modules (we just have to add it to the list of steps in `core/builder.rs` rather than also `ferrocene/parent.rs`).

src/bootstrap/src/ferrocene/dist/flip_link.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    fn make_run(run: RunConfig<'_>) {
+        run.builder.ensure(FlipLink { target: run.target });
+    }
+
+    fn run(self, builder: &Builder<'_>) -> Self::Output {
+        let flip_link = builder.ensure(crate::ferrocene::tool::FlipLink { target: self.target });
+
+        let tarball = Tarball::new(builder, "flip-link", &self.target.triple);
+        tarball.add_file(flip_link, "bin", 0o755);
```
pietroalbini, line 27: Should flip-link be something users invoke directly? If so, we should register it as a binary proxy.

Urhengulas: Not directly, but they would configure the compiler to invoke it. See https://github.com/knurling-rs/flip-link?tab=readme-ov-file#usage.

```toml
# .cargo/config.toml

[target.'cfg(all(target_arch = "arm", target_os = "none"))']
# (..)
linker = "flip-link"

# OR

[target.'cfg(all(target_arch = "arm", target_os = "none"))']
# (..)
rustflags = [
  "-C", "linker=flip-link",
  # (..)
]
```

src/bootstrap/src/ferrocene/dist.rs
Urhengulas: Done

src/bootstrap/src/ferrocene/dist/flip_link.rs
Urhengulas: So should we add it? (not exactly sure what a binary proxy is tbh)

Urhengulas: Added it

pietroalbini (approved):
bors merge

PR #436, opened 2024-03-21
Refactor `ferrocene-self-test` more

Urhengulas:
Apart from some small cleanups, the highlights of this refactor are:
- Combining all environment variable related stuff in the `env` module.
- Using `thiserror` to save 50+ lines of code.
- ~~Use `pub` instead of `pub(crate)`, because that's the same in binary crates.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1862,6 +1862,7 @@ dependencies = [
  "atty",
  "insta",
  "tempfile",
+ "thiserror",
 ]
 
 [[package]]
--- a/ferrocene/tools/self-test/Cargo.toml
+++ b/ferrocene/tools/self-test/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 [dependencies]
 atty = "0.2.14"
 tempfile = "3.3.0"
+thiserror = "1"
 
 [dev-dependencies]
 insta = { version = "1.31.0" }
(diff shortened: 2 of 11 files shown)


Hoverbear:
`pub` and `pub(crate)` are unfortunately not equivalent.

Using `pub`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6feabb3059b241f41bd5f5e76dec80f

Using `pub(crate)`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=300cc9e5ad60eb58b64be474b0de43c3

Notice the warning about dead code? That's actually quite useful at times, and often desirable.

tshepang:
other than reverting the `pub(crate)` removal commit (thanks @Hoverbear), this is good to go... I especially appreciate moving env things to own module, because I found it surprising that being defined in main.rs made it accessible elsewhere

Urhengulas:
@Hoverbear said:
> `pub` and `pub(crate)` are unfortunately not equivalent.
> 
> Using `pub`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6feabb3059b241f41bd5f5e76dec80f
> 
> Using `pub(crate)`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=300cc9e5ad60eb58b64be474b0de43c3
> 
> Notice the warning about dead code? That's actually quite useful at times, and often desirable.

I did not know! Thank you for pointing out.

I'd argue that Rust should still warn about dead code for `pub` items _if it is a binary crate_. But that's not the behavior now, so let's revert it.

PR #457, opened 2024-03-27
CI: Update `actions/checkout` to `v4`

Urhengulas:
This should silence the warning:

Screenshot from 2024-03-27 17-08-46

### DIFF
--- a/.github/workflows/automation-backport.yml
+++ b/.github/workflows/automation-backport.yml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: Checkout the Ferrocene repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Fetch all the history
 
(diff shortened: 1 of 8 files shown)


ferrocene/doc/specification/.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     # 2022-11-24. See https://github.com/actions/setup-python/issues/544.
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
tshepang, line 24: this one should be done in ferrocene/specification

Urhengulas: Good point.

But just trying to wrap my head around it: Since it is a subtree, couldn't we also make the change here and push it to ferrocene/specification?

tshepang: that can be, but we don't have automation for that (and we don't want to have to remember to do the push)

tshepang:
can you squash commits, just in case upstream tree gets confused by what happened here

pietroalbini (approved):
bors merge

PR #458, opened 2024-03-27
Minor - doc: add aws-profile key to to config.example.toml

amanjeev:
I think adding public doc link to the key section in example config is good but please correct me if I am mistaken.


tshepang (approved):
bors merge

PR #468, opened 2024-04-02
Prevent releasing commits not authored by bors

pietroalbini:
Johann discovered that if you try to start a manual release with a commit that is not authored by bors and built by CI, you'll get a confusing `AccessDenied` error. To alleviate the problem, this PR checks for the common failure case (someone putting their development branch rather than a bors try commit) and shows a descriptive error in that case.


Urhengulas:
bors merge

Urhengulas (approved):
bors merge

PR #476, opened 2024-04-03
avoid confusion and simplify

tshepang:
- This could lead people to think the behavior is different from upstream.
- "rustc" is short for Rust compiler... we need not add another "compiler"

### DIFF
--- a/ferrocene/doc/user-manual/src/using-compiler/executable.rst
+++ b/ferrocene/doc/user-manual/src/using-compiler/executable.rst
@@ -4,8 +4,7 @@
 Building an executable
 ======================
 
-This chapter describes how to build an executable using Ferrocene's rustc
-compiler.
+This chapter describes how to build an executable using rustc.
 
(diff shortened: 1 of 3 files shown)


Urhengulas:
Could this be a "trademark" problem? Because `rustc` refers to the "official" upstream compiler. And we build and ship it separately.

And we potentially modify it, but as far as I am aware we didn't make any changes to the compiler, just build system, right?

tshepang:
I don't think "rustc" is trademarked, and also was thinking we should create a glossary entry defining what is meant by that term.

As for what we modified, I only remember that we have one target that upstream doesn't have. Beyond that, it's indeed the build system that is different... we use a different linker and build options.

Urhengulas (approved):

Urhengulas:
bors merge

Urhengulas:
> we should create a glossary entry defining what is meant by that term

We should!

PR #479, opened 2024-04-04
Fix "last commit detection" in mirror-sphinx-shared-resources automation

Urhengulas:
`str.lstrip` is too hungry[^1], therefore `str.removeprefix`[^2] will take over from now on and hopefully behave better.

`str.removeprefix` is new in Python 3.9, which is okay since the GitHub Actions Runner contains Python 3.10[^3].

[^1]: It ate a leading `e` of a commit hash 😱 
[^2]: https://docs.python.org/3/library/stdtypes.html#str.removeprefix
[^3]: https://github.com/actions/runner-images/blob/069e7c963c42992e601f64880bccd98681b953dc/images/ubuntu/Ubuntu2204-Readme.md?plain=1#L27


tshepang (approved):
bors merge

PR #486, opened 2024-04-08
Refactor `ferrocene-self-test`

Urhengulas:
Changes
- move BinBuilder to own module
- reduce repetition in target tests
- use `TestUtils` in `test_find_libraries_in`
- some drive-by changes
- please double check the binary condition in `extract_library_name` is still correct (I inverted it to simplify the code)

### DIFF
--- a/ferrocene/tools/self-test/src/targets.rs
+++ b/ferrocene/tools/self-test/src/targets.rs
@@ -130,18 +130,15 @@ fn find_libraries_in(path: &Path) -> Result<HashMap<String, usize>, Error> {
 
 /// Extract `name` from `libname-hash.rlib`.
 fn extract_library_name(file_path: &Path) -> Option<&str> {
-    let (library, hash) = file_path
+    let (library_name, hash) = file_path
         .file_name()?
         .to_str()?
         .strip_prefix("lib")?
         .strip_suffix(".rlib")?
         .rsplit_once('-')?;
 
-    if hash.len() != 16 || hash.chars().any(|c| !c.is_ascii_hexdigit()) || library.is_empty() {
-        None
-    } else {
-        Some(library)
-    }
+    (hash.len() == 16 && hash.chars().all(|c| c.is_ascii_hexdigit()) && !library_name.is_empty())
+        .then_some(library_name)
 }
 
 #[cfg(test)]
--- a/ferrocene/tools/self-test/src/targets.rs
+++ b/ferrocene/tools/self-test/src/targets.rs
@@ -70,15 +70,11 @@ fn check_target(
         return Ok(CheckTargetOutcome::Missing);
     }
 
-    check_libraries(
-        target,
-        &target_dir,
-        if target.std {
-            &["core", "alloc", "std", "test", "proc_macro"]
-        } else {
-            &["core", "alloc"]
-        },
-    )?;
+    let expected_libs = match target.std {
+        true => ["core", "alloc", "std", "test", "proc_macro"].as_slice(),
+        false => ["core", "alloc"].as_slice(),
+    };
+    check_libraries(target, &target_dir, expected_libs)?;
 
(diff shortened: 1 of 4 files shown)


ferrocene/tools/self-test/rust-toolchain.toml
tshepang, line 1: why add this

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -256,17 +256,21 @@ mod tests {
...
+            .create();
 
-        let create_lib = |name| std::fs::write(dir.join(name), b"").unwrap();
-        create_lib("libcore-0123456789abcdef.rlib");
-        create_lib("libcore-abcdef0123456789.rlib");
-        create_lib("liballoc-0123456789abcdef.rlib");
-        create_lib("libproc_macro-0123456789abcdef.rlib");
-        create_lib("foo-0123456789abcdef.so"); // Invalid files are not counted.
+        let lib_dir = utils.target_dir(triple).join("lib");
+        std::fs::write(lib_dir.join("foo-0123456789abcdef.so"), b"").unwrap(); // Invalid files are not counted.
```
tshepang, line 267: would a `utils.lib("foo", "..")` (in line 269) also work

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -70,15 +70,11 @@ fn check_target(
...
-            &["core", "alloc", "std", "test", "proc_macro"]
-        } else {
-            &["core", "alloc"]
-        },
-    )?;
+    let expected_libs = match target.std {
+        true => ["core", "alloc", "std", "test", "proc_macro"].as_slice(),
+        false => ["core", "alloc"].as_slice(),
+    };
+    check_libraries(target, &target_dir, expected_libs)?;
```
tshepang, line 77: does this work
```suggestion
    let expected_libs = match target.std {
        true => ["core", "alloc", "std", "test", "proc_macro"],
        false => ["core", "alloc"],
    };
    check_libraries(target, &target_dir, &expected_libs)?;
```

ferrocene/tools/self-test/rust-toolchain.toml
Veykril: Smell like a setup mistake somewhere, you should be using the bootstrap toolchain and nothing else here

Urhengulas: Ups, didn't want to commit it

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -267,10 +267,11 @@ mod tests {
         create_lib("foo-0123456789abcdef.so"); // Invalid files are not counted.
 
         let output = find_libraries_in(dir).unwrap();
-        assert_eq!(3, output.len());
-        assert_eq!(2, output["core"]);
-        assert_eq!(1, output["alloc"]);
-        assert_eq!(1, output["proc_macro"]);
+        assert_eq!(output.len(), 3);
+        assert_eq!(output.get("core"), Some(&2));
```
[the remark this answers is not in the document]
Urhengulas: Mostly to be consistent with the last I added (`assert_eq!(output.get("other"), None`).

ferrocene/tools/self-test/src/targets.rs
Urhengulas: No. Because here we are specifically adding a `.so` library. But `.lib` does add an `.rlib`. To goal is to test that non-rlib libraries are ignored.

ferrocene/tools/self-test/src/targets.rs
Urhengulas: I _think_ this does not compile, because the type of `expected_libs` is unclear, because the types in the both arms are conflicting . Is it `[&str; 5]` or is it `[&str; 2]`?

tshepang (approved):
bors merge

PR #509, opened 2024-04-12
Include `.gitmodules` in the source tarball

pietroalbini:
A recent-ish change in bootstrap (not sure exactly which PR) started requiring the .gitmodules file to be included in the source tarball, otherwise bootstrap will not start.

The upstream source tarball does include that file, so this commit replicates the change in Ferrocene too.


Hoverbear (approved):

Hoverbear:
bors merge

PR #536, opened 2024-04-19
Build x86_64 Darwin too

Hoverbear:


Hoverbear:
bors try

pietroalbini (changes requested):
I would not make the `aarch64-darwin` builders build x86_64 as well.  That would result in those builders being way slower, as they need to build two copies of the compiler and the tools.

Rather, I'd create a separate set of `x86_64-darwin` builders running on Intel macs. Since that might take a bit of time to implement, for now I'd remove the x86_64 variant from `packages.toml`.

Hoverbear:
Ok!

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -98,15 +98,15 @@ should not be used in production.
 
    * - :target:`x86_64-apple-darwin`
      - ``x86_64-apple-darwin``
-     - Host platform
+     - Cross-compilation
      - Full
-     - Also available as a cross-compile target for :target:`aarch64-apple-darwin`.
+     - \-
```
pietroalbini, line 103: We need to remove this then?

pietroalbini (changes requested):

ferrocene/doc/user-manual/src/targets/index.rst
Hoverbear: I believe our aarch64 host can build x86_64 darwin targets.

Hoverbear: image
https://app.circleci.com/pipelines/github/ferrocene/ferrocene/3394/workflows/43ad5931-637b-41c6-baf5-f4ac9a5ac20f/jobs/16291?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

pietroalbini (approved):
bors merge

PR #542, opened 2024-04-22
Improve vendoring

pietroalbini:
This PR improves vendoring of dependencies in the source tarball by:

* Vendoring dependencies for `cargo` and `rustc_codegen_gcc` too, which are outside of the workspace.
* Using the correct path when remapping vendored dependencies, as we store our vendored crates in `vendor/rust` rather than just `vendor` (to support vendoring of other languages in the future).
* Passes `--versioned-dirs` to match upstream.


pietroalbini:
bors merge

PR #544, opened 2024-04-22
Experimental QNX Support

Hoverbear:
Introduces experimental QNX support.

## Open Problems

* Moving beyond QNX 7.1.0 we likely won't see Mac versions, so we should consider how to deal with that.


.circleci/workflows.yml
```diff
@@ -159,11 +159,12 @@ jobs:
     resource_class: large # 4-core
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
-      FERROCENE_TARGETS: aarch64-unknown-linux-gnu,aarch64-unknown-none,thumbv7em-none-eabi,thumbv7em-none-eabihf,armv8r-none-eabihf,wasm32-unknown-unknown,armv7r-none-eabihf,armebv7r-none-eabihf
+      FERROCENE_TARGETS: aarch64-unknown-linux-gnu,aarch64-unknown-none,thumbv7em-none-eabi,thumbv7em-none-eabihf,armv8r-none-eabihf,wasm32-unknown-unknown,armv7r-none-eabihf,armebv7r-none-eabihf,aarch64-unknown-nto-qnx710,x86_64-pc-nto-qnx710
```
Hoverbear, line 162: Noting I still need to add the self-test bits but also that requires more circle.yaml changes since I need to get the QNX toolchain there, too.

Hoverbear:
bors try

.circleci/workflows.yml
```diff
@@ -681,6 +685,20 @@ commands:
           steps:
             - ferrocene-setup-darwin
       - aws-oidc-auth
+      - when:
+          condition:
+            equal: [<<parameters.qnx>>, true]
+          steps:
+            - run:
+                name: Fetch QNX710 toolchain
+                command: aws s3 cp s3://ferrocene-ci-artifacts/qnx/qnx710.tar.zst - | zstd -d | tar x
```
pietroalbini, line 694: Can we store this in `s3://ferrocene-ci-mirrros/manual`? That's where everything else we mirrored is stored.

.circleci/workflows.yml
```diff
@@ -681,6 +685,20 @@ commands:
...
+            - run:
+                name: Set PATH to include QNX710 tools
+                command: echo 'source $(pwd)/qnx710/qnxsdp-env.sh' >> "$BASH_ENV"
+            - run:
+                name: Start QNX License Server port forwarding
+                command: aws ssm start-session --target i-0accef99ca2244097 --document-name AWS-StartPortForwardingSession --parameters '{"portNumber":["56859"], "localPortNumber":["56859"]}'
```
pietroalbini, line 700: I'd add a tag (probably `Name`?) to the instance [and first search the instance by tag](https://serverfault.com/a/560534) to get its ID. That way, if we ever need to recreate the instance for whatever reason we don't need to go and manually change the CI configuration.

pietroalbini (commented):
The permission error in CI is due to the policy don't allowing access to the whole `s3://ferrocene-ci-artifacts`, but just `s3://ferrocene-ci-artifacts/ferrocene`. I left a comment for where it should be stored.

Also, when adding the target in `packages.toml` please add it with the `qnx` subset rather than the `default` subset, as we want to gate access to it separately.

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -22,6 +22,10 @@ static SUPPORTED_TARGETS: &[TargetSpec] = &[
     TargetSpec { triple: "thumbv7em-none-eabi", std: false, linker: Linker::BundledLld },
     TargetSpec { triple: "armv7r-none-eabihf", std: false, linker: Linker::BundledLld },
     TargetSpec { triple: "armebv7r-none-eabihf", std: false, linker: Linker::BundledLld },
+    TargetSpec { triple: "armv7r-none-eabihf", std: false, linker: Linker::BundledLld },
+    TargetSpec { triple: "armebv7r-none-eabihf", std: false, linker: Linker::BundledLld },
```
Hoverbear, line 26: Oops

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try-

Hoverbear:
bors try

.circleci/workflows.yml
Hoverbear: Given that `Describe*` style actions [can't be limited to specific resources](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-policy-structure.html#ec2-supported-iam-actions-resources) I don't want to do that since it would require permitting the CI token to enumerate all the ec2 nodes of our infrastructure. (It might seem harmless now, but I don't think it's a wise road to travel down)

Perhaps we could have terraform [provision a file to the bucket with the QNX artifacts](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object) which gets recreated if the node changes?

Hoverbear:
bors try

How about this? It **should** fail since you can't actually create the SSM session and thus can't get the license.

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

I think I fixed the permissions...

Hoverbear:
bors try

I knew I forgot something!

Hoverbear:
bors try

I can't believe it, I forgot to set QNXLM_LICENSE_FILE after all that. 😅

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors merge

Hoverbear:
bors try

ferrocene/doc/internal-procedures/src/partners/qnx.rst
```diff
@@ -0,0 +1,206 @@
...
+<https://blackberry.qnx.com/>`_, an embedded OS designed for automotive use.
+This page contains the information relevant to Ferrous Systems employees
+working on Ferrocene.
+
+.. note::
+    
+    QNX only supports x86-64 Linux (glibc) and x86-64 Windows.
+    
+    QNX does not support Apple Silicon macOS. QNX is deprecating support
+    for x86-64 macOS.
```
pietroalbini, line 17: Very small nit: use `:target:` to refer to the targets?

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -102,6 +102,18 @@ should not be used in production.
      - Bare-metal
      - \-
 
+   * - :target:`aarch64-unknown-nto-qnx710`
+     - ``aarch64-unknown-nto-qnx710``
+     - Cross-compilation
+     - Full
+     - Only available on targets supported by QNX (:ref:`x86_64-unknown-linux-gnu` and :ref:`x86_64-pc-windows-msvc`)
```
pietroalbini, line 109: I'd write it like "Can only be cross-compiled from :ref:`x86_64-unknown-linux-gnu` and :ref:`x86_64-pc-windows-msvc`"

pietroalbini (commented):
Minor docs nits, otherwise looks good!

Hoverbear:
bors try

Hoverbear:
I updated our deployment image to remove armv7le support and the docs. That should reduce the number of files Windows struggles with.

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
I removed the QNX windows self-test and updated the docs to note this for now. I'll circle back around to try to fix 43+ minute tarball unpack on Windows. I'm not sure how to resolve it as it seems to be a simple case of "No IOPS"

Hoverbear:
bors try

pietroalbini (approved):
bors merge

PR #549, opened 2024-04-23
Prevent excluded steps from being executed at all

pietroalbini:
Before this PR, whenever we needed to exclude a step from CI because it didn't make sense for Ferrocene, we changed the split-tasks script to always `--exclude` it. That was a fine approach for CI, as the script controlled what we executed.

We have customers building from source now though, and they don't use the script (nor should they). This PR thus moves to a different way of permanently excluding those jobs: changing bootstrap to prevent executing the jobs at all.

### DIFF
--- a/ferrocene/ci/split-tasks.py
+++ b/ferrocene/ci/split-tasks.py
@@ -55,6 +55,18 @@ Job: TypeAlias = list[str]
 Kind: TypeAlias = dict[str, Job]
 JobsDefinition: TypeAlias = dict[str, Kind]
 
+# If you are trying to exclude a step from being executed at all in our CI,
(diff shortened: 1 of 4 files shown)


Hoverbear (approved):
I bumped into this a few times trying to run 

```
./x.py --stage 2 dist ...(python ferrocene/ci/split-tasks.py dist | split row " ") ...(python ferrocene/ci/split-tasks.py dist:tools | split row " ")
```

So this is a good change!

tshepang:
bors merge

Hoverbear (approved):

pietroalbini:
bors merge

PR #552, opened 2024-04-23
remove versioning from ferrocene tools

tshepang:
These are either internal tools,
or are to be distributed with the toolchain itself.


Dajamante (approved):
bors merge

PR #570, opened 2024-04-26
Refactor `ferrocene-self-test`

Urhengulas:


ferrocene/tools/self-test/src/env.rs
```diff
@@ -25,11 +27,19 @@ pub(crate) struct Env {
...
 }
 
 impl Env {
-    pub(crate) fn gather() -> Self {
-        Self { path: std::env::var_os("PATH") }
+    pub(crate) fn gather() -> Result<Self, Error> {
+        Ok(Self {
+            path: std::env::var_os("PATH").ok_or_else(|| Error::CCompilerNotFound {
+                error: crate::error::FindBinaryInPathError::NoEnvironmentVariable,
+            })?,
```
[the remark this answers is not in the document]
Urhengulas: Well, actually semantics do change, but I think it is better. I unset `PATH` and it behaves like following:

```console
$ # before
$ ./bin/ferrocene-self-test 
    Info: using sysroot /home/urhengulas/Documents/ferrocene
 Success: binary cargo is valid
 Success: target installed correctly: x86_64-unknown-linux-gnu
 Success: target installed correctly: armv8r-none-eabihf
 Success: bundled linker detected
 Success: bundled linker-wrapper detected
   Error: unable to find LLD-compatible C compiler for `x86_64-unknown-linux-gnu`
    Note: the error code is FST_023
$ # after
$ ./bin/ferrocene-self-test2
   Error: C compiler not found on the system
   Cause: the PATH environment variable is not set
    Note: the error code is FST_011
```

The first one does not even mention `PATH`.

Hoverbear (approved):
Seems a nice overall improvement.

tshepang:
`x test ferrocene/tools/self-test` fails

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -52,6 +52,15 @@ impl Deref for Target {
     }
 }
 
+impl TargetSpec {
+    fn expected_libs(&self) -> &[&str] {
```
tshepang, line 56: me would keep this inlined
- there is only one user
- is trivial

Urhengulas: Reverted it

Urhengulas:
> `x test ferrocene/tools/self-test` fails

Fixed the tests

ferrocene/tools/self-test/src/env.rs
```diff
@@ -18,18 +20,55 @@ pub(crate) const SELFTEST_RUST_HASH: Option<&str> = option_env!("SELFTEST_RUST_H
...
 }
 
 impl Env {
-    pub(crate) fn gather() -> Self {
-        Self { path: std::env::var_os("PATH") }
+    pub(crate) fn gather() -> Result<Self, Error> {
+        Self::new(
+            std::env::var_os("PATH"),
+            std::env::var("FST_PRINT_DETAILED_ARGS").ok(),
+            std::env::var("FST_PRINT_DETAILED_ERRORS").ok(),
```
pietroalbini, line 41: We should just store a boolean of whether the variable is present or not, as that's all the only way we are using those variables.

Urhengulas: Done

tshepang:
@Urhengulas can you remove the revert commit from history... keep it clean

tshepang:
would also be kool to have the tests pass for each commit that would make them fail (instead of fix them in a later commit)

tshepang (approved):
bors merge

PR #377, opened 2024-03-08
Experimental Windows support

Hoverbear:
Adds Windows support to Ferrocene as an experimental (unqualified) target.

Mostly based on the `pa-windows-ci` branch.

PR #571, opened 2024-04-26
Experimental Windows Support

Hoverbear:
> A version of https://github.com/ferrocene/ferrocene/pull/377 with less bors spam.

Adds Windows support to Ferrocene as an experimental (unqualified) target.

Mostly based on the `pa-windows-ci` branch.

The timing of split vs unsplit is basically the same.

Split: https://github.com/ferrocene/ferrocene/runs/24268639528 2h 49m 59s
Unsplit: https://github.com/ferrocene/ferrocene/runs/24037881731 2h 41m 39s

If we could get the `split` jobs tarballing faster the overall timing would be even closer...

image

image

But... I'm not convinced it's better? It still ends up being more expensive.


Hoverbear:
bors try

ferrocene/ci/scripts/llvm-cache.sh
```diff
@@ -98,29 +113,35 @@ build_llvm_tarball() {
...
+            if [[ "${OSTYPE}" = "msys" ]]; then
+                if [[ "${name}" == "${wanted}.exe" ]]; then
+                    keep="yes"
+                    break
+                fi
+            else
+                if [[ "${name}" == "${wanted}" ]]; then
+                    keep="yes"
+                    break
+                fi
```
pietroalbini, line 128: To reduce duplication, we could do something like:

```bash
exe_suffix=""
if [[ "${OSTYPE}" = "msys" ]]; then
    exe_suffix=".exe"
fi
for wanted in "${KEEP_LLVM_BINARIES[@]}"; do
    if [[ "${name}" == "${wanted}${exe_suffx}" ]]; then
        keep="yes"
        break
    fi
done
```

ferrocene/ci/scripts/persist-between-jobs.sh
```diff
@@ -47,10 +54,19 @@ case "$1" in
...
+                --exclude build/x86_64-pc-windows-msvc/stage0-sysroot/lib/rustlib/rustc-src/rust --exclude build/x86_64-pc-windows-msvc/stage0-sysroot/lib/rustlib/src/rust --exclude build/x86_64-pc-windows-msvc/stage1/lib/rustlib/rustc-src/rust --exclude build/x86_64-pc-windows-msvc/stage1/lib/rustlib/src/rust --exclude build/x86_64-pc-windows-msvc/stage2/lib/rustlib/rustc-src/rust --exclude build/x86_64-pc-windows-msvc/stage2/lib/rustlib/src/rust --exclude build/host \
+                --preserve-permissions --format=posix $@ \
+                | zstd -1 -T0 \
+                | aws s3 cp - "$(s3_url "${CIRCLE_JOB}")"
+        else
+            # Call `zstd` separately to be able to use all cores available (`-T0`)
+            # and the lowest compression level possible, to speed the compression
+            # as much as possible.
+            "${TAR}" c --exclude build/metrics.json $@ | zstd -1 -T0 | aws s3 cp - "$(s3_url "${CIRCLE_JOB}")"
+        fi
```
pietroalbini, line 69: Can we keep the underlying command the same? `-cf-` vs `c` should be equivalent, and the `--exclude` could be added as a `tar_flags` variable.

ferrocene/tools/generate-tarball/src/signatures.rs
```diff
@@ -78,7 +79,10 @@ fn collect_files(
             package.files.push(PackageFile {
                 path: relative_path.into(),
                 sha256: hash_file(&entry)?,
+                #[cfg(not(windows))]
                 posix_mode: entry.metadata()?.mode(),
+                #[cfg(windows)]
+                posix_mode: 0, // TODO: Not used! Should we cfg it in criticaltrust?
```
pietroalbini, line 85: Let's not cfg in criticaltrust: a Linux criticaltrust should be able to understand Windows criticaltrust files (otherwise the release process can't parse them).

ferrocene/tools/generate-tarball/src/signatures.rs
```diff
@@ -97,14 +101,25 @@ fn hash_file(path: &Path) -> Result<Vec<u8>, Error> {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(not(windows))]
```
pietroalbini, line 104: Why don't we just cfg the whole tests module?

src/bootstrap/src/ferrocene/doc.rs
```diff
@@ -812,7 +812,6 @@ impl Step for EmptyStep {
 // Note: this function is correct for the use made in this module, but it will not work correctly
 // if paths don't exists or the paths do not have any segments in common.
 fn relative_path(base: &Path, path: &Path) -> PathBuf {
-    let base = std::fs::canonicalize(base).unwrap_or_else(|_| base.to_path_buf());
```
pietroalbini, line 815: We need to be super careful with this, have you checked it doesn't break anything?

`relative_path` has been added because if any full path gets into Sphinx signatures break.

Hoverbear:
bors try

src/bootstrap/src/ferrocene/doc.rs
Hoverbear: Will this step get run at all on Windows? I bumped into this while running `x.py doc` thinking I was on Linux if I recall. It may be I can just yank this.

It was causing an issue on Windows due to `.canonicalize()` producing UNC paths.

ferrocene/tools/generate-tarball/src/signatures.rs
Hoverbear: I was scared we'd add a test and not realize it's not running on Windows. 😅  I'll block cfg it.

ferrocene/ci/scripts/persist-between-jobs.sh
Hoverbear: I initially had this but had some issues with shell expansion. I'll revisit it.

Hoverbear: Wait, we're not even calling this. I'm just gonna revert.

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

ferrocene/ci/configure.sh
```diff
@@ -138,6 +143,39 @@ add --release-description="Ferrocene by Ferrous Systems"
...
+    # MSVC one is correct.
+    add --build="x86_64-pc-windows-msvc"
+elif [[ "${FERROCENE_HOST:-}" = "x86_64-unknown-linux-gnu" ]]; then
+    add --build="x86_64-unknown-linux-gnu"
+elif [[ "${FERROCENE_HOST:-}" = "aarch64-apple-darwin" ]]; then
+    add --build="aarch64-apple-darwin" # We only build on ARM Macs right now!
+fi
+
+# The Rust build system defaults to calling `cc` on Windows, which does not exist
+if [[ "${FERROCENE_BUILD_HOST:-}" = "x86_64-pc-windows-msvc" ]]; then
```
pietroalbini, line 164: We need to gate this on `is_internal`.

ferrocene/ci/configure.sh
```diff
@@ -138,6 +143,39 @@ add --release-description="Ferrocene by Ferrous Systems"
...
+    # In theory, when run under MSYS the default ABI would be GNU, not MSVC.
+    # We're running under MSYS just for compatibility with our CI scripts, not
+    # because we target the GNU ABI, so switching the build platform to the
+    # MSVC one is correct.
+    add --build="x86_64-pc-windows-msvc"
+elif [[ "${FERROCENE_HOST:-}" = "x86_64-unknown-linux-gnu" ]]; then
+    add --build="x86_64-unknown-linux-gnu"
+elif [[ "${FERROCENE_HOST:-}" = "aarch64-apple-darwin" ]]; then
+    add --build="aarch64-apple-darwin" # We only build on ARM Macs right now!
+fi
```
pietroalbini, line 161: This seems redundant with the `FERROCENE_BUILD_HOST` environment variable. Also, if the content of the `FERROCENE_BUILD_HOST` environment variable would be the same as the one bootstrap auto-detects, we should be able to skip setting it.

I'd do something like this here:

```bash
if [[ -x "${FERROCENE_BUILD_HOST+x}" ]]; then
    add "--build=${FERROCENE_BUILD_HOST}"
fi
```

.circleci/workflows.yml
```diff
@@ -404,10 +435,78 @@ jobs:
...
+      size: large # 8-core
+      shell: bash.exe -euo pipefail
+    environment:
+      FERROCENE_BUILD_HOST: x86_64-pc-windows-msvc
+      FERROCENE_HOST: x86_64-pc-windows-msvc
+      FERROCENE_TARGETS: << pipeline.parameters.targets--x86_64-pc-windows-msvc--build >>
+      SCRIPT: |
+        ferrocene/ci/scripts/llvm-cache.sh download
+        ./x.py --stage 2 build library src/tools/rustdoc
+        ./x.py --stage 1 build src/tools/rustdoc
```
pietroalbini, line 479: I think we can comment these lines out until we split this into multiple jobs. The only purpose of them was to cache stuff in the `-build` job, but since we're running everything in a single job we don't have to worry about manually pre-building dependencies. Bootstrap will do that for us.

.circleci/workflows.yml
```diff
@@ -71,6 +81,7 @@ jobs:
     resource_class: large # 4-core
     environment:
       FERROCENE_CUSTOM_LLVM: /usr/lib/llvm-17
+      FERROCENE_BUILD_HOST: x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
pietroalbini: *Linux and macOS

pietroalbini (approved):
bors merge

Hoverbear:
bors try

pietroalbini (approved):
bors merge

Hoverbear:
bors merge

Please...?

amanjeev (approved):
bors merge

by the power vested in me as a rubber stamp.

PR #583, opened 2024-04-29
[24.05.0] Skip checking the cargo hash in the self-testing tool if missing

pietroalbini:
The self-testing tool checks whether all binaries match the expected commit hash, but unfortunately that is broken when building from source. That's because Cargo doesn't pick up any commit hash when building from the source tarball (this is an upstream problem).

While we fix the problem upstream, we need to make the self-testing tool useful in builds from source of 24.05.0. This hacky patch does that by allowing *just Cargo* not to have any hash in its version output. It won't affect binary distribution as there the hash is present.


Hoverbear (approved):

pietroalbini:
bors merge

PR #584, opened 2024-04-29
Gracefully handle cargo errors when updating the lockfile during pulls

pietroalbini:
This should fix the automation failure.


tshepang (approved):
bors merge

PR #589, opened 2024-04-30
Sign the internal procedures for 1.76

pietroalbini:
Signature for the second document bundle to send to TÜV.


tshepang (approved):
bors merge

PR #593, opened 2024-04-30
Fix documentation index page

pietroalbini:
The documentation index page right now is showing upstream's index page, which is not what we want. To work around that, this PR changes that step to panic when trying to execute it, and disables another step that calls it.


Hoverbear (commented):
I assume this was resulting from the recent `split-tasks.py` change? #549

Hoverbear (approved):

Veykril (approved):

Veykril:
bors merge

pietroalbini:
bors retry

PR #594, opened 2024-04-30
`rustfmt` qualification docs

Urhengulas:


ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -0,0 +1,30 @@
...
+
+To use ``rustfmt`` in a safety critical setup you need to adhere to following workflow:
+
+1. Make a change to the source code of your crate.
+2. Commit or stage the change. The working tree must be clean.
+3. Run ``rustfmt`` on the Rust files of your crate.
+4. Perform following checks to ensure correct behaviour of ``rustfmt``
+    1. Compile your crate, to ensure that ``rustfmt`` did not introduce any syntax errors.
+    2. Review the changes ``rustfmt`` made, to ensure there are only syntactic, but no semantic changes.
+    3. Run your test suite, to ensure no semantic changes occured.
```
pietroalbini, line 30: This should not be part of the user manual, it should be in the "constraints" page of the safety manual for rustfmt.

ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -0,0 +1,30 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+Using ``rustfmt``
+=================
```
pietroalbini, line 5: This is missing `rustfmt --check`, and using it through `cargo fmt`.

ferrocene/doc/user-manual/src/rustfmt/install.rst
```diff
@@ -0,0 +1,37 @@
...
+
+   .. code-block::
+
+      rustfmt-x86_64-unknown-linux-gnu-af9a431b1.tar.xz
+
+2. Unpack the archive
+
+   .. note::
+
+      It is recommended to unpack it to the same directory as the rest of the Ferrocene toolchain, but it is not strictly neccessary.
```
pietroalbini, line 25: This is wrong, `rustfmt` requires rustc to be installed in the same directory, as it dynamically links to it.

ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -0,0 +1,30 @@
...
+
+To format a single file, invoke ``rustfmt`` like this:
```
pietroalbini, line 7: We should mention that this also formats all the child modules, so doing `rustfmt src/main.rs` / `rustfmt src/lib.rs` formats the whole project.

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
```diff
@@ -0,0 +1,86 @@
...
+Command-Line Interface
+======================
+
+.. cli:program:: rustfmt
+
+   .. cli:option:: --check
+
+      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
+
+   .. cli:option:: --emit [files|stdout|coverage|checkstyle|json]
```
pietroalbini, line 13: This needs to document what each of the possible value does, and when you might want to use each possible value.

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
```diff
@@ -0,0 +1,86 @@
...
+
+      What data to emit and how
+
+   .. cli:option:: --backup
+
+      Backup any modified files.
```
pietroalbini, line 19: I'm pretty sure this flag is not covered by the hazop (or at least it wasn't last time I checked), so we should probably mention the flag isn't allowed in safety-critical contexts.

Also, this is not documenting *where* modified files are backed up.

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
```diff
@@ -0,0 +1,86 @@
...
+
+   .. cli:option:: --config-path
+
+      Recursively searches the given path for the rustfmt.toml config file. If not found reverts to the input file path
+
+   .. cli:option:: --edition [2015|2018|2021]
+
+      Rust edition to use
```
pietroalbini, line 27: What changes when changing the edition? We don't need to list all changes, but at least mention the general shape of what someone could expect.

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
```diff
@@ -0,0 +1,86 @@
...
+
+      Format specified line ranges. Run with ``--help=file-lines`` for more detail (unstable).
+
+   .. cli:option:: --error-on-unformatted
+
+      Error if unable to get comments or string literals within max_width, or they are left with trailing whitespaces (unstable).
+
+   .. cli:option:: --skip-children
+
+      Don't reformat child modules (unstable).
```
pietroalbini, line 61: We shouldn't document unstable features.

ferrocene/doc/user-manual/src/rustfmt/install.rst
```diff
@@ -0,0 +1,37 @@
...
+Installing ``rustfmt``
+======================
+
+1. Download the ``rustfmt`` archive for your host platform
+
+   The pattern of the name is:
+
+   .. code-block::
+
+      rustfmt-<target>-<hash>.tar.xz
```
pietroalbini, line 13: This is not always true. For stable releases we use the version number rather than the hash.

ferrocene/doc/user-manual/src/rustfmt/install.rst
```diff
@@ -0,0 +1,37 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
```
pietroalbini, line 1: Thinking about it, I'd prefer if we just link to the main installation page (https://public-docs.ferrocene.dev/main/user-manual/install.html). That page is more accurate, and will be expanded as we add more installation methods (like criticalup).

That page says:

> For each of the targets you identified above, download all the archives listed in the “installation archives” section of the target documentation.

So we should add a "The following archives are needed when installing the optional rustfmt tool:" in https://public-docs.ferrocene.dev/main/user-manual/targets/x86_64-unknown-linux-gnu.html#archives-to-install

pietroalbini (changes requested):

ferrocene/doc/user-manual/src/rustfmt/usage.rst
Urhengulas: 👍🏾

ferrocene/doc/user-manual/src/rustfmt/usage.rst
Urhengulas: Should we mention `cargo fmt` even if it is not going to be qualified (yet)?

ferrocene/doc/user-manual/src/rustfmt/install.rst
Urhengulas: Oh really? I did not know.

Is this Ferrocene specific or does rustfmt always require to be placed in the same dir as rustc?

ferrocene/doc/user-manual/src/rustfmt/usage.rst
Urhengulas: Didn't know that either. I thought it really just formats the files you state explicitly and that the "formatting the whole project" was a feature of `cargo fmt`.

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
Urhengulas: Okay. Right now I just copied the contents from `rustfmt --help`.

ferrocene/doc/user-manual/src/rustfmt/usage.rst
pietroalbini: Yes, we can add a note that `cargo fmt` is not qualified yet (as we do in the CLI page for rustc).

ferrocene/doc/user-manual/src/rustfmt/rustfmt-cli.rst
Urhengulas: Kk

Urhengulas: (The existence of this option should have been a hint that it does reformat child modules 😅)

ferrocene/doc/user-manual/src/rustfmt/install.rst
pietroalbini: You always need to install rustfmt in the same directory as rustc. If you run `ldd` on a rustfmt binary you see it dynamically links to `librustc_driver.so`, which is provided by rustc.

ferrocene/doc/user-manual/src/rustfmt/usage.rst
Urhengulas: `cargo fmt` has a few additional cli options (package, manifest-path etc.). Should we also document them?

pietroalbini:
Can we please split this PR into a PR per document? i.e. a PR for the safety manual, a PR for the user manual, etc.

Urhengulas:
> Can we please split this PR into a PR per document? i.e. a PR for the safety manual, a PR for the user manual, etc.

I don't think I can finish one without working on the other. But I can split them into two PRs when they are done-ish.

ferrocene/doc/user-manual/src/rustfmt/usage.rst
tshepang: same

PR #595, opened 2024-04-30
Prototype out aarch64 Linux with host tools

Hoverbear:
Soon we will be introducing `riscv64gc-unknown-linux-gnu` which will be a platform which includes host tools.

While exploring implementation it became rather clear that not only could we use the same strategy for `aarch64-unknown-linux-gnu`, and in fact, without adding support the related code would gather some notable cruft.

So, this PR goes ahead and enables aarch64 Linux with Host tools.

It's going to fail until I fix the self-test step.


Hoverbear:
bors try

I'm expecting this to fail still.

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

.circleci/calculate-parameters.py
```diff
@@ -116,7 +117,7 @@ def calculate_llvm_rebuild(target):
     """
     url = urllib.parse.urlparse(subprocess.run(
         ["ferrocene/ci/scripts/llvm-cache.sh", "s3-url"],
-        env={"FERROCENE_HOST": target},
+        env={"FERROCENE_HOST": target, "PATH": os.environ.get("PATH")},
```
Hoverbear, line 120: I'm *really* surprised I needed to include this, but I did.

Hoverbear:
bors try

Hoverbear:
bors try

Now the self-testers should have everything they need...

Hoverbear:
Oh!

bors try-

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
For increased motivation https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/

.circleci/calculate-parameters.py
```diff
@@ -34,11 +34,11 @@
 # How long should it take before an image is rebuilt.
 REBUILD_IMAGES_OLDER_THAN_DAYS = 7
 
-# Targets only built (and self-tested!) on Linux.
-LINUX_ONLY_TARGETS = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu"]
+AARCH64_LINUX_BUILD_HOST = "aarch64-unknown-linux-gnu"
+X86_64_LINUX_BUILD_HOST = "x86_64-unknown-linux-gnu"
```
Urhengulas, line 38: Maybe make the `*_HOST`s lists as well? Then you don't need to wrap them in square brackets whenever you add them to another list.

PS. I know it is odd because it will only ever be a single element in the list, but it also avoids the bug (which you hit iirc) of accidentally adding two strings together.

```suggestion
AARCH64_LINUX_BUILD_HOST = ["aarch64-unknown-linux-gnu"]
X86_64_LINUX_BUILD_HOST = ["x86_64-unknown-linux-gnu"]
```

.circleci/workflows.yml
```diff
@@ -47,7 +56,13 @@ parameters:
   targets--x86_64-unknown-linux-gnu--self-test:
     type: string
     default: ""
-  targets--x86_64-unknown-linux-gnu--std-only:
+  targets--x86_64-unknown-linux-gnu--std:
```
Urhengulas, line 59: Why don't we need `targets--aarch64-unknown-linux-gnu--std` as well?

.circleci/workflows.yml
```diff
@@ -180,19 +195,19 @@ jobs:
           restore-from-job: x86_64-linux-build
 
   x86_64-linux-dist-targets:
-    executor: docker-ubuntu-20
+    executor: docker-x86_64-ubuntu-20
     resource_class: large # 4-core
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
-      FERROCENE_TARGETS: << pipeline.parameters.targets--x86_64-unknown-linux-gnu--std-only >>
+      FERROCENE_TARGETS: << pipeline.parameters.targets--x86_64-unknown-linux-gnu--std >>
```
Urhengulas, line 202: Are `*--build` the cross compilation targets and `*--std` the host targets?

.circleci/workflows.yml
```diff
@@ -328,7 +343,80 @@ jobs:
...
+  aarch64-linux-build:
+    executor: docker-aarch64-ubuntu-20
+    resource_class: arm.xlarge # 8-core
+    environment:
+      FERROCENE_HOST: aarch64-unknown-linux-gnu
+      FERROCENE_TARGETS: << pipeline.parameters.targets--aarch64-unknown-linux-gnu--build >>
+      SCRIPT: |
+        ferrocene/ci/scripts/llvm-cache.sh download
+        ./x.py --stage 2 build library src/tools/rustdoc
+        ./x.py --stage 1 build src/tools/rustdoc
```
Urhengulas, line 378: Not PR specific but: Why are we first building stage 2 and then stage 1?

.circleci/workflows.yml
```diff
@@ -681,6 +812,26 @@ workflows:
...
+          requires:
+            - aarch64-linux-build
+      - aarch64-linux-dist-tools:
+          requires:
+            - aarch64-linux-build
+      - aarch64-linux-self-test:
+          requires:
+            - aarch64-linux-dist
+            - aarch64-linux-dist-tools
+            - x86_64-linux-dist-targets
```
Urhengulas, line 833: Not `x86_64-linux-dist-tools` as well?

Urhengulas (dismissed):
It seems good to me, just a few nits. But I have to say that I didn't work so much on our CI so far, so maybe there are some things I oversee.

.circleci/calculate-parameters.py
```diff
@@ -124,7 +125,7 @@ def calculate_llvm_rebuild(target: str):
     url: urllib.parse.ParseResult = urllib.parse.urlparse(
         subprocess.run(
             ["ferrocene/ci/scripts/llvm-cache.sh", "s3-url"],
-            env={"FERROCENE_HOST": target},
+            env={"FERROCENE_HOST": target, "PATH": os.environ.get("PATH")},
```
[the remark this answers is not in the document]
pietroalbini: On one hand, as long as we need just a few environment variables, just passing the variables needed by the script is cleaner, as it avoids external environment variables from accidentally messing things up.

On the other hand, to inherit the parent environment this syntax is slightly nicer:

```python
env={**os.environ, "FERROCENE_HOST": target}
```

.circleci/workflows.yml
pietroalbini: Because cross-compilation targets can be built on just a single builder, so we don't need to also build them on aarch64 Linux.

.circleci/workflows.yml
pietroalbini: Because some tests need stage 1 rustdoc and some other tests need stage 2 rustdoc. We thus build both of them. The order doesn't really matter.

.circleci/workflows.yml
```diff
@@ -556,6 +644,41 @@ jobs:
           name: Empty step to make sure the job always has steps
           command: echo
 
+  build-docker-aarch64:
+    description: Job that builds and uploads a Docker image, if needed.
+    docker:
+      - image: cimg/base:2023.06 # Only specific images are supported by ARM https://discuss.circleci.com/t/product-launch-arm-docker-generally-available/48601
```
pietroalbini, line 650: Can we see whether `cimg/base:current` works nowadays? That message in their discussion forum might just be because at the time of the release older images didn't support Arm.

ferrocene/ci/docker-images/aarch64-ubuntu-20/Dockerfile
```diff
@@ -0,0 +1,128 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
```
pietroalbini, line 1: I wonder how much we can reduce duplication here with the x86_64 Dockerfile? Can we get to a point where we have a single Dockerfile for the two images?

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -10,12 +10,22 @@ use crate::linkers::Linker;
 use crate::report::Reporter;
 
 static SUPPORTED_TARGETS: &[TargetSpec] = &[
+    #[cfg(target_arch = "x86_64")]
```
pietroalbini, line 13: Why are these cfg'd?

ferrocene/ci/docker-images/aarch64-ubuntu-20/Dockerfile
Hoverbear: I was looking a bit at that last day and I think the main issue is checksums, which is a surmountable problem.

ferrocene/tools/self-test/src/targets.rs
Hoverbear: Otherwise the `aarch64` host tools will try to run the `x86_64` tests with the host CC, which doesn't work.

.circleci/workflows.yml
Hoverbear: This would be a good rule of thumb, yes!

.circleci/calculate-parameters.py
Hoverbear: I waffled on this for quite some time while writing it but I think you're right, and past-me made the wrong choice.

Notably, the `config.toml` treats `host` as an array, so we should too.

ferrocene/ci/docker-images/aarch64-ubuntu-20/Dockerfile
Hoverbear: I'm kind of focusing on this path in https://github.com/ferrocene/ferrocene/pull/690, unless you think I should prioritize getting it into this PR.

Hoverbear:
bors try

(TO test the current cimg)

Hoverbear:
Yay!

.circleci/workflows.yml
Hoverbear: Yes!

tshepang (approved):
bors merge

tshepang:
bors retry

ferrocene/ci/docker-images/aarch64-ubuntu-20/Dockerfile
pietroalbini: Let's do it in a followup PR.

pietroalbini (approved):

Hoverbear:
Landing this after https://github.com/ferrocene/ferrocene/pull/690

Hoverbear:
Hmmm we should take a stab at https://docs.docker.com/build/building/multi-platform/

Hoverbear:
Resolving the merge was kinda annoying and with #721 landed there were some real material changes. I'm making a new PR.

PR #599, opened 2024-05-02
Require Linux targets for the macOS self-test job

pietroalbini:
The macOS self test job might be executed slightly before the Linux cross-compilation targets get built. This causes spurious failures like we saw in https://github.com/ferrocene/ferrocene/pull/593.

cc @Hoverbear we want to do the same for the Windows PR


Hoverbear (approved):
bors merge

PR #604, opened 2024-05-03
Make self-test easier to debug

Hoverbear:
While trying to debug a self-test failure in #595 I noticed two complicating issues:

* A cleanup of downloaded artifacts was always performed, meaning it was not possible to inspect a failed state.
* Command failures did not include `stderr`, making it very difficult to discern the source of the problem without reproducing the state manually (cleanup also happens here)

This PR:

* Adds `COMMIT="..."` so we can tell the script which commit to test against if we want.
* Adds `SKIP_CLEANUP=true` so we tell the script not to clean up downloaded artifacts.

### DIFF
--- a/ferrocene/ci/scripts/run-self-test.sh
+++ b/ferrocene/ci/scripts/run-self-test.sh
@@ -5,16 +5,20 @@
 set -euo pipefail
 IFS=$'\n\t'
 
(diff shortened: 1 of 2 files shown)


tshepang (approved):
bors merge

Hoverbear:
bors merge

PR #610, opened 2024-05-06
Prepare rustfmt docs

Urhengulas:
This PR moves all ferrocene specific documents into a subfolder for both the user and safety manual.


Urhengulas:
> * SM chapter 1.1: title says _Using Ferrocene_, and I think that is not meant to be specific to Ferrocene. There is also another _Ferrocene_ mention in that chapter.

"meant" or "not meant"?

>     * SM chapter 2: should probably move to _Ferrocene_ section, because there is a lot of mention of Ferrocene, and will not apply to rustfmt.

I think it does. The problem is that we use "Ferrocene" for the organization (us) as well as the toolchain. For example it says
> Interactions between customers and Ferrocene are primarily done through the [customer portal](https://customers.ferrocene.dev/).

This clearly refers to the organization, not the toolchain. The same is true for most (all?) other mentions of Ferrocene on that page. If you will, rustfmt could be part of the Ferrocene toolchain. Maybe we can actually rename the "Ferrocene" section to "rustc"?

tshepang (approved):
bors merge

Urhengulas:
bors retry

Urhengulas:
bors try

japaric (approved):

PR #612, opened 2024-05-06
update lockfiles during upstream-pull

tshepang:
This has some advantages
- avoids the manual work of approving mechanical PRs
- avoids using CI time for changes that can only rarely break a CI build
- avoids lockfiles and upstream pulls getting out of sync, which can result in confusion


Dajamante (approved):
bors merge

pietroalbini:
This is also updating the lockfiles on beta and stable, while it should only do it on main.

tshepang:
working on a PR

tshepang:
> This is also updating the lockfiles on beta and stable, while it should only do it on main.

https://github.com/ferrocene/ferrocene/pull/613

PR #565, opened 2024-04-25
Use a compatible shasum for Mac

Hoverbear:
`sha256sum` is not a valid binary on Mac, and appears to have been silently failing.

This is causing the llvm cache to not be used and adding ~20 mins to most Mac workflows.

```
$ sha256sum
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #1:1:1]
 1 │ sha256sum
   · ────┬────
   ·     ╰── executable was not found
   ╰────
  help: No such file or directory (os error 2)

Exit status 1
$ shasum -a 256
asdasd^D
61b4c705859f4158d38090c1e38e8fdc4f3d29db007f012766276aa498835cf6  -
```

image
https://app.circleci.com/pipelines/github/ferrocene/ferrocene/3574/workflows/9f9f9a89-56a2-4074-bb18-ef5b7a29fa9f/jobs/17241

PR #614, opened 2024-05-07
Make llvm-cache.sh log more useful and verbose information

Hoverbear:
We store our LLVM caches in a specific S3 bucket location. This PR ensures we log the location, it also ensures we log both when we start, and when we complete, both actions. (Previously we only logged starting storing the cache, and after restoring)

This fell mostly out of #565.


tshepang (approved):
bors merge

PR #623, opened 2024-05-13
Call section `rustc` not `Ferrocene`

Urhengulas:
Pietro said to call the section `rustc` and move the `general` pages to the root and not a subsection. This PR implements this.


pietroalbini (approved):
bors merge

PR #629, opened 2024-05-16
Write `rustfmt` user manual

Urhengulas:
This PR adds `rustfmt` to the user manual. Added sections are:
- installation
- usage
- command-line interface _(this section is bare bones atm and will be extended in a follow-up PR)_.

Are any sections missing?

### DIFF
--- a/ferrocene/doc/user-manual/src/rustfmt/usage.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/usage.rst
@@ -9,7 +9,7 @@ Using ``rustfmt``
 ``rustfmt <file>...``
 ---------------------
 
-To format one or more files, just pass the paths to ``rusfmt``. This will format the passed files and all of their child modules.
+To format one or more files, just pass the paths to ``rusfmt``. This will format the files and all of their child modules.
 
 Assuming following project structure, ``rustfmt main.rs`` will format the whole project, while ``rustfmt mod1.rs`` will format ``mod1.rs``, ``mod1/mod10.rs`` and ``mod1/mod11.rs``, but neither ``mod2.rs`` nor ``mod2/mod20.rs``.
 
@@ -30,7 +30,7 @@ Assuming following project structure, ``rustfmt main.rs`` will format the whole 
 ``rustfmt --check <file>...``
 -----------------------------
 
-If the ``--check`` option is passed ``rusfmt`` checks if the files are well formatted. It reports what it would format, but will not modify any files. For example:
+If the ``--check`` option is used, ``rusfmt`` checks if the files are well formatted. It reports what it would format, but will not modify any files. For example:
 
 .. code-block::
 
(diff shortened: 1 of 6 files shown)


ferrocene/doc/user-manual/src/targets/x86_64-unknown-linux-gnu.rst
```diff
@@ -49,6 +49,10 @@ target as a cross-compilation target:
 
 * ``rust-std-x86_64-unknown-linux-gnu``
 
+To install the optional ``rustfmt`` tool following archive is needed:
```
tshepang, line 52:
```suggestion
To install the optional ``rustfmt`` tool, the following archive is needed:
```

ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -0,0 +1,58 @@
...
+   ├── mod1.rs
+   ├── mod2
+   │   └── mod20.rs
+   └── mod2.rs
+
+
+``rustfmt --check <file>...``
+-----------------------------
+
+If the ``--check`` option is passed ``rusfmt`` checks if the files are well formatted. It reports what it would format, but will not modify any files. For example:
```
tshepang, line 33:
```suggestion
If the ``--check`` option is used, ``rusfmt`` checks if the files are well formatted. It reports what it would format, but will not modify any files. For example:
```

ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Using ``rustfmt``
+=================
+
+``rustfmt`` formats Rust code or checks if Rust code is well formatted.
+
+``rustfmt <file>...``
+---------------------
+
+To format one or more files, just pass the paths to ``rusfmt``. This will format the passed files and all of their child modules.
```
tshepang, line 12:
```suggestion
To format one or more files, just pass the paths to ``rusfmt``. This will format the files and all of their child modules.
```

tshepang:
Since this can go stale (and it already did, since my current `rustfmt` didn't have some `--emit` values), I wonder if we can automatically keep it fresh by using the text from `--help`... I imagine there is a Sphinx thing for this. This would need to be accompanied by improved `--help` output upstream, since [what we have is indeed barebones](https://github.com/ferrocene/ferrocene/pull/594#discussion_r1587189148).

Urhengulas:
> Since this can go stale (and it already did, since my current `rustfmt` didn't have some `--emit` values), I wonder if we can automatically keep it fresh by using the text from `--help`... I imagine there is a Sphinx thing for this. This would need to be accompanied by improved `--help` output upstream, since [what we have is indeed barebones](https://github.com/ferrocene/ferrocene/pull/594#discussion_r1587189148).

I like the idea of automating it. At least checking that the options and values are correct. For the text I am not sure if upstream would like the long descriptions with examples. At least in the terminal I assume it makes more sense if it is more concise.

tshepang (approved):
bors merge

PR #634, opened 2024-05-20
sync x86_64-unknown-linux-gnu settings file with upstream

tshepang:
No need to carry the diff since upstream also moved the target to use lld by default.


tshepang:
bors try

pietroalbini:
Have we checked `CFG_USE_SELF_CONTAINED_LINKER` is actually set by our CI?

tshepang:
> `CFG_USE_SELF_CONTAINED_LINKER`

[it's set automatically when `--enable-lld` is used](https://github.com/ferrocene/ferrocene/blob/cd6a73770d8aa28a5ca5fe706673ffdd2996b214/src/bootstrap/src/core/build_steps/compile.rs#L1148-L1150), which we do in `configure.sh` (unless am not reading it well)

tshepang:
> Also, there are more changes to revert from #100.

the upstream change affects just x86_64-unknown-linux-gnu

pietroalbini:
We should still keep the coretest target consistent.

pietroalbini (approved):
bors merge

tshepang:
bors retry

PR #639, opened 2024-05-21
move submodule entry to a place where it will not conflict

tshepang:
An upstream change resulted in a conflict,
due to https://github.com/rust-lang/rust/pull/125166.


pietroalbini (approved):
bors merge

PR #643, opened 2024-05-22
replace external dep with functionality from std

tshepang:
- is unmaintained
- is 2 less crates to build


Urhengulas (approved):
bors merge

pietroalbini (approved):
bors merge

PR #645, opened 2024-05-22
Fix release scheduler to account for channel metadata v3

pietroalbini:
In https://github.com/ferrocene/publish-release/pull/51 we introduced channel metadata v3, to store the full list of releases rather than just the last release. This PR adds support for it in the release matrix calculation.

Also, it fixes two unrelated problems I discovered while developing this.


tshepang (approved):
bors merge

PR #647, opened 2024-05-22
Write `rustfmt` safety manual

Urhengulas:
Do the constraints and tool option make sense?

### DIFF
--- a/ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
+++ b/ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
@@ -1,0 +1,29 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+.. default-domain:: qualification
+
+Constraints
+===========
+
+Review diff
+-----------
+
+.. id:: RUSTFMT_CSTR_0010_CHECK_DIFF
+
+Associated requirement ID: :id:`RUSTFMT_AVD_CHECK_DIFF_001`
+
+After executing ``rustfmt``, the user shall review the modifications
+``rustfmt`` performed and ensure that there are no semantic changes.
+
+Test coverage
+-------------
+
+.. id:: RUSTFMT_CSTR_0020_TESTS
+
+Associated requirement ID: :id:`RUSTFMT_AVD_TESTS_002`
+
+Before executing ``rustfmt``, the user shall ensure that the code is covered by
+sufficient tests that ensure the correctness of the code.
+
+After executing ``rustfmt``, the user shall ensure that all tests pass.
--- a/ferrocene/doc/safety-manual/src/rustfmt/options.rst
+++ b/ferrocene/doc/safety-manual/src/rustfmt/options.rst
@@ -1,0 +1,10 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+Tool Options
+============
+
+``rustfmt`` can be used with all stable CLI options.
+
+For convenience, :doc:`user-manual:rustfmt/cli` provides further details on CLI
+options available to the end user.
--- a/ferrocene/doc/safety-manual/src/index.rst
+++ b/ferrocene/doc/safety-manual/src/index.rst
@@ -32,6 +32,13 @@ scope, in accordance to the standards above.
    rustc/degraded-environment
    rustc/constraints
 
+.. toctree::
+   :maxdepth: 2
+   :caption: rustfmt
+   :numbered:
+
+   rustfmt/usage
+
 .. appendices::
    :maxdepth: 2
    :caption: Appendices:
--- a/ferrocene/doc/safety-manual/src/index.rst
+++ b/ferrocene/doc/safety-manual/src/index.rst
@@ -39,6 +39,7 @@ scope, in accordance to the standards above.
 
    rustfmt/usage
    rustfmt/constraints
+   rustfmt/options
 
 .. appendices::
    :maxdepth: 2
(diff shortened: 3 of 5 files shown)


ferrocene/doc/safety-manual/src/rustfmt/options.rst
```diff
@@ -0,0 +1,15 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+Tool Options
+============
+
+``rustfmt`` is qualified exclusively for the following command line options:
+
+* Users shall pass command line option ``--emit files`` to each invocation of ``rustfmt``.
```
pietroalbini, line 9: Why are we forcing people to pass `--emit files`?

ferrocene/doc/safety-manual/src/rustfmt/options.rst
```diff
@@ -0,0 +1,15 @@
...
+* Users shall pass command line option ``--edition 2021`` to each invocation of ``rustfmt``.
```
pietroalbini, line 10: Why are we limiting to just `--edition 2021`? We do so for rustc as we only specified the 2021 edition in the FLS, but there is no such constraint for rustfmt.

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
```diff
@@ -0,0 +1,15 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+Constraints
```
pietroalbini, line 4: Constraints should be styled like rustc constraints: a section for each constraint, with a unique ID for each constraint and linking to the nodes of the HazOp they address.

ferrocene/doc/safety-manual/src/rustfmt/options.rst
```diff
@@ -0,0 +1,15 @@
...
+
+.. |Tool| replace:: ``rustfmt``
+.. |Link| replace:: :doc:`user-manual:rustfmt/cli`
+.. include:: /partials/options.rst
```
pietroalbini, line 14: I would not share the snippet across the various tools.

A lot of the text in the snippet for rustc doesn't apply: there are no flags that we restrict in the documentation, I don't think we need to mandate people pass any flags all the time, and these flags don't affect codegen. I think we can get away with a far smaller piece of text here.

Also, this approach for partials is pretty hacky, and I'd prefer a better templating solution if we were to need to share snippets in the future. Let's discuss about that in Zulip.

pietroalbini (changes requested):

Urhengulas:
Blocked on #666

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Install version control system
+------------------------------
+
+.. id:: RUSTFMT_CSTR_0010_HAVE_VCS
+
+Associated requirement ID: :id:`RUSTFMT_AVD_CHECK_DIFF_001`
+
+Before executing ``rustfmt``, the user shall ensure that the a version control
+system (VCS, e.g. ``git``) is installed and used.
```
pietroalbini, line 17: If we want to require a VCS (see next comment), having it installed shouldn't be a constraint, it should be a pre-requisite in the installation process.

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Review diff
+-----------
+
+.. id:: RUSTFMT_CSTR_0020_USE_VCS
+
+Associated requirement ID: :id:`RUSTFMT_AVD_CHECK_DIFF_001`
+
+After executing ``rustfmt``, the user shall review, with the help of the VCS, the
+modifications ``rustfmt`` performed and ensure that there are no semantic changes.
```
pietroalbini, line 27: Should we mandate the use of a VCS at all? I would hope that our customers use one, but if they want to use a different technique to review the changes they should be free to.

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
```diff
@@ -0,0 +1,46 @@
...
+sufficient tests that ensure the correctness of the code.
+
+Tests pass
+----------
+
+.. id:: RUSTFMT_CSTR_0030_USE_TESTS
+
+Associated requirement ID: :id:`RUSTFMT_AVD_TESTS_002`
+
+After executing ``rustfmt``, the user shall ensure that all tests pass.
```
pietroalbini, line 46: We should probably merge the two requirements into one.

pietroalbini (changes requested):

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
tshepang: I think we should not mandate VCS, especially since `rustfmt` works fine without it

pietroalbini: Like, we do want to mandate that they manually review the changes done by rustfmt did not cause any difference in the semantic meaning of the code, but we shouldn't say *how* they do it.

ferrocene/doc/safety-manual/src/rustfmt/constraints.rst
```diff
@@ -6,30 +6,22 @@
...
+.. id:: RUSTFMT_CSTR_0010_CHECK_DIFF
 
 Associated requirement ID: :id:`RUSTFMT_AVD_CHECK_DIFF_001`
 
-After executing ``rustfmt``, the user shall review, with the help of the VCS, the
-modifications ``rustfmt`` performed and ensure that there are no semantic changes.
+After executing ``rustfmt``, the user shall review the modifications
+``rustfmt`` performed and ensure that there are no semantic changes.
+
+Using a version control system for that makes sense, but is not mandated.
```
tshepang, line 19: would not mention version control at all... users can decide what *makes sense* for them

tshepang (approved):
bors merge

tshepang:
bors retry

PR #652, opened 2024-05-23
Refactor `.circleci/calculate-parameters.py`

Urhengulas:
This PR
- mostly adds type hints so that I get information about the variables and functions when hovering over them
- replaces deprecated `.utcnow()`
- in `calcualte_paramameters`, I switched from `target +=` to `target =`, because iiuc exactly one case will always match.
- sort imports alphabetically :smile:

---

PS. I also wanted to find a type for the global `s3` and `ecr` variable, but unfortunately I could not https://github.com/ferrocene/ferrocene/blob/a5a6acb2f0c1f84468c82c760c83d5a883a56a9e/.circleci/calculate-parameters.py#L58-L59

Python reports `botocore.client.S3` and `botocore.client.ECR`, but I am unable to import that:
```python
>>> import boto3
>>> type(boto3.client("s3", region_name="us-east-1"))
<class 'botocore.client.S3'>
>>> from botocore.client import S3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'S3' from 'botocore.client' (/usr/lib/python3.12/site-packages/botocore/client.py)
```

### DIFF
--- a/.circleci/calculate-parameters.py
+++ b/.circleci/calculate-parameters.py
@@ -16,6 +16,7 @@ import sys
 import datetime
 import urllib.parse
 import subprocess
+from typing import Callable
 
 
 # Path of the YAML file to extract the needed parameters from.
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):
bors merge

PR #663, opened 2024-05-29
Fix `x.py install`

Hoverbear:
@pietroalbini noticed that `x.py install` was broken due to some of the changes we'd made.

This is largely due to us packaging our tarballs slightly differently and not generating an `install.sh`.

This PR creates a new ferrocene specific install step for each package that directly renames the tarball outputs to their configured destinations.

It supports all the `config.toml` options as well as `DESTDIR`.

### Testing

* [ ] Test `x.py install` with a `install.prefix` set in `config.toml`
* [ ] Test `x.py install` with any number of `install.*dir` set in `config.toml`
* [ ] Test `x.py install` with a `install.prefix` and any number of `install.*dir` set in `config.toml`
* [ ] Test `x.py install` with the `DESTDIR` env set
* [ ] Test `x.py install` with the `DESTDIR` env set and a `install.prefix` set in `config.toml`
* [ ] Test `x.py install` with the `DESTDIR` env set and any number of `install.*dir` set in `config.toml`
* [ ] Test `x.py install` with the `DESTDIR` env set and a `install.prefix` set in `config.toml` and any number of `install.*dir` set in `config.toml`

### DIFF
--- a/src/bootstrap/src/ferrocene/install.rs
+++ b/src/bootstrap/src/ferrocene/install.rs
@@ -1,0 +1,332 @@
+// Based off ../core/build_steps/install.rs
+
+//! Implementation of the install aspects of the compiler.
+//!
+//! This module is responsible for installing the standard library,
+//! compiler, and documentation.
+
+use std::env;
+use std::fs;
+use std::path::{PathBuf, Path};
+// use std::process::Command;
+
+use crate::core::build_steps::dist;
+use crate::core::builder::{Builder, RunConfig, ShouldRun, Step};
+use crate::core::config::{Config, TargetSelection};
+use crate::utils::helpers::t;
+use crate::utils::tarball::GeneratedTarball;
+use crate::{Compiler, Kind};
+
+// #[cfg(target_os = "illumos")]
+// const SHELL: &str = "bash";
+// #[cfg(not(target_os = "illumos"))]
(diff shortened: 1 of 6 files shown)


src/bootstrap/src/ferrocene/install.rs
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            fn run($sel, $builder: &Builder<'_>) {
+                $run_item
+            }
+        })+
+    }
+}
+
+install!((self, builder, _config),
```
pietroalbini, line 141: If we want to provide support, we should replace this with the list of components we want people to actually be able to install.

pietroalbini (changes requested):

src/bootstrap/src/ferrocene/install.rs
Hoverbear: Oh, ya! This should have like `flip-link` etc. Are there any I should remove? 🤔

Hoverbear:
Looks like the `ferrocene-docs` job has some broken links...

src/bootstrap/src/ferrocene/install.rs
pietroalbini: We should probably remove the upstream docs.

pietroalbini (approved):
bors merge

PR #665, opened 2024-05-31
Installation docs for aarch64-apple-darwin

Hoverbear:
Similar to #664, but Apple flavored instead of glass-flavored.

### DIFF
--- a/ferrocene/doc/user-manual/src/targets/aarch64-apple-darwin.rst
+++ b/ferrocene/doc/user-manual/src/targets/aarch64-apple-darwin.rst
@@ -65,7 +65,3 @@ To use the target, the following additional flags must be provided to
 ``rustc``:
 
 - ``--target=x86_64-unknown-linux-gnu``
-
-- ``-C linker=<your-c-compiler>``
-
-  - e.g. ``-C linker=/usr/bin/clang``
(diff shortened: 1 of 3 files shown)


ferrocene/doc/user-manual/src/targets/aarch64-apple-darwin.rst
```diff
@@ -0,0 +1,71 @@
...
+
+.. warning::
+   
+   Experimental targets cannot be used in safety-critical contexts, and there is
+   no guarantee that the Ferrocene test suite is successfully executed on the
+   target. They are provided as a preview, with limited support available. They
+   should not be used in production.
+
+The ``aarch64-apple-darwin`` Ferrocene target provides support for macOS (Darwin) on
+aarch64.
```
pietroalbini, line 17: Nit: I'd use Apple Silicon here.

ferrocene/doc/user-manual/src/targets/aarch64-apple-darwin.rst
```diff
@@ -0,0 +1,71 @@
...
+-----------------------
+
+To use the target, the following additional flags must be provided to
+``rustc``:
+
+- ``--target=x86_64-unknown-linux-gnu``
+
+- ``-C linker=<your-c-compiler>``
+
+  - e.g. ``-C linker=/usr/bin/clang``
```
pietroalbini, line 71: Do we actually need users to specify this?

Hoverbear: Don't think so, but the Linux docs specify it so I did here.

pietroalbini: Response is the same as the Windows comment.

pietroalbini (approved):
bors merge

Hoverbear:
I'll fix this up.

PR #679, opened 2024-06-07
Enable 2 of 3 compiler-arguments target tests on all architectures

Veykril:
The output of these two tests does not contain any host specific strings, so there is no reason not to run these only on one target.


pietroalbini (approved):
bors merge

PR #681, opened 2024-06-07
Remove ignore-cross-compile directive from ui/macros/proc_macro.rs

Veykril:


Veykril:
bors try

tests/ui/macros/proc_macro.rs
```diff
@@ -1,6 +1,5 @@
 //@ run-pass
 //@ aux-build:proc_macro_def.rs
-//@ ignore-cross-compile
```
Veykril, line 3: This seems to be no longer required given by newer proc-macro tests not having this directive anymore. If that's the case (should try succeed) I'll upstream that change.

pietroalbini (approved):
bors merge

PR #682, opened 2024-06-07
Add `ferrocene.test-outcomes=local`

pietroalbini:
This adds the new `ferrocene.test-outcomes = "local"` option in `config.toml`, which uses the locally generated test outcomes to build the documentation. This was something already possible before, but you had to manually move `build/metrics.json` to a different directory and set that different directory as the custom outcomes path. The new option ensures this happens in the background.


tshepang:
would be good to mention this option in `config.example.toml`

ferrocene/doc/internal-procedures/src/docs/index.rst
```diff
@@ -109,19 +109,64 @@ system's "build metrics" feature. They are usually generated by CI, but it is
...
+If you are building Ferrocene from source on a single builder machine, or you
+need to use test outcomes from a local test suite invocation, you can add this
+to your ``config.toml``:
+
+.. code-block::
+
+   build.metrics = true
+   ferrocene.test-outcomes = "local"
+
+Note that there are some limitations to be aware when doing so:
```
tshepang, line 136:
```suggestion
Note that there are some limitations to be aware of when doing so:
```

tshepang (approved):
bors merge

PR #678, opened 2024-06-07
give user-manual access to the glossary

tshepang:

PR #683, opened 2024-06-08
expand user-manual for rustfmt

tshepang:
made on top of #678

### DIFF
--- a/ferrocene/doc/user-manual/src/conf.py
+++ b/ferrocene/doc/user-manual/src/conf.py
@@ -10,6 +10,8 @@ extensions = [
     "ferrocene_qualification",
     "ferrocene_toctrees",
     "ferrocene_domain_cli",
+    "ferrocene_autoglossary",
+
     "sphinx.ext.intersphinx",
     "sphinx.ext.autosectionlabel",
 ]
@@ -22,3 +24,8 @@ ferrocene_id = "UM"
 html_theme = "ferrocene"
 html_title = "Ferrocene User Manual"
 html_short_title = "User Manual"
+
+# Do not generate the index pages. We don't need them, and they cause
+# linkchecker to fail due to them including *all* glossary entries, including
+# entries that were removed by autoglossary.
+html_use_index = False
--- a/ferrocene/doc/user-manual/src/conf.py
+++ b/ferrocene/doc/user-manual/src/conf.py
@@ -1,9 +1,6 @@
 # SPDX-License-Identifier: MIT OR Apache-2.0
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
-import os
-import sys
-
 project = "Ferrocene User Manual"
 copyright = "The Ferrocene Developers"
 author = "The Ferrocene Developers"
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -53,7 +53,9 @@
 
       Short option: ``-l``.
 
-      Prints the names of mismatched files that were formatted. Prints the names of files that would be formatted when used with ``--check``.
+      Prints the names of mismatched files that were formatted.
+      When used with --check,
+      this option instead prints the names of files that are not well-formatted.
 
    .. cli:option:: --config [key1=val1,key2=val2...]
 
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -36,7 +36,10 @@
 
    .. cli:option:: --edition [2015|2018|2021]
 
-      Rust edition to use
+      Rust edition to use, which defaults to 2015 edition.
+
+      If using a different edition than the default, specify it here,
+      because the formatting may be different.
 
    .. cli:option:: --color [always|never|auto]
 
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -15,7 +15,10 @@
 
    .. cli:option:: --emit [files|stdout]
 
-      What data to emit and how
+      What kind of output to emit.
+
+      - ``files`` emits formatted output to files; this is the default
+      - ``stdout`` emits the formatted output to stdout
 
    .. cli:option:: --backup
 
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -30,7 +30,9 @@
 
    .. cli:option:: --config-path
 
-      Recursively searches the given path for the rustfmt.toml config file. If not found reverts to the input file path
+      Recursively searches the given path for the ``rustfmt.toml`` or ``.rustfmt.toml`` config file.
+      If not found, it reverts to the default behavior of searching the parent directories of the
+      file to be formatted.
 
    .. cli:option:: --edition [2015|2018|2021]
 
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -53,7 +53,7 @@
 
       Short option: ``-l``.
 
-      Prints the names of mismatched files that were formatted. Prints the names of files that would be formatted when used with ``--check`` mode.
+      Prints the names of mismatched files that were formatted. Prints the names of files that would be formatted when used with ``--check``.
 
    .. cli:option:: --config [key1=val1,key2=val2...]
 
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -10,7 +10,7 @@
 
       Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
 
-   .. cli:option:: --emit [files|stdout|coverage|checkstyle|json]
+   .. cli:option:: --emit [files|stdout]
 
       What data to emit and how
 
@@ -67,4 +67,3 @@
       Short option: ``-h``.
 
       Show this message or help about a specific topic: ``config`` or ``file-lines``
-   
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -24,6 +24,10 @@
 
       Backup any modified files.
 
+      Before a file is modified, rustfmt will create a copy of it in the same directory,
+      with the ``.bk`` extension.
+      Note that existing ``.bk`` files are over-written.
+
    .. cli:option:: --config-path
 
       Recursively searches the given path for the rustfmt.toml config file. If not found reverts to the input file path
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -82,3 +82,4 @@
       Short option: ``-h``.
 
       Show this message or help about a specific topic: ``config`` or ``file-lines``
+
--- a/ferrocene/doc/user-manual/src/rustfmt/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/cli.rst
@@ -8,7 +8,10 @@
 
    .. cli:option:: --check
 
-      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
+      Check if the input is formatted correctly.
+
+      Exits with status code 0 if input is formatted correctly.
+      Exits with status code 1 and prints a diff if formatting is not correct.
 
    .. cli:option:: --emit [files|stdout]
 
--- a/ferrocene/doc/user-manual/src/rustfmt/usage.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/usage.rst
@@ -11,22 +11,6 @@ Using ``rustfmt``
 
 To format one or more files, just pass the paths to ``rusfmt``. This will format the files and all of their child modules.
 
-Assuming following project structure, ``rustfmt main.rs`` will format the whole project, while ``rustfmt mod1.rs`` will format ``mod1.rs``, ``mod1/mod10.rs`` and ``mod1/mod11.rs``, but neither ``mod2.rs`` nor ``mod2/mod20.rs``.
-
-.. code-block::
-
-   $ tree src/
-   src/
-   ├── main.rs
-   ├── mod1
-   │   ├── mod10.rs
-   │   └── mod11.rs
-   ├── mod1.rs
-   ├── mod2
-   │   └── mod20.rs
-   └── mod2.rs
-
-
 ``rustfmt --check <file>...``
 -----------------------------
 
--- a/ferrocene/doc/user-manual/src/rustfmt/usage.rst
+++ b/ferrocene/doc/user-manual/src/rustfmt/usage.rst
@@ -46,7 +46,7 @@ If the ``--check`` option is used, ``rusfmt`` checks if the files are well forma
 Exit codes
 ----------
 
-When running with ``--check``, ``rustfmt`` will exit with ``0`` if ``rustfmt`` would not make any formatting changes to the input, and ``1`` if ``rustfmt`` would make changes. In other modes, ``rustfmt`` will exit with ``1`` if there was some error during formatting (for example a parsing or internal error) and ``0`` if formatting completed without error (whether or not changes were made).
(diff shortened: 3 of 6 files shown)


ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
 
    .. cli:option:: --check
 
-      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
+      Run in 'check' mode.
```
Dajamante, line 11: What's the reason of removing the explanation?

Dajamante (approved):
bors merge

pietroalbini:
bors cancel

ferrocene/doc/user-manual/src/rustfmt/cli.rst
pietroalbini, line 11: Rather than saying "Run in 'check' mode." I would explain here what it actually does.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
+      What kind of output to emit.
+
+      - ``files``, which emits formatted output to files, is the default
+      - ``stdout`` emits the formatted output to stdout
 
    .. cli:option:: --backup
 
       Backup any modified files.
 
+      This copies the files before they are unmodified in same directory as the original files,
```
pietroalbini, line 27: s/unmodified/modified. Also, this sentence is kinda hard to read, maybe something like this?

> Before a file is modified, rustfmt will create a copy of it in the same directory, with the ``.bk`` extension.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
+      If not found, it reverts to the default behavior of searching the parent directories of the
+      file to be formatted.
 
    .. cli:option:: --edition [2015|2018|2021]
 
-      Rust edition to use
+      Rust edition to use, which defaults to 2015 edition.
+
+      The only Edition handling that is currently supported is the module system,
+      which changed with the 2018 Edition.
```
pietroalbini, line 41: Do we have to specify this? I would only say that formatting might change depending on which edition you specify.

pietroalbini (changes requested):

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
 
    .. cli:option:: --check
 
-      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
+      Run in 'check' mode.
 
-   .. cli:option:: --emit [files|stdout|coverage|checkstyle|json]
+      Exits with status code 0 if input is formatted correctly.
+      Exits with status code 1 and prints a diff if formatting is not correct.
```
Urhengulas, line 15: Let's add examples to all options, like we do for `rustc`.

```suggestion
      .. code-block::

         $ # well-formatted file
         $ rustfmt --check src/main.rs
         $ echo $?
         0
         $ # not well-formatted file
         $ rustfmt --check src/main.rs 
         Diff in /home/urhengulas/Documents/krate/src/main.rs at line 1:
          fn main() {
         -    println!(
         -        "Hello, world!"
         -    );
         +    println!("Hello, world!");
          }
         $ echo $?
         1
```

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
 
-      What data to emit and how
+   .. cli:option:: --emit [files|stdout]
+
+      What kind of output to emit.
+
+      - ``files``, which emits formatted output to files, is the default
```
Urhengulas, line 20:
```suggestion
      - ``files`` emits formatted output to files, is the default
```

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: 172caf7887033820d290fae171d16a13d8a12ca0

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
+      - ``stdout`` emits the formatted output to stdout
 
    .. cli:option:: --backup
 
       Backup any modified files.
 
+      This copies the files before they are unmodified in same directory as the original files,
+      and changes the file extension to ``.bk``, replacing the ``.rs`` extension.
```
Urhengulas: Let's note that it overwrites existing `bk` files.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: ff01638c2c2765b00a19d8060bc22a07a7b5128b

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
 
-      Recursively searches the given path for the rustfmt.toml config file. If not found reverts to the input file path
+      Recursively searches the given path for the ``rustfmt.toml`` or ``.rustfmt.toml`` config file.
+      If not found, it reverts to the default behavior of searching the parent directories of the
+      file to be formatted.
 
    .. cli:option:: --edition [2015|2018|2021]
 
-      Rust edition to use
+      Rust edition to use, which defaults to 2015 edition.
```
Urhengulas, line 38: Does it? I assumed 2021 aka. the most recent one

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,37 @@
...
+
+      - ``files``, which emits formatted output to files, is the default
+      - ``stdout`` emits the formatted output to stdout
 
    .. cli:option:: --backup
 
       Backup any modified files.
 
+      Before a file is modified, rustfmt will create a copy of it in the same directory,
+      with the .bk extension.
```
Urhengulas, line 28: Let's note that it overwrites existing `bk` files

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -38,7 +52,7 @@
 
       Short option: ``-l``.
 
-      Prints the names of mismatched files that were formatted. Prints the names of files that would be formatted when used with ``--check`` mode.
+      Prints the names of mismatched files that were formatted. Prints the names of files that would be formatted when used with ``--check``.
```
Urhengulas, line 55: Would or should?
```suggestion
      Prints the names of mismatched files that were formatted. Prints the names of files that should be formatted when used with ``--check``.
```

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -38,7 +52,7 @@
...
 
    .. cli:option:: --config [key1=val1,key2=val2...]
```
Urhengulas, line 57: Should or must we specify which rustfmt config options there are?

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -67,4 +81,3 @@
       Short option: ``-h``.
 
       Show this message or help about a specific topic: ``config`` or ``file-lines``
```
Urhengulas, line 83:
```suggestion
      Show this message or help about a specific topic: ``config`` or ``file-lines``

```

ferrocene/doc/user-manual/src/rustfmt/usage.rst
```diff
@@ -11,22 +11,6 @@ Using ``rustfmt``
...
-   ├── main.rs
-   ├── mod1
-   │   ├── mod10.rs
-   │   └── mod11.rs
-   ├── mod1.rs
-   ├── mod2
-   │   └── mod20.rs
-   └── mod2.rs
-
-
```
Urhengulas, line 29: Why do you think we should delete this?

Urhengulas (changes requested):

ferrocene/doc/user-manual/src/conf.py
```diff
@@ -25,3 +24,8 @@
 html_theme = "ferrocene"
 html_title = "Ferrocene User Manual"
 html_short_title = "User Manual"
+
+# Do not generate the index pages. We don't need them, and they cause
+# linkchecker to fail due to them including *all* glossary entries, including
+# entries that were removed by autoglossary.
+html_use_index = False
```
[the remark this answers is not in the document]
tshepang: me did not check the faulure... only added this because it's in the other conf.py files

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: 926e0c9c8f0920f64b22f8a7840432f97657b71c

ferrocene/doc/user-manual/src/conf.py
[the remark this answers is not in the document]
pietroalbini: Because we just added the glossary to this (as the comment mentions).

ferrocene/doc/user-manual/src/rustfmt/cli.rst
pietroalbini: This example doesn't seem to provide much information. We already say above what the exit codes say. The only useful thing might be the diff, but even then it's described above.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
pietroalbini: No, all rustc-related tools assume the 2015 edition if no edition flag is specified. Otherwise every published edition would result in a breaking change of the `rustc foo.rs` behavior.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: 85819601c15c532dc66cc9edb2943754660603de

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: _would_ fits in that these files would be formatted if the `--check` flag was not used

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: hm, not sure... there are so many (78), so perhaps we should consult

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: 85819601c15c532dc66cc9edb2943754660603de

ferrocene/doc/user-manual/src/rustfmt/cli.rst
Urhengulas: So then we say "that would be formatted when not used with `--check`"?

ferrocene/doc/user-manual/src/rustfmt/cli.rst
Urhengulas: @pietroalbini what do you think? Maybe we can link to the rustfmt docs and/or include a copy of it? https://github.com/rust-lang/rustfmt/blob/master/Configurations.md

ferrocene/doc/user-manual/src/rustfmt/usage.rst
Urhengulas: I agree it is not so easy to read, but I think it makes sense to have an example in this place. Feel free to rewrite it.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
pietroalbini: I would include a copy of it in a "rustfmt configuration" page or something.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,54 @@
...
-      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
-
-   .. cli:option:: --emit [files|stdout|coverage|checkstyle|json]
-
-      What data to emit and how
+      Check if the input is formatted correctly.
+
+      Exits with status code 0 if input is formatted correctly.
+      Exits with status code 1 and prints a diff if formatting is not correct.
+      .. code-block::
```
pietroalbini, line 15: I still think this example is... too much? It doesn't add much information compared to what is explained in the text before, and it uses a lot of screen space.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: c0800f07861

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: hm, this is actually complex

how about 2e97dfee919

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: ed0f014feb1d2cd4bce738a1ecd14a78d9a97210

ferrocene/doc/user-manual/src/rustfmt/cli.rst
Urhengulas: I thought that we should include and example for each, since we do that for the `rustc` docs[^1]. And those examples also mostly show what was described in the text.

[^1]: https://public-docs.ferrocene.dev/main/user-manual/rustc/cli.html

ferrocene/doc/user-manual/src/rustfmt/cli.rst
Urhengulas: Mhh, I still don't think it is correct, but maybe it is also me not being a native speaker and therefore understanding it differently.

What about?
>  When used with --check, this option instead prints the names of files that are not well-formatted.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
Urhengulas: Our repo contains it at `src/tools/rustfmt/Configurations.md`. Can and should we tell sphinx to copy it from there?

ferrocene/doc/user-manual/src/rustfmt/cli.rst
```diff
@@ -8,23 +8,38 @@
 
    .. cli:option:: --check
 
-      Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 and prints a diff if formatting is required.
+      Check if the input is formatted correctly.
 
-   .. cli:option:: --emit [files|stdout|coverage|checkstyle|json]
+      Exits with status code 0 if input is formatted correctly.
+      Exits with status code 1 and prints a diff if formatting is not correct.
```
Urhengulas, line 15: Let's still include an example of how to call it? We do that for all rustc options, even if they are straight forward.

```suggestion
      .. code-block::

         $ rustfmt --check src/main.rs
```

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: that is totally better... 54ae328fc6b

ferrocene/doc/user-manual/src/rustfmt/cli.rst
tshepang: me think we should not do what was done before just because it was done before, unless of course such is required for qualification

Urhengulas: > we should not do what was done before just because it was done before

Good point.

ferrocene/doc/user-manual/src/rustfmt/cli.rst
pietroalbini: Yeah we should include the file directly from Sphinx.

There is a Sphinx extension called `myst_parser`. It is already installed in our virtual environment (the release notes ext requires it), and you [can enable it as shown in the docs](https://myst-parser.readthedocs.io/en/latest/intro.html#enable-myst-in-sphinx). At that point, including the file *should* just work.

I would do this in a separate PR.

Urhengulas: Agree to do that in a separate PR.

Urhengulas (approved):
I think we are good to go for this one.

Two follow-up PRs
- include rustfmt config docs (https://github.com/ferrocene/ferrocene/pull/683/files#r1634938840)
- add option to `cli:program` directive to not generate traceability matrix

ferrocene/doc/user-manual/src/rustfmt/usage.rst
pietroalbini: I think for now the glossary is enough.

pietroalbini (approved):
bors merge

PR #689, opened 2024-06-12
Change digital signatures configuration and allow loading them from docs tarballs

pietroalbini:
Before this PR, there were only two config knobs to manage the private files containing the digital signatures: either downloading them from S3 (the default, only available to Ferrous employees and contractors) or disabling digital signatures altogether (with `ferrocene.ignore-document-signatures=true`).

This meant that customers who paid for access to digital signatures did not have any way to build the same documentation from source. They do have a copy of the signature files though, in the `ferrocene-docs-signatures` tarball. We can thus let the build system extract the files from there, if the customer has access to the tarball.

With this PR, the configuration of digital signatures completely changes:

* To download the signatures from S3 you don't need to configure anything. This is equivalent to this explicit config:

  ```toml
  ferrocene.document-signatures = "s3"  # Default
  ferrocene.document-signatures-s3-bucket = "ferrocene-document-signatures"  # Default
  ```

* To fetch the signatures from a documentation tarball, you need to configure:

  ```toml
  ferrocene.document-signatures = "docs-tarball"
  ferrocene.document-signatures-tarball = "path/to/ferrocene-document-signatures.tar.xz"
  ```
  
* To disable document signatures handling:

  ```toml
  ferrocene.document-signatures = "disabled"
  ```


src/bootstrap/src/ferrocene/sign/cosign.rs
```diff
@@ -0,0 +1,101 @@
...
+        for candidate in COSIGN_ARTIFACTS {
+            if candidate.target == &*builder.config.build.triple {
+                artifact = Some(candidate);
+                break;
+            }
+        }
+        let Some(artifact) = artifact else {
+            eprintln!();
+            eprintln!("error: unsupported platform for cosign: {}", builder.config.build);
+            eprintln!("note:  add support for it in src/bootstrap/ferrocene/sign.rs");
```
tshepang, line 59: maybe...
```suggestion
            eprintln!("note:  add support for it in {}", file!());
```

Urhengulas: It's not in the same file

Urhengulas: The code is in ` src/bootstrap/ferrocene/cosign.rs`, while the message tells to put it into `src/bootstrap/ferrocene/sign.rs` (the parent module).

ferrocene/tools/document-signatures/src/main.rs
```diff
@@ -42,28 +43,45 @@ fn main() -> Result<(), Error> {
...
     }
 }
 
-fn env<T>(var: &str) -> Result<T, Error>
+fn maybe_env<T>(var: &str) -> Result<Option<T>, Error>
 where
     T: FromStr,
     T::Err: Send + Sync + std::error::Error + 'static,
 {
     let var = format!("DOCUMENT_SIGNATURES_{var}");
```
Urhengulas, line 65:
```suggestion
    let var = env_name(var);
```

ferrocene/tools/document-signatures/src/signature_files.rs
```diff
@@ -75,12 +75,15 @@ impl<'opts> SignatureFiles<'opts> {
...
+        let Some(s3_bucket) = &self.options.s3_bucket else {
+            panic!("uploading signatures is only supported with the s3 backend");
+        };
         let uuid = Uuid::new_v4();
 
         // First off, we upload the file to S3, named after the UUID.
         let mut command = Command::new("aws")
             .args(["s3", "cp", "-"])
-            .arg(format!("s3://{}/{uuid}", self.options.s3_bucket))
+            .arg(format!("s3://{}/{uuid}", s3_bucket))
```
Urhengulas, line 86:
```suggestion
            .arg(format!("s3://{s3_bucket}/{uuid}"))
```

src/bootstrap/src/ferrocene/sign/cosign.rs
tshepang: the parent module is `sign`, but the file `sign.rs` has been removed... the logic is in `mod.rs`

src/bootstrap/src/core/config/config.rs
```diff
@@ -1978,13 +1990,49 @@ impl Config {
...
+                    FerroceneDocumentSignatures::S3 { bucket }
+                }
+                (None | Some("s3"), None, None) => FerroceneDocumentSignatures::S3 {
+                    bucket: "ferrocene-document-signatures".into(),
+                },
+                (None | Some("s3"), _, _) => panic!(
+                    "ferrocene.document-signatures=\"s3\" only supports \
+                     ferrocene.document-signatures-s3-bucket as the additional \
+                     setting related to document signatures"
+                ),
```
Urhengulas, line 2010: This is the case when there is a tarball path given, but no s3 bucket, right? It took me a few seconds to think that through. Maybe following is slightly more clear?

```suggestion
                (None | Some("s3"), None, Some(_)) => panic!(
                    "ferrocene.document-signatures=\"s3\" only supports \
                     ferrocene.document-signatures-s3-bucket as the additional \
                     setting related to document signatures"
                ),
```

src/bootstrap/src/core/config/config.rs
```diff
@@ -1978,13 +1990,49 @@ impl Config {
...
+
+                (Some("docs-tarball"), None, Some(tarball)) => {
+                    FerroceneDocumentSignatures::DocsTarball { tarball }
+                }
+                (Some("docs-tarball"), None, None) => panic!(
+                    "ferrocene.document-signatures=\"docs-tarball\" requires setting \
+                     ferrocene.document-signatures-tarball"
+                ),
+                (Some("docs-tarball"), _, _) => panic!(
```
Urhengulas, line 2019: Same here
```suggestion
                (Some("docs-tarball"), Some(_), None) => panic!(
```

src/bootstrap/src/ferrocene/doc.rs
```diff
@@ -303,32 +299,35 @@ impl<P: Step> Step for SphinxBook<P> {
...
                 // do a full build of the documents we want.
-                builder.info(&format!("Gathering references of {}", self.src));
+                builder.info(&format!("Gathering references of {}", P::SOURCE));
 
                 // When gathering the objects.inv files, we use a custom Sphinx builder that only
                 // outputs the objects.inv file.
                 cmd.args(&["-b", "ferrocene-intersphinx"]);
+
+                // Avoid circular dependencies.
+                allow_injecting_ids = false;
```
Urhengulas, line 309: When is is set to true? It is initialized as `false`

src/bootstrap/src/ferrocene/sign/cosign.rs
Urhengulas: But apparently `sign.rs` was deleted, but now we have `sign/mod.rs` 🤔

Urhengulas (changes requested):
First part of the review. Didn't look at `src/bootstrap/src/ferrocene/sign/*.rs` yet.

src/bootstrap/src/ferrocene/sign/cosign.rs
Urhengulas: > the parent module is `sign`, but the file `sign.rs` has been removed... the logic is in `cosign.rs`

Yeah, but I am unsure if it meant to be `sign/cosign.rs` or `sign/mod.rs`.

pietroalbini: This got messed up by the rename. I'd rather not use `file!()`, as that includes absolute paths and could cause things to get messy.

ferrocene/tools/document-signatures/src/main.rs
pietroalbini: Good catch!

src/bootstrap/src/ferrocene/sign/cosign.rs
tshepang: yeah... it could have been `sign.rs`, as in this is also valid structure

```
src/bootstrap/src/ferrocene
├── sign.rs
├── sign
│  ├── cosign.rs
│  └── signature_files.rs
...

File: src/bootstrap/src/ferrocene/doc.rs (hunk 6 above)
Comment (pietroalbini) [reply src/bootstrap/src/ferrocene/doc.rs:309]: Good catch!

File: src/bootstrap/src/core/config/config.rs (hunk 4 above)
Comment (pietroalbini) [reply src/bootstrap/src/core/config/config.rs:2010]: This is the case where you configured the `s3` mode, but you also used configuration keys the `s3` directive doesn't know how to handle. It's written this way to require minimal changes when a new mode will be added in the future.

Right now we have something like this (pseudocode):

```rust
match (mode, s3_bucket, docs_tarball) {
    ("s3",           None,         None         ) => ok(),
    ("s3",           Some(bucket), None         ) => ok(),
    ("s3",           _,          , _            ) => err(),

    ("docs-tarball", None,       , Some(tarball)) => ok(),
    ("docs-tarball", None,       , _            ) => err(),

    ("disabled",     None,       , None         ) => ok(),
    ("disabled",     _,          , _            ) => err(),

    (_,              _,          , _            ) => err(),
}
```

The purpose of the `("mode", _,  _)` lines is to handle all other possibilities that the mode doesn't support.

We *could* right now do (for example) `("s3", None | Some(_), Some(_))`, where the first `None | Some(_)` handles both values for the supported option and `Some(_)` handles the value that should not be there.

This would break as soon as we add a new config option though. Then we'd need to have `("s3", None | Some(_), None | Some(_), None | Some(_))`: the first has to be `None | Some(_)` as before, and the next two have to be `None | Some(_)` because either of them could be set. And that is equivalent to `("s3", _, _, _)` :slightly_smiling_face:

src/bootstrap/src/core/config/config.rs
pietroalbini: See https://github.com/ferrocene/ferrocene/pull/689#discussion_r1636565860

pietroalbini:
Addressed all review feedback.

Urhengulas (approved):
Sounds good to me. I ended up reviewing it commit by commit, since the `src/bootstrap/src/ferrocene/sign/*` files made my head spin a bit.

Are there any kind of tests for this code? Or a way to manually test it? Or some CI logs to review to make sure it works correctly?

I have two small suggestions, but both are optional.

pietroalbini:
Let's merge it as-is for now.

> Are there any kind of tests for this code? Or a way to manually test it? Or some CI logs to review to make sure it works correctly?

It's quite hard for people that are not me or Flo to test this :frowning_face: I did test it locally with my creds though.

pietroalbini:
bors merge

PR #694, opened 2024-06-13
Adopt uv for venv purposes

Hoverbear:
Adopt `uv` and set up a `venv` so local users don't need to set up a python environment beyond `python`/`pip` (eg, no `boto`, no `pyyaml`, etc).

As a consequence, the way we build our docker images has changed, as they need to have access to `requirements.txt`.

This PR, surprisingly, needs no documentation updates, as it does not impact and customer facing procedures yet.


ferrocene/ci/docker-images/ubuntu-20/requirements.txt
```diff
@@ -0,0 +1,39 @@
+# This file was autogenerated by uv via the following command:
```
Hoverbear, line 1: Hmmm... I did actually make these hardlinks. I think I need to fiddle with this more.

Hoverbear:
bors try

ferrocene/ci/docker-images/ubuntu-20/Dockerfile
```diff
@@ -149,3 +146,12 @@ RUN mkdir /home/ci \
     && chown -R ci: /home/ci
 USER ci
 WORKDIR /home/ci
+
+# We prefer uv over other solutions, it's in Rust and fast
+ENV PATH="/home/ci/.local/bin:$PATH"
+RUN python3 -m pip install uv
```
pietroalbini, line 152: Is the `pipx` in the Ubuntu 20.04 archives too old?

pietroalbini (commented):
Looks good, only one question.

ferrocene/ci/docker-images/ubuntu-20/Dockerfile
Hoverbear: It seemed so to me. I tried to use the same instructions for both and had failures. 😵

pietroalbini (approved):
bors merge

requirements.txt
```diff
@@ -0,0 +1,42 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+# This file was autogenerated by uv via the following command:
+#    uv pip compile - -o requirements.txt
```
zanieb, line 5: Just wondering, where do your input requirements come from?

ferrocene/ci/docker-images/ubuntu-20/Dockerfile
zanieb, line 152: One thing you can do here is copy a built `uv` from our base image instead of installing it with `pip` or `pipx` (example in https://github.com/astral-sh/uv/issues/4385)

requirements.txt
Hoverbear: This was done by following the instructions from https://github.com/astral-sh/uv?tab=readme-ov-file#getting-started

image

Is there a better way to approach this? (I've hardly touched python in over a decade)

ferrocene/ci/docker-images/ubuntu-20/Dockerfile
Hoverbear: I was thinking of installing `uv` from your curl command, but that'd work too I think! I think `uv toolchain` can replace this: https://github.com/ferrocene/ferrocene/blob/40f588b69a7af489c78c438f15e11a5f6c2f412d/ferrocene/ci/docker-images/ubuntu-20/Dockerfile#L74-L85

zanieb: Oh you build from source! Eek. Yes you could use `uv toolchain` to fetch a Python. There are some [behavioral quirks](https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html) to the standalone builds, but usually it's great — I'm not sure about your exact use case.

requirements.txt
zanieb: Generally it's best practice to declare your dependencies and some constraints  in a `requirements.in` file then compile those to a `requirements.txt` so you have the input requirements checked in to your repository.

PR #712, opened 2024-06-19
Stripped white line to avoid conflict, same as beta.

Dajamante:
https://github.com/ferrocene/ferrocene/pull/710


Urhengulas (approved):
bors merge

PR #714, opened 2024-06-19
Add more annotations for single tested only paragraphs

Veykril:
This just adds a few more annotations for some that only have a single test associated (there are still more)


pietroalbini (approved):
bors merge

PR #723, opened 2024-06-21
Reformat tables in rustc tool-analysis

Urhengulas:
# Motivation

I find it hard to view the tables in the rustc tool analysis. One problem are the very wide columns. Therefore this PRs aims to reduce the column width. In turn it introduces some whitespace, which isn't pretty either, but okayish.

Obviously it would be better if this formatting would be automated, because manual formatting takes quite a lot of time. Either with a script or with CSS.

# Changes

- [x] arrange IDs vertically
- [x] remove `:widths:` (because it seems to have no effect)
- [ ] ~~manually format columns~~
- [x] Move use cases out of table if all use cases of all potential errors are the same (which is the case for all of the tables).

# Before

Screenshot from 2024-06-21 16-30-31
Screenshot from 2024-06-21 16-30-18

# After

Screenshot from 2024-07-08 14-04-54
Screenshot from 2024-07-08 14-04-38

### DIFF
--- a/ferrocene/doc/evaluation-plan/src/method.rst
+++ b/ferrocene/doc/evaluation-plan/src/method.rst
@@ -93,15 +93,15 @@ HazOp analysis:
 
 .. list-table::
 
-   * - Error identifier 
+   * - Error identifier
      - ERR_TOOLID_XXX where TOOLID is the name of the tool causing the error, and XXX a number.
-   * - Use case 
+   * - Use case
      - The use case where the error is determined.
-   * - Description 
+   * - Description
      - The potential error descriptions.
-   * - Risk 
+   * - Risk
      - Possible effects of the potential error.
-   * - Mitigation 
+   * - Mitigation
      - The detection or mitigation method identifier, if available.
    * - Detectable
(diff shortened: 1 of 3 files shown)


pietroalbini:
So, here I would not add manual line breaks in the text, its maintenance is fairly high and can mess with accessibility (as it introduces screen readers).

A thing we can do to still improve the layout of the page is to reorder the columns, so that the columns with the text users care about is closest to the left side of the screen. Possible ordering:
  * Error identifier
  * Description
  * Risk
  * Detectable
  * Mitigation
  * Use case

Urhengulas:
I realized that for each table all the use cases were the same. So I moved it out of the table.

This would obviously not work if a single potential error had different use cases. But this is not the case.

pietroalbini:
Let's not move the use cases out.

Urhengulas:
@pietroalbini said:
> Let's not move the use cases out.

Why not? It does save a lot of repetition (and table width).

Urhengulas:
Close, as it will become obsolete with sphinx-needs

PR #727, opened 2024-06-22
ep: we now have 2 qualified releases

tshepang:


pietroalbini (approved):
bors merge

pietroalbini:
We shouldn't backport this to 1.76, that is already qualified.

PR #742, opened 2024-07-03
Only try serving the docs when building html docs

pietroalbini:
Serving multiple docs at the same time is not supported by the build system, since we can only have one server running at the time. To enforce that, there is a check in bootstrap ensuring the `builder.should_serve()` function is only called once.

A previous refactor broke the flag though, because it resulted in the `builder.should_serve()` method being called by all `objects.inv` gathering steps in addition to the html building step. This PR fixes the problem.


tshepang (approved):
bors merge

PR #747, opened 2024-07-05
[1.79] Add Ferrocene and Rust version to the Safety Manual

pietroalbini:
This is required by ISO 26262 part 8 clause 11.4.4.1.a, which links to the environment page in our [standard traceability page](https://public-docs.ferrocene.dev/main/qualification/plan/standards/iso-26262.html). I discovered we are not doing that during a routine reading of the documents.

For now this fix just targets the next 24.08 release, I'll send a larger refactoring to the main branch in a followup PR.

PR #748, opened 2024-07-05
Merge Environment and User Responsibility page of the Safety Manual

pietroalbini:
Originally my plan was to just backport https://github.com/ferrocene/ferrocene/pull/747, but I also took the occasion to refactor a bit those pages of the safety manual.

I merged "environment" and "user-responsibility" into a single page (as both of them were quite short), and:

* Added the Ferrocene, rustc and rustfmt version numbers as required by ISO 26262 11.4.6.2.a and 11.4.4.1.a.
* Turned "The qualification scope is limited to the set of supported compilation options
described in the Tool Options" into a bullet list outlining the scope for each tool.


Urhengulas (approved):
bors merge

PR #749, opened 2024-07-05
Add `no_traceability_matrix` flag to `cli:program` directive

Urhengulas:
We do not want to (i.e. need to) test the cli options of `rustfmt`, therefore this PR allows disabling the generation of entries in the traceability matrix. It also sets the flag for the rustfmt cli.

Before there was a red "UM: rustfmt Command-Line Interface" entry. After this PR it will be gone.

Before: Screenshot from 2024-07-05 13-10-06

After: Screenshot from 2024-07-05 13-10-20

### DIFF
--- a/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
+++ b/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
@@ -27,6 +27,5 @@ def setup(app):
         #
         # Version history:
         # - 0: initial implementation
-        # - 1: add no_traceability_matrix flag
-        "env_version": "1",
+        "env_version": "0",
     }
--- a/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
+++ b/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
@@ -27,5 +27,6 @@ def setup(app):
         #
         # Version history:
         # - 0: initial implementation
-        "env_version": "0",
+        # - 1: add no_traceability_matrix flag
+        "env_version": "1",
     }
--- a/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
+++ b/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
@@ -1,14 +1,19 @@
 # SPDX-License-Identifier: MIT OR Apache-2.0
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
+
+from dataclasses import dataclass
+import re
+import string
+
 from docutils import nodes
+from docutils.parsers.rst import directives
+
 from sphinx import addnodes
 from sphinx.directives import SphinxDirective, ObjectDescription
 from sphinx.domains import Domain, ObjType
 from sphinx.roles import XRefRole
-import re
 import sphinx
-import string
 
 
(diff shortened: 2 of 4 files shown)


ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
```diff
@@ -27,5 +27,6 @@ def setup(app):
         #
         # Version history:
         # - 0: initial implementation
-        "env_version": "0",
+        # - 1: add no_traceability_matrix flag
+        "env_version": "1",
```
Urhengulas, line 31: Is is actually necessary to bump the `env_version`, since CI builds the docs from scratch anyways?

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
```diff
@@ -18,17 +22,28 @@ class ProgramDirective(SphinxDirective):
...
         if PROGRAM_STORAGE in self.env.temp_data:
             warn("cli:program inside cli:program isn't supported", self.get_location())
             return []
-        self.env.temp_data[PROGRAM_STORAGE] = self.arguments[0]
 
+        # store arguments, so they can be accessed by child `OptionDirective`s
+        self.env.temp_data[PROGRAM_STORAGE] = (
+            self.arguments[0],
+            "no_traceability_matrix" in self.options,
+        )
```
pietroalbini, line 38: Nit: can we define a dataclass here, so that we can have named fields for program name and no_traceability_matrix?

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/__init__.py
pietroalbini: So, in general yes, whenever we change the environment we should increase the env version. You are right that in CI we build the docs from scratch, but we also build docs incrementally on our developer machines too. Without incrementing the environment version we would get confusing error messages locally.

*But* in this specific case we don't need to bump the env version. We only changed `env.temp_data`, which is not persisted to disk between incremental builds and is cleared as soon as reading the document finished.

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
Urhengulas: Sure

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
```diff
@@ -149,3 +171,9 @@ def warn(message, location):
 
 def setup(app):
     app.add_domain(CliDomain)
+
+
+@dataclass
+class ProgramStorage:
+    program_name: str = "PLACEHOLDER"
+    no_traceability_matrix: bool = False
```
pietroalbini, line 179: We don't have to provide default values, just this would work:

```suggestion
    program_name: str
    no_traceability_matrix: bool
```

pietroalbini (changes requested):

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
Urhengulas: 👍🏾

pietroalbini (approved):
bors merge

PR #762, opened 2024-07-09
Simplify the traceability matrix

pietroalbini:
This PR includes a lot of changes I wanted to make to the traceability matrix for a long time, and I finally got annoyed enough to implement them:

* Completely removed paragraphs from the traceability matrix. Those were added originally while we were still planning to link tests to individual paragraphs, but as we are not doing that there is little point in bloating our documents with useless lines.

* Changed informational sections to be always hidden, with no way to show them back. They have been replaced with a simple count of the informational sections. This continues to reduce the bloat by not including 700+ lines for each glossary entry.

* Removed the opt-in checkbox for annotations mode, and made it the default. With the paragraphs and the glossary sections removed there is little benefit for that to be opt-in.

* Added a link back to the documentation index.

### DIFF
--- a/ferrocene/tools/traceability-matrix/src/documentations.rs
+++ b/ferrocene/tools/traceability-matrix/src/documentations.rs
@@ -40,9 +40,7 @@ pub(crate) struct Section {
 
 #[derive(Debug, Deserialize)]
 pub(crate) struct Paragraph {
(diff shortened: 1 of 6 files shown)


Veykril (approved):

Veykril:
bors merge

PR #758, opened 2024-07-08
Use a python script for untarring qnx

Hoverbear:
We have thusfar held off testing QNX on Windows since untarring it on CircleCI's Windows runners takes.... over 43 minutes. We're hoping moving to Python will avoid that as it did in other places.

If this *doesn't* help, then I guess CircleCI just doesn't provide Windows Runners with enough IOPS for this to make sense.

PR #771, opened 2024-07-10
Setup venvs on Mac and Windows as well.

Hoverbear:
While working on #758 I realized we don't setup our venvs on Mac/Windows. *Whoops*. This fixes that.


Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

ferrocene/ci/scripts/setup-venv.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+set -xeuo pipefail
+
+if command -v uv &> /dev/null; then
+    echo "uv was already installed, skipping"
+    exit 0
+fi
```
Urhengulas, line 10: Is there a case where `uv` could be installed, but no venv has been set up? Iiuc the script currently assumes either both or none are true.

pietroalbini (approved):

ferrocene/ci/scripts/setup-venv.sh
Hoverbear: I think it's possible that this could happen, yeah. I wasn't really sure how thorough to be since theoretically if this script has run already, it completed wholly (if it failed, the CI job would fail and be rejected). This likely isn't going to be run by us developers.

Hoverbear:
bors merge

ferrocene/ci/scripts/setup-venv.sh
Urhengulas: Oki ☺️

Hoverbear:
(Barkporting since this is needed for qnx stuff to work)

Hoverbear:
Seems this isn't needed for that, removing the backport label.

PR #783, opened 2024-07-17
Merge `ferrocene_intersphinx_support` in `ferrocene_qualification`

pietroalbini:
This also initializes `sphinx.ext.intersphinx` inside of `ferrocene_qualification`, so that there is no need to depend from the extension manually. This is a pre-requisite for the `sphinx-needs` integration.


Urhengulas (approved):
bors merge

PR #784, opened 2024-07-17
Document how to run CI for external contributions

Urhengulas:
I am unsure if the Qualification Plan (QP) is the right place to document this? Initially I thought about the Internal procedures, but they just link to the QP for the Product engineering procedures.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/external-contributions.rst
+++ b/ferrocene/doc/internal-procedures/src/external-contributions.rst
@@ -1,0 +1,49 @@
+.. SPDX-License-Identifier: MIT OR Apache-2.0
+   SPDX-FileCopyrightText: The Ferrocene Developers
+
+External contributions
+======================
+
+External contributions are desired, but CI needs to be protected against
+arbitrary code execution. Therefore CI does not automatically execute for PRs
+from forks. Before executing, confidence in the code needs to be established.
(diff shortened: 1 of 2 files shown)


ferrocene/doc/qualification-plan/src/development.rst
```diff
@@ -208,3 +208,42 @@ current status of a branch after the change.
 The GitHub issue is used to track the analysis performed by either the original
 author of the patch or the Release Manager, post mortems and conclusions, and
 any additional tests introduced to prevent future recurrences of the problem.
+
+
+External contributions
+----------------------
+
+CircleCI does not run PRs from forks by default, which is good because it
```
Hoverbear, line 216: This section doesn't fit the overall tone/writing style for the QP. I think you are right that this should go into internal procedures.

Perhaps it can be added to the `working-with-the-ci.rst`?

Urhengulas: Should we adapt the tone or should we put it in a different place?

Urhengulas: I moved it to a new page in the internal procedures.

Urhengulas: And reworded it.

Hoverbear (approved):
bors merge

PR #794, opened 2024-07-20
those are for rustc

tshepang:


ferrocene/doc/qualification-plan/src/standards/iso-26262.rst
```diff
@@ -73,11 +73,11 @@ ISO 26262 requirements
 
    * - 11.4.4.2.e
      - :doc:`Safety Manual - Known problems <safety-manual:known-problems>`
-     - See ``rustc`` column
+     - N/A
```
tshepang, line 76: not sure about this... we lack the doc for rustfmt defect list, but I suspect it's not needed, since there are mitigation strategies (manual review and testing) should it create dangerous code

pietroalbini:
We should still track the known problems in problems.ferrocene.dev, this is accurate.

PR #796, opened 2024-07-21
rustfmt has no qualification methods

tshepang:


Urhengulas:
Are you saying 11.4.6.2 only needs to be filled if we applied a qualification method? If so, then we should put "N/A" in all the subsections.

Urhengulas (changes requested):
Are you saying 11.4.6.2 only needs to be filled if we applied a qualification method? If so, then we should put "N/A" in all the subsections.

tshepang:
I think it's just 11.4.6.2f-h that needs to be omitted... the rest look applicable

Urhengulas (approved):
bors merge

PR #801, opened 2024-07-22
It's qualified to use `rust-lld` if you play by the rules

Hoverbear:
Note that `rust-lld` is, in fact, qualified so long as you use it within the rules laid out in https://public-docs.ferrocene.dev/main/safety-manual/rustc/options.html#linker-options.


pietroalbini:
This isn't correct. Our approach for qualification is to treat everything as a black box: we never qualified rust-lld, what we qualified is that if you use the entire toolchain as we ship it (from source code to the executable) the resulting whole toolchain is qualified, without caring about the components we use within it.

Saying that rust-lld is qualified (even with those constraints) would instead mean we did all the qualification activities for rust-lld.

PR #802, opened 2024-07-23
Test sphinx-shared-resources on pull requests to monorepo

Urhengulas:
Before the workflow was run in [ferrocene/sphinx-shared-resources](https://github.com/ferrocene/sphinx-shared-resources/), but since we don't make changes in that repo anymore, and instead make changes to the monorepo and sync them, it was not run anymore.

@Hoverbear Is this okay as a Github Action or does it need to be converted to CircleCI? I would want this to block a PR if it fails.


Urhengulas:
Blocked by @pietroalbini

Urhengulas:
Close as work was paused.

PR #814, opened 2024-07-30
fix accepted linker option

tshepang:


Veykril (approved):

Veykril:
bors merge

jonathanpallant:
I'm leaving a link here to the LD user manual, which does indeed confirm that I spelled the linker option incorrectly: https://sourceware.org/binutils/docs/ld/Options.html

PR #817, opened 2024-07-31
Allow building ferrocene docs on windows

Veykril:
Canonicalizing paths yields [extended length paths](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file) on windows causing the paths to differ significantly if one of the two doesn't exist. From what I can tell we mainly want to get rid of any potential `..` and `.` segments from the paths though (and make them absolute), not resolve symlinks, so absolutizing should suffice.

Additionally this sets the encoding for text file `open` calls to `utf8` as that defaults to `cp1252` on windows.

### DIFF
--- a/ferrocene/doc/document-list/exts/ferrocene_document_list.py
+++ b/ferrocene/doc/document-list/exts/ferrocene_document_list.py
@@ -18,7 +18,7 @@ class DocumentId(SphinxRole):
             return [], [self.error(self.text)]
         if not os.path.exists(path):
             return [], [self.error(path)]
-        with open(path) as file:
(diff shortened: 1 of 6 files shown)


tshepang:
why the utf-8 encoding

Veykril:
Was just editing the reason into the PR message, you were too fast for me :)

tshepang:
is your normalize different from `Path::absolute` (introduced in 1.79)

Veykril:
Oh, TIL about that function. Yes I think we should use that

tshepang (approved):
bors merge

PR #825, opened 2024-08-02
Add x86_64 Darwin

Hoverbear:


Hoverbear:
bors try

Hoverbear:
It turns out to be quite difficult to setup x86_64 Homebrew and pass `arch -x86_64` to everything. `arch -x86_64 zsh` may work, however since the Mac runners use `mac:` not `executor:` I'm unsure how to do that yet.

PR #836, opened 2024-08-12
[try] qnx: unignore failing `std::process` tests

japaric:
I have been unable to get these to fail locally after rebasing. Let's see if that's the case in CI


japaric:
bors try

PR #849, opened 2024-08-19
Remove stale exceptions from requirements-traceability

Veykril:
If https://github.com/ferrocene/ferrocene/pull/842 lands, the two tests are in fact green in the matrix


pietroalbini (approved):
bors merge

PR #855, opened 2024-08-20
keep library/ lockfile fresh

tshepang:
library/ has been moved to a separate workspace, so it has its own lockfile that needs to be kept updated


Veykril (approved):

Veykril:
bors merge

PR #867, opened 2024-08-22
Add commit-checks to GHA

Hoverbear:
Migrates our commit checks to GHA (using Azure).

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -4,8 +4,8 @@
 # Gate PRs on the `full` CircleCI workflow
 status = ["full"]
 
-# Wait until the `commit` CircleCI workflow passes before queueing a PR
-pr_status = ["commit"]
+# Wait until the `commit` GHA workflow passes before queueing a PR
+pr_status = ["Commit / %"]
 
 # Consider the build as failed if it takes more than three hours to finish
 # A normal build should take less, but there are slower parts of the build that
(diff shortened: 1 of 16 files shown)


ferrocene/ci/scripts/github_app_auth.py
```diff
@@ -54,6 +46,23 @@ def req(session, method, url):
 
     token = req(http, "POST", installations[0]["access_tokens_url"])["token"]
 
+    # We never want to print this on CI.
+    if 'CI' in os.environ and os.environ['CI'] == True:
+        print(f"::add-mask::{token}")
+    
+    return token
+
+if __name__ == "__main__":
```
Hoverbear, line 55: These changes are unnecessary and not strictly related to this PR, but act as general cleaning as I was previously using them in an earlier revision.

I think it's worth including this despite being unrelated.

bors.toml
```diff
@@ -4,8 +4,8 @@
 # Gate PRs on the `full` CircleCI workflow
 status = ["full"]
 
-# Wait until the `commit` CircleCI workflow passes before queueing a PR
-pr_status = ["commit"]
+# Wait until the `commit` GHA workflow passes before queueing a PR
+pr_status = ["Commit / %"]
```
Hoverbear, line 8: According to https://bors.tech/documentation/, this should be allowed.

Hoverbear:
I wonder if we can somehow get rid of this: 
image

PR #876, opened 2024-08-27
Use GHA for commit checks

Hoverbear:
Based off https://github.com/ferrocene/ferrocene/pull/867, but requires less trust, thus providing contributors outside the project with a (hopefully) frictionless experience.


bors.toml
```diff
@@ -4,8 +4,8 @@
 # Gate PRs on the `full` CircleCI workflow
 status = ["full"]
 
-# Wait until the `commit` CircleCI workflow passes before queueing a PR
-pr_status = ["commit"]
+# Wait until the `commit` GHA workflow passes before queueing a PR
+pr_status = ["Commit / %"]
```
pietroalbini, line 8: I'm not sure if this is what we want:

* If we have only one job, the wildcard is not needed.
* If we have more than one job, there could be cases where *at a point in time* all jobs are green just because other ones have not started yet (with job dependencies).

I'd either hardcode `Commit / Checks`, or create a "build finished" job like we do elsewhere using bors.

pietroalbini (approved):
bors merge

pietroalbini:
bors cancel

pietroalbini:
We still need to have the commit job in this PR: bors uses the `bors.toml` of the branch it merges into (in this case `main`), and so it's still expecting the CircleCI job. We can keep it in for this PR, and then send a followup PR removing it.

Hoverbear:
Ah, heck!

ferrocene/ci/scripts/setup-ubuntu.sh
```diff
@@ -0,0 +1,31 @@
...
+    apt upgrade -y
+fi
+
+apt install -y \
+    git \
+    build-essential \
+    ninja-build \
+    zlib1g-dev
+
+if [[ -z "${INSTALL_LLVM}" ]]; then
```
jonathanpallant, line 25: I assume I'm having a moment, but doesn't `-z` test if a variable is empty (or of **z**ero length)? So doesn't this install llvm-17-tools and llvm-17-dev if and only if INSTALL_LLVM is of zero length, and won't install llvm tools is INSTALL_LLVM is not of zero length?

Likewise above, doesn't it only run `apt upgrade` if and only if SKIP_UPGRADE is of non-zero length? That seems backwards.

Hoverbear: Indeed you're right. I don't think I tested this enough.

pietroalbini (approved):

pietroalbini:
bors merge

PR #877, opened 2024-08-27
Move `gha-app-auth.py` to scripts as `gha_app_auth.py`

Hoverbear:
This kinda fell out of some work with #876 #867 and just brings the `gha-app-auth.py` script into the fold so other scripts can call it.


pietroalbini (approved):
bors merge

PR #880, opened 2024-08-28
Allow specifying which subtree to pull

Veykril:


ferrocene/tools/pull-subtrees/pull.py
```diff
@@ -385,6 +385,7 @@ def run_capture(*args, **kwargs):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--target", help="branch this pull is targeting", default=None)
+    parser.add_argument("--subtree", help="subtree to pull", default=None)
```
Dajamante, line 388: when will we use this flag?

Dajamante (approved):
bors merge

ferrocene/tools/pull-subtrees/pull.py
Dajamante: @Veykril how will this be used?

Veykril: When you don't want to pull all subtrees at once

PR #883, opened 2024-08-28
Allow running `./x.py test compiler`

Veykril:


src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -2441,18 +2441,32 @@ impl Step for CrateLibrustc {
     const ONLY_HOSTS: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
-        run.crate_or_deps("rustc-main")
+        run.crate_or_deps("rustc-main").path("compiler")
```
Veykril, line 2444:
```suggestion
        run.crate_or_deps("rustc-main").alias("compiler")
```
would be nicer for the `make_run` logic (just check if paths are empty) but alias checks if the input exists as a path and complains then so

Veykril:
Submitted this upstream, though we no longer need it

PR #890, opened 2024-08-29
Make `./x.py <cmd> compiler/<crate>` feature aware

Veykril:


Veykril:
Will upstream this either way even though it doesnt really fix the general issue

PR #892, opened 2024-08-29
Introduce quality managed targets

Hoverbear:


ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -55,6 +55,36 @@ code change.
      - Bare-metal
      - \-
 
+Quality managed targets
```
tshepang, line 58: sounds more correct to say *quality-managed* (instead of *quality managed*)

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -55,6 +55,36 @@ code change.
...
+-----------------------
+
+Quality managed targets are targets which are suitable for use in production 
+outside of safety critical contexts. Support is available. The Ferrocene test
+suite successfully executed on the target. Known problems are tracked and
+available.
+
+For any of the following reasons, the target is not qualified:
+
+* The target is deemed unlikely to be part of a safety critical context.
```
tshepang, line 68: would it be more clear (and retain same meaning) if we said *to be used in a safety-critical context*

ferrocene/doc/user-manual/src/targets/index.rst
Hoverbear: I have requested clarification from the relevant persons who approved this wording (Quality managed) to see if they meant your suggestion.

Hoverbear: The relevant persons have rejected this suggestion, "Quality Managed" is the approved term.

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -26,14 +26,18 @@ There are also two variants of the standard library available:
...
 -----------------
 
-Supported targets are provided with the highest level of assurance. They are
+Qualified targets are provided with the highest level of assurance. They are
 qualified for use in safety-critical contexts (when used with a qualified
 rustc), and the full Ferrocene test suite is executed on the target for every
 code change.
 
+Only stable releases of qualified targets are qualified. Other releases, such
+as beta, should be considered Quality Managed. Specific releases can be
```
tshepang, line 38: to make it clear we are talking about the same ones mentioned in the previous sentence
```suggestion
as beta, should be considered Quality Managed. Such releases can be
```

tshepang (approved):
bors merge

PR #894, opened 2024-08-30
Only update `library/Cargo.lock` in subtree pulls

Veykril:
There is no reason for us to currently constantly updating all entries in this lock file to their latest semver release. This can create unnecessary problems and the only entry in the lockfile that is relevant for us to keep up to date is the one for our subtree(s), so we should only be updating that one such that it is always in sync with the subtree version number.

This PR resets our `library/Cargo.lock` back to upstream to remove divergence there

### DIFF
--- a/ferrocene/tools/pull-subtrees/subtrees.yml
+++ b/ferrocene/tools/pull-subtrees/subtrees.yml
@@ -31,5 +31,3 @@
 - path: ferrocene/library/backtrace-rs
   repo: rust-lang/backtrace-rs
   ref: master
-  after:
-    - update-cargo-lock backtrace library
--- a/ferrocene/tools/pull-subtrees/subtrees.yml
+++ b/ferrocene/tools/pull-subtrees/subtrees.yml
@@ -26,10 +26,10 @@
   repo: rust-lang/libc
   ref: libc-0.2
   after:
-    - update-cargo-lock
+    - update-cargo-lock libc library
 
 - path: ferrocene/library/backtrace-rs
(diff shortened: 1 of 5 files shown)


ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+git #!/bin/bash
```
tshepang, line 1: stray

ferrocene/tools/pull-subtrees/subtrees.yml
```diff
@@ -26,10 +26,10 @@
...
   after:
-    - update-cargo-lock
+    - update-cargo-lock libc library
 
 - path: ferrocene/library/backtrace-rs
   repo: rust-lang/backtrace-rs
   ref: master
   after:
-    - update-cargo-lock
+    - update-cargo-lock backtrace library
```
Veykril, line 35: We aren't yet using the subtree in `library` I think, but we will likely do that soon

Veykril: actually nevermind, this is a patch dependence, it doesn't even show in the library lock file

ferrocene/tools/pull-upstream/pull.sh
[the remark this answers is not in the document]
Veykril: Oh yes, I only removed the comment line 🤦

tshepang (approved):
bors merge

PR #903, opened 2024-09-02
some doc nits

tshepang:


ferrocene/doc/user-manual/src/overview.rst
```diff
@@ -6,11 +6,7 @@ Overview
 
 This Manual describes the use of Ferrocene |ferrocene_version|, the |iso_ref|
 and |iec_ref| qualified version of the Rust toolchain.
-
-Ferrocene is based on:
-
-* rustc version |rust_version|.
-* rustfmt version |rustfmt_version|.
+Ferrocene is based on ``rustc`` version |rust_version|.
```
pietroalbini, line 9: The rustfmt version is different than the rustc version.

pietroalbini (changes requested):

ferrocene/doc/user-manual/src/overview.rst
tshepang: the versions for each are specified [in SM](https://public-docs.ferrocene.dev/main/safety-manual/scope.html), so rustfmt does not need to be mentioned here

pietroalbini: What harm does it do to mention it here?

tshepang: it is another place to remember to mention tools that are qualified (which means it might get stale), when one already exists

the inital reason I decided to remove this is I was about to add `ferrocene-self-test`, and realized that the line "Ferrocene is based on" is not even right, and I saw that the tools are mentioned elsewhere anyways

pietroalbini (approved):
bors merge

PR #907, opened 2024-09-02
quit when unused args are given

tshepang:


ferrocene/tools/self-test/src/error.rs
```diff
@@ -113,6 +115,7 @@ impl Error {
             Error::DuplicateTargetLibrary { .. } => 9,
             Error::TargetLibraryDiscoveryFailed { .. } => 10,
             Error::CCompilerNotFound { .. } => 11,
+            Error::CliArgsProvided => 12,
```
pietroalbini, line 118: We also need to list the error code in https://public-docs.ferrocene.dev/main/user-manual/self-test/error-codes.html.

pietroalbini (changes requested):

ferrocene/tools/self-test/src/error.rs
tshepang: https://github.com/ferrocene/ferrocene/pull/907/commits/f9ba66227d85d938e28680641c81f2aadd915e3b

ferrocene/ci/scripts/run-self-test.sh
```diff
@@ -70,5 +71,12 @@ for archive in ${TEMPDIR}/archives/*; do
     ${TAR} -C "${TEMPDIR}/sysroot" -xf "${archive}"
 done
 
+echo "===> running the self-test tool, badly"
+${TOOL_PATH} --some--cli-flag || true
+if [ "$?" -eq 0 ]; then
+    "error: should fail when given CLI args"
+    exit 1
+fi
```
pietroalbini, line 79: This is not the right place to test the change. This file is part of the validation of the artifacts produced by CI by running the self-testing tool on them, it's not to test the self-testing tool itself.

Tests for the self-testing tool are currently done with `#[test]`, and we should do the same for this test as well. It might be needed to test this as an integration test rather than an unit test though, so that it's possible to do `Command::new(env!("CARGO_BIN_EXE_ferrocene-self-test")`.

pietroalbini (changes requested):

pietroalbini (approved):
bors merge

pietroalbini (approved):
bors merge

PR #908, opened 2024-09-02
fix error codes for self-test tool

tshepang:


Dajamante (approved):
aligns with 
https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/self-test/src/error.rs#L12-L99

tshepang:
bors merge

PR #926, opened 2024-09-06
Fix automation commits not triggering CI

pietroalbini:
After migrating CI checks to GitHub Actions, it was discovered that automated PRs do not trigger CI anymore. This is because GitHub Actions refuses to start any workflow on commits pushed by a GitHub Action's `$GITHUB_TOKEN`, to prevent infinite loops. Still, we do want CI to run on those commits, so we have to resort to using our own GitHub App token.

This PR does that by creating a new `automations` environment containing a GitHub App with write access to this repository's contents, and adding a `--set-git-credentials` flag to `github_app_auth.py` to configure git to use the newly created token.

It also cleans up the existing separate GitHub apps that automations had:

* The app for `automation-backport` didn't seem to be used anywhere.
* The app for `automation-pull-subtrees` was only used to fetch private repositories for subtrees, but we don't have any private subtree anymore.
* The app for `automation-pull-upstream` was used to create new branches (due to the builtin workflow not being able to push commits containing the `.github/workflows` directory), but that has been superseded by the new app.

### DIFF
--- a/.github/workflows/automation-backport.yml
+++ b/.github/workflows/automation-backport.yml
@@ -21,8 +21,6 @@ jobs:
     name: Create PR
     runs-on: ubuntu-latest
 
-    environment: automation-backport
-
     steps:
       - name: Checkout the Ferrocene repository
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


Hoverbear (approved):
bors merge

PR #930, opened 2024-09-08
adjust text for rustfmt

tshepang:


pietroalbini (approved):
bors merge

pietroalbini:
Can't backport it now, but we should when medical is also backported.

PR #943, opened 2024-09-13
Disable parallel-compiler in CI

Veykril:


japaric (approved):

Veykril:
bors try

Veykril:
bors cancel

Veykril:
bors try

src/bootstrap/src/lib.rs
```diff
@@ -721,7 +719,7 @@ impl Build {
         if self.config.rustc_parallel && check("rustc_use_parallel_compiler") {
             features.push("rustc_use_parallel_compiler");
         }
-        if self.config.rust_randomize_layout {
+        if self.config.rust_randomize_layout && check("rustc_randomized_layouts") {
```
tshepang, line 722: what is `check` for here... it does not exist upstream

Veykril: it does, its a closure a couple lines up. It landed in the same rollup as the `rustc_randomize_layout` config being added hence the discrepancy.

Veykril:
Good failure, this is progress!

Veykril:
bors try

src/bootstrap/src/lib.rs
```diff
@@ -700,15 +700,13 @@ impl Build {
...
             .iter()
             .flat_map(|krate| &self.crates[krate].features)
             .map(std::ops::Deref::deref)
             .collect();
-        let check = |_feature: &str| -> bool {
-            //crates.is_empty() || possible_features_by_crates.contains(feature)
-            // change necessary while resolver is broken
-            true
+        let check = |feature: &str| -> bool {
+            crates.is_empty() || possible_features_by_crates.contains(feature)
```
Urhengulas, line 709: `possible_features_by_crates` is constructed from `crates`, so iiuc `possible_features_by_crates` will be empty if `crates` is empty, correct?

Also `crates` is immutable.

If above is true, then it would be enough to only do `possible_features_by_crates.contains(feature)`, because if `possible_features_by_crates` is empty it will not contain `feature`.

compiler/rustc_index_macros/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
-syn = { version = "2.0.9", features = ["full"] }
+syn = { version = "2.0.9", features = ["full", "extra-traits"] }
```
Urhengulas, line 10: Why do you need to enable the feature here and in `compiler/rustc_type_ir_macros/Cargo.toml`? Did it work with resolver 1 but not resolver 2?

compiler/rustc_transmute/src/maybe_transmutable/tests.rs
```diff
@@ -107,7 +107,7 @@ mod bool {
             #[cfg(feature = "rustc")]
             let mut set = crate::Set::default();
             #[cfg(not(feature = "rustc"))]
-            let mut set = std::collections::HashSet::new();
+            let mut set = rustc_data_structures::fx::FxHashSet::default();
```
Urhengulas, line 110: Also how is this change related?

Urhengulas (changes requested):

compiler/rustc_index_macros/Cargo.toml
Veykril: Due to how resolver 2's feature unification works, resolver 1 always unifies across the workspace, resolver 2 only across the packages being built (which is the entire problem we were having as upstream has set up the features incorrectly).

This change will be upstreamed afterwards.

compiler/rustc_transmute/src/maybe_transmutable/tests.rs
Veykril: rustc runs clippy that lints against using bare `HashSet`, upstream doesnt catch this because they dont build without the `rustc` feature.

This change will be upstreamed afterwards.

src/bootstrap/src/lib.rs
Veykril: This change brings this code back in sync with upstream.

Also those do not mean the same thing, empty `possible_features_by_crates` means no crates have any features that are allowed, so the closure should always return false in that case, where as an empty `crates` means we will build all crates so any feature is fine to be passed.

> Also crates is immutable.

?

Urhengulas (approved):
bors merge

Veykril:
bors cancel
I'll investigate the discussed alternative first

Veykril:
Upstream removed the cfg entirely

PR #945, opened 2024-09-13
Split tasks and run.sh output what they're running

Hoverbear:
After chatting with the team some folks expressed desire to improve the insight into what exactly the CI is running when `run.sh` or `split-tasks.py` is run.

This adds a bit of echo (like `set -x` but more limited) so folks can know what's going on.

### DIFF
--- a/ferrocene/ci/split-tasks.py
+++ b/ferrocene/ci/split-tasks.py
@@ -210,7 +210,9 @@ def cli():
     else:
         flags += get_flags_for_default_job(JOBS_DEFINITION, split[0])
 
-    print(" ".join(flags))
+    flag_arg = " ".join(flags)
+    print(f"Split tasks: {flag_arg}", file=sys.stderr)
+    print(flag_arg)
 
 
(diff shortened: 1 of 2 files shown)


ferrocene/ci/split-tasks.py
```diff
@@ -210,7 +210,9 @@ def cli():
     else:
         flags += get_flags_for_default_job(JOBS_DEFINITION, split[0])
 
-    print(" ".join(flags))
+    flag_arg = " ".join(flags)
+    print(f"Split tasks: {flag_arg}", file=sys.stderr)
+    print(flag_arg)
```
tshepang, line 215: did you mean to repeat displaying `flag_arg` here

Hoverbear: Yes, one will output to stdout (which is being bash expanded) and the other will get printed to console. So folks can see what is being run. THat was the request. :)

tshepang (approved):
bors merge

PR #955, opened 2024-09-18
Backport qnx tests to 24.11

Hoverbear:
We also need to backport https://github.com/ferrocene/ferrocene/pull/859 which depends on https://github.com/ferrocene/ferrocene/pull/854.

This backports 

#758 #811 #780 #787 #851 #852 #858 #860 #866

PR #859, opened 2024-08-20
fix QNX backtrace tests

japaric:
- switch libstd from crates.io backtrace-rs to our subtree
- cherry pick upstream backtrace-rs's PR648

PR #960, opened 2024-09-20
Manually add backtrace-rs subtree to 24.11

Hoverbear:
In https://github.com/ferrocene/ferrocene/pull/955 we had to skip https://github.com/ferrocene/ferrocene/pull/859 because https://github.com/ferrocene/ferrocene/pull/854 was a subtree backport.

This does that subtree backport, so that we can backport #859 .


tshepang:
we could alternatively depend on the newly-released backtrace crate, which has the fix

Hoverbear:
Think you're right. I believe @japaric is gonna do this.

PR #964, opened 2024-09-23
QNX: copy some std::error unit tests into a UI test

japaric:
to work around an issue where the `std::error` API hits an OOM condition when used from a thread

### DIFF
--- a/library/std/src/error/tests.rs
+++ b/library/std/src/error/tests.rs
@@ -114,6 +114,10 @@ fn error_with_no_sources_formats_multi_line_correctly() {
 }
 
 #[test]
+// NOTE(ignore/aarch64/nto) this and the `error_with_backtrace_outputs_correctly_with_two_sources`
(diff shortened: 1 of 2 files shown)


japaric:
bors try

library/std/src/error/tests.rs
```diff
@@ -114,6 +114,9 @@ fn error_with_no_sources_formats_multi_line_correctly() {
 }
 
 #[test]
+// NOTE(ignore/aarch64/nto) this and the next ignored test are executed as a
```
Hoverbear, line 117: I'd prefer to note both tests individually since someone might go add a test in between these two.

japaric:
bors try

Hoverbear (approved):
Noting that these are not tests which are involved in qualification, however we should note that this is a known problem.

Hoverbear:
Noting backport to 1.81

PR #1008, opened 2024-10-08
Minor fixes coming out of 24.11 review

skade:
Two patches that

a) fix minor typos that don't merit a whole review round

and

b) always confirm that we want to publish our data at cosign, increasing script resilience and signing ease


Urhengulas (approved):
bors merge

PR #1033, opened 2024-10-30
Add ARMv8-M experimental

Hoverbear:
Adds ARMv8-M experimental, might as well collect them all, eh?


Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try

jonathanpallant:
This adds support for the Armv8-M Baseline architecture, as implemented by the Cortex-M23.

1. Please can we rename the Baseline target to: "Armv8-M Baseline bare-metal".

2. Please can we also add:

   * Armv8-M Mainline with Soft Float ABI (`thumbv8m.main-none-eabi`)
   * Armv8-M Mainline with Hard Float ABI (`thumbv8m.main-none-eabihf`)
   * Armv7-M (`thumbv7m-none-eabi`)
   * Armv6-M (`thumbv6m-none-eabi`)

(The v8-M Baseline, v6-M and v7-M architectures don't allow for the existence of an FPU, so there is only one possible ABI and that's the soft-float ABI)

jonathanpallant:
Arm docs on Baseline vs Mainline: https://developer.arm.com/documentation/107656/0101/Introduction-to-Armv8-M-architecture/Baseline-and-Mainline

Hoverbear:
Heck ya!

jonathanpallant:
It would be interesting to know what this costs us in terms of build minutes, per target, per PR.

Hoverbear:
So I'd open new tickets for us to add armv7-m and armv6-m, as armv8m is the only one covered by my current milestone. While we want to add 6/7, that's for later.

Regarding build minutes: Adding this baseline target was roughly 1-2 minutes of CI time to the `targets` task, and a few seconds to `self-test`, mostly because we already had the tools installed to do it.

jonathanpallant:
OK, feel free to @ me when the mainline targets are in and I can review.

ferrocene/doc/target-names.toml
```diff
@@ -7,18 +7,21 @@
...
-x86_64-apple-darwin       = "x86-64 macOS"
-x86_64-pc-windows-msvc    = "x86-64 Windows"
-aarch64-unknown-nto-qnx710 = "Armv8-A QNX 7.1.0"
-x86_64-pc-nto-qnx710       = "x86-64 QNX 7.1.0"
+aarch64-unknown-none        = "Armv8-A bare-metal"
+thumbv7em-none-eabi         = "Armv7E-M bare-metal (soft-float)"
+thumbv7em-none-eabihf       = "Armv7E-M bare-metal (hard-float)"
+armv8r-none-eabihf          = "Armv8-R bare-metal (hard-float)"
+"thumbv8m.base-none-eabi"   = "Armv8-M Baseline bare-metal"
+"thumbv8m.main-none-eabi"   = "Armv8-M Mainline with Soft Float ABI"
```
jonathanpallant, line 15: Sorry, I've added an inconsistency here by accident.

The other targets say "(soft-float)" etc and these ones should match instead of saying "with Soft Float ABI". Same for hard float.

jonathanpallant (changes requested):

ferrocene/doc/target-names.toml
Hoverbear: Fixed!

Hoverbear:
bors merge

PR #1035, opened 2024-11-01
Repair setup action misconceptions

Hoverbear:
In our setup action we falsely assumed that composite actions could take booleans, they cannot.

Also mmakes us no longer blindly `apt upgrade` as there is zero point, we build artifacts in docker containers that are no more than a week old.


amanjeev (approved):

Hoverbear:
bors merge

Hoverbear:
That looks transient.

bors merge

PR #1049, opened 2024-11-08
New Sphinx directive: Subcommand

amanjeev:


ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/traceability_ids.py
```diff
@@ -24,11 +25,28 @@ def write_traceability_ids(app):
...
+                "option": subcommand.subcommand,
+                "link": app.builder.get_target_uri(subcommand.document) + "#" + subcommand.id(),
+            }
+        )
+
+    # Do not emit the traceability IDs if no option or subcommand was defined.
+    if not options_by_document and not subcommands_by_document:
         return
 
+    # Prepare the list of documents, adding options, subcommands separately.
```
Urhengulas, line 47: Is there a reason why you have to do it separately? Does it work combined?

Urhengulas (changes requested):
Generally looks good. There is a lot of potential for code deduplication but I understand this is a draft.

I tried running `./x doc ferrocene/doc/user-manual` after adding `..cli:subcommand:: check` to the rustc cli docs. But unfortunately the build fails and presents me some errors.

First it complains:

```text
'Subcommand' object has no attribute 'no_traceability_matrix'
```

After adding `self.no_traceability_matrix = False` to `Subcommand.__init__` it complains:

```text
Object of type ProgramStorage is not JSON serializable
```

That one puzzles me a bit because I do not know where/why it is getting json serialized. Probably somewhere sphinx internally. But also you did not change anything about the program storage, but it works on main, but not with your changes.

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/traceability_ids.py
amanjeev: Tbh I assumed that since `option`'s `document`s have this dict setup.

```python
documents.append(
            {
                "title": title.astext(),
                "link": app.builder.get_target_uri(docname),
                "options": options_by_document[docname],
                "informational": False,
            }
        )
```

I need to create another dict for `subcommand`

```python
documents.append(
            {
                "title": title.astext(),
                "link": app.builder.get_target_uri(docname),
                "subcommands": subcommands_by_document[docname],
                "informational": False,
            }
        )
```

If you think this is incorrect, please let me know and I can merge them but I do not know how to merge them? Just add another key?

Urhengulas: I do not know which option is better, or if the combined one even works. I was wondering if following could work as well:

```python
documents.append(
    {
        "title": title.astext(),
        "link": app.builder.get_target_uri(docname),
        "options": options_by_document[docname],
        "subcommands": subcommands_by_document[docname],
        "informational": False,
    }
)
```

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
```diff
@@ -47,6 +47,32 @@ def run(self):
...
+    def add_target_and_index(self, name_cls, sig, signode):
+        if PROGRAM_STORAGE not in self.env.temp_data:
+            warn("cli:subcommand outside cli:program isn't supported", self.get_location())
+            program_storage = ProgramStorage("PLACEHOLDER", False)
+        else:
+            program_storage: ProgramStorage = self.env.temp_data[PROGRAM_STORAGE]
+
+        subcommand = Subcommand(
+            self.env.docname,
+            program_storage,
```
Urhengulas, line 67: I am noting that here we pass `program_storage`, while for `Option` we pass `program_storage.name`. https://github.com/ferrocene/ferrocene/blob/a6cc042d73d53be95dab510276c6a0eb8f15c3d3/ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py#L92-L97

For both we can probably just pass the `program_storage` and extract the `name` and the `no_traceability_matrix` in the constructor.

Urhengulas (changes requested):
I tested the latest changes and now it builds for me successfully 🎉

Urhengulas (changes requested):
Actually it does not work and still gives the JSON error. Or rather it works for the rustfmt cli but not the rustc cli.

The reason is that rustfmt specifies `no_traceability_matrix` but rustc does not. If adding the subcommand directive to rustc (or adding it to rustfmt and removing the `:no_traceability_matrix:` option) it fails.

The fix seems to be to pass the program_name instead of the program_storage (see https://github.com/ferrocene/ferrocene/pull/1049/files#r1840623720).

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/traceability_ids.py
amanjeev: Personally, combining them would require checking for whether either of `options` or `sbucommands` is `None`. That subscript access can and will fail because I do not know if they are not mutually exclusive.

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
amanjeev: I actually renamed "program" to "program_name" for clarity.

amanjeev:
@Urhengulas I have renamed `program` to `program_name` to make it explicit what is being passed.

Hoverbear:
image

```rst
.. cli:program:: criticalup

    .. cli:subcommand:: install [OPTIONS] <COMMAND>

            Test install docs

            .. cli:option:: --project <PROJECT>

               This is a doc of one of the install options

            .. cli:option:: --offline

               This is a doc of the second of the install options

    .. cli:subcommand:: verify [OPTIONS] <COMMAND>

            Test verify docs

            .. cli:option:: --project <PROJECT>

               This is a doc of one of the install options

            .. cli:option:: --reinstall

               This is a doc of the second of the install options
```

It looks like this handles adding options to subcommands as well, nice.

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
```diff
@@ -105,6 +131,28 @@ def id(self):
...
+        subcommand = (
+            ARGUMENT_PLACEHOLDER_RE.sub("_", self.subcommand)
+            .replace("=", "")
+            .replace("-", "_")
+            .replace(" ", "_")
+            .strip("_")
+        )
+        subcommand = MULTIPLE_UNDERSCORES_RE.sub("_", subcommand)
+
+        return f"um_{self.program}_{subcommand}"
```
Hoverbear, line 155: So I am a bit curious how this will work (if at all) when we document programs with subcommands with options.

Looking at the "Option" class, will `criticalup install --project .` and `criticalup verify --project .` have the same `um_criticalup_project` id? Will this be a problem? 🤔

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/traceability_ids.py
Hoverbear: What about subcommands with options?

ferrocene/doc/sphinx-shared-resources/exts/ferrocene_domain_cli/domain.py
amanjeev: Oooh This is a great observation. I think for subcommand I can update the ID to 

```
**f"um_{self.program}_subcommand_{subcommand}"**
```

But I don't know how it will interact with the rest of the codebase. :grimacing: But let me try.

Urhengulas (commented):
Is me review currently desired here?

Hoverbear:
Nominating for closure.

amanjeev:
We decided to not do this. This is not needed for Ferrocene and can be revisited later if needed.

PR #1066, opened 2024-11-13
Add more uv docs

Hoverbear:
Takes a shot at clarifying and improving our `uv` docs.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
+++ b/ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
@@ -120,7 +120,7 @@ Or, to run the ``split-tasks.py`` script for debugging:
    
    uv run ferrocene/ci/split-tasks.py test
 
-In CI workflows you should always use ``uv run`` for calling Python. In the case of ``make.py`` it
+In CI workflows you should always use ``uv run`` for calling Python. In the case of ``x.py`` it
 is not required as the script has been carefully written to work with most Python versions.
 
 
--- a/ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
+++ b/ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
@@ -70,37 +70,39 @@ Using the Python virtual environment
    You do not need to install Python, or modify your system Python.
 
 
-Ferrocene uses ``uv`` to manage Python and dependencies. There are a number of
-ways to install `listed in the repository <https://github.com/astral-sh/uv>`_,
-but we recommend the ``curl`` (Linux, MacOS) or ``powershell`` (Windows)
-methods:
+Ferrocene uses ``uv`` to manage Python and dependencies. ``uv`` is similar to ``rustup``, ``rvm``,
+``nvm``, or ``rbenv``. It provides a way for a project (Ferrocene) to depend on a specific version
+of Python and dependencies, creating a consistent development environment across any operating
+system or distribution it supports.
+
+This is particularly useful on platforms like :ref:`aarch64-apple-darwin` or
+:ref:`x86_64-pc-windows-msvc`, which do not always include modern versions of Python.
+
+You can install it using the same script our CI runners use:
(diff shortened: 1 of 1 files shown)


Urhengulas (approved):

Urhengulas (changes requested):
Actually I am unsure if it works for. I did run `setup-venv.sh` which gives:

```sh
$ ./ferrocene/ci/scripts/setup-venv.sh 
+ command -v uv
+ curl -LsSf https://astral.sh/uv/install.sh
+ sh
downloading uv 0.5.1 x86_64-unknown-linux-gnu
no checksums to verify
installing to /home/urhengulas/.local/bin
  uv
  uvx
everything's installed!
+ source /home/urhengulas/.cargo/env
++ case ":${PATH}:" in
+ echo 'source /home/urhengulas/.cargo/env'
./ferrocene/ci/scripts/setup-venv.sh: line 15: BASH_ENV: unbound variable
```
The last line sounds like an error? Assumingly from line 15.

When I then run `uv run which python` I get my system python which is not what happened in your demo today.

```console
$ which python
/usr/bin/python
$ uv run which python
/usr/bin/python
```

When I then run `uv venv ~/.venv` explcitly it actually creates the venv and then which gives the right python:

```console
$ uv venv ~/.venv
Using CPython 3.12.7 interpreter at: /usr/bin/python
Creating virtual environment at: /home/urhengulas/.venv
Activate with: source /home/urhengulas/.venv/bin/activate
$ uv run which python
/home/urhengulas/.venv/bin/python
```

Hoverbear:
Great catch, thanks. :) I'll look into this.

Hoverbear:
@Urhengulas Can I ask you to use the latest main? I think yours was out of date and we only recently made this portable outside CI.

Dajamante:
Thanks for clarifying!
The point that I still find a bit confusing. Do we still need to update our `config. tool` to match CIs?
I am asking because the indentation for this instruction does not make it clear.

ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
```diff
@@ -112,6 +114,15 @@ For example, to run our licensing checks you can run:
    
    uv run reuse --include-submodules lint
 
+Or, to run the ``split-tasks.py`` script for debugging:
+
+.. code-block:: bash
+   
+   uv run ferrocene/ci/split-tasks.py test
+
+In CI workflows you should always use ``uv run`` for calling Python. In the case of ``make.py`` it
```
Hoverbear, line 123:
```suggestion
In CI workflows you should always use ``uv run`` for calling Python. In the case of ``x.py`` it
```

Hoverbear:
> Do we still need to update our config.toml to match CI's when working with uv, aka running CI=true FERROCENE_HOST=aarch64-unknown-linux-gnu ./ferrocene/ci/configure.sh?
I am asking because the indentation for this instruction does not make it clear.

No, I'm not really sure where this comes from?

Are you referring to this separate heading after this one? 

image


They are separate ideas, and unrelated.

Urhengulas (approved):
With the latest main it works!

Dajamante:
> No, I'm not really sure where this comes from?
> 
> Are you referring to this separate heading after this one?

Yes, I was referring to the separate idea.
Because I never use `uv`, just docker, where I needed to change the `config.toml` to match CI, I didn't know if it was the case for when we use `uv`.

ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
Dajamante: Doesn't `make` do it as well, by calling its own file of requirements?

Hoverbear:
@Dajamante you do not need to change your config to match CI unless you are trying to reproduce the CI. (Even then, you likely will not use the same config exactly since you likely don't want to have to do AWS auth to sign packages etc)

PR #1067, opened 2024-11-13
Repair `download-ci-llvm` via including overlay in `rust-dev` component

Hoverbear:
The `download-ci-llvm` config option is currently not working

```toml
# config.toml
[llvm]
download-ci-llvm = true
```

If a user tries to use it they'll get something like

```
download: s3://ferrocene-ci-artifacts/ferrocene/dist/fb8e927c45ed5d7861fb9ff4b9e55ece779a4382/rust-dev-aarch64-unknown-linux-gnu-fb8e927c4.tar.xz to build/tmp/rust-dev-aarch64-unknown-linux-gnu-fb8e927c4.tar.xz
extracting /home/ana.linux/git/ferrocene/ferrocene/build/cache/llvm-fb8e927c45ed5d7861fb9ff4b9e55ece779a4382-false/rust-dev-aarch64-unknown-linux-gnu-fb8e927c4.tar.xz to /home/ana.linux/git/ferrocene/ferrocene/build/aarch64-unknown-linux-gnu/ci-llvm
thread 'main' panicked at src/core/config/config.rs:1336:13:
fs::read_to_string(file) failed with No such file or directory (os error 2) ("config file /home/ana.linux/git/ferrocene/ferrocene/build/aarch64-unknown-linux-gnu/ci-llvm/builder-config not found")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Build completed unsuccessfully in 0:00:37
```

Which, is a huge bummer!

It turns out our `rust-dev` component doesn't have the `builder-config` present, so the cache doesn't work. This is built as part of our `dist` jobs.

I consulted some folks I know and it seems that sometimes other components can cause problems, or the fact that this is not depended upon directly may be the cause.

This PR is so we can `bors try` and see if we get a working cache from CI.

### DIFF
--- a/ferrocene/ci/configure.sh
+++ b/ferrocene/ci/configure.sh
@@ -275,7 +275,7 @@ add --enable-extended
 #
 # NOTE: If you add a new tool here, make sure to also change
 # `ferrocene/packages.toml` to include it in new releases.
-add --tools=rustdoc,cargo,llvm-tools,rustfmt,rust-analyzer,clippy
+add --tools=rust-dev,rustdoc,cargo,llvm-tools,rustfmt,rust-analyzer,clippy
 
 # Build and enable the profiler runtime.
 #
--- a/ferrocene/ci/configure.sh
+++ b/ferrocene/ci/configure.sh
@@ -275,7 +275,7 @@ add --enable-extended
 #
 # NOTE: If you add a new tool here, make sure to also change
 # `ferrocene/packages.toml` to include it in new releases.
-add --tools=rust-dev,rustdoc,cargo,llvm-tools,rustfmt,rust-analyzer,clippy
+add --tools=rustdoc,cargo,llvm-tools,rustfmt,rust-analyzer,clippy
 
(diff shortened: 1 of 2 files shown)


Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
I manually overrode https://github.com/ferrocene/ferrocene/blob/9923cb84166510e7312107de017462afa1770290/src/bootstrap/src/core/build_steps/llvm.rs#L160 to be:

```rust
let llvm_sha = String::from("f693c52df30c1dbb8979c116be1a5f8973795448");
```

Now it seems to work.

pvdrz (approved):
This looks good to me. Did you have the chance to test it this with the hash override already?

Hoverbear:
Yes I overrode the hash and watched my local dev copy download the right LLVM and use it instead of building. It was quite magical.

Hoverbear:
bors merge

PR #1070, opened 2024-11-13
Allow choosing which branches go in an upstream pull

pvdrz:
This adds a check :ballot_box_with_check:  selector to the upstream pull action so we can choose which branches will be pulled (`master`, `beta`, `stable`). 

The default behavior still is that the three of them are pulled so the only added value here is to save CI time and remove clutter if you only want manually trigger an upstream pull from a particular branch.


Hoverbear (approved):

tshepang:
bors merge

tshepang (approved):

PR #1073, opened 2024-11-14
Add editor-friendly syntax for conflicts

pvdrz:
This PR adds the `--fetch` flag to the conflicts script so they're printed in the `path:line` format ccommonly used by stacktraces, compilers, etc. The main motivation here is that you can do
```bash
your-fave-editor $(ferrocene/ci/scripts/detect-conflict-markers.py --fetch)
```
and open all the files with conflicts at the line where the conflict starts

### DIFF
--- a/ferrocene/ci/scripts/detect-conflict-markers.py
+++ b/ferrocene/ci/scripts/detect-conflict-markers.py
@@ -71,11 +71,11 @@ def conflict_markers_in_file(file):
             yield CustomDeleteMarker(file=file)
 
 
-def main(directory):
+def main(directory, fetch):
     found_conflicts = False
     for file in files_with_possible_conflict_markers(directory):
         for conflict in conflict_markers_in_file(file):
-            print(conflict.repr())
+            print(conflict.repr(fetch))
             found_conflicts = True
 
     if found_conflicts:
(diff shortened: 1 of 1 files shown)


tshepang:
am thinking of 2 alternatives:
- we should instead use your `--fetch` output as the only output, because the more explainy output does not feel very useful
- pour the more explainy output to stderr, while keeping your `--fetch` output on stdout, so that the right thing still happened when you use fave-editor

pvdrz:
I don't have the big picture to know if your suggestion of modifying the default behavior of this script is appropriate or not.. Adding an extra flag without touching the default behavior is less controversial imho

tshepang:
what I am implying is that we don't need an extra flag to achieve your goal

pvdrz:
My goal is to improve the workflow for some of us without impairing the workflow of those who use this script as it is

tshepang:
that would mean going for option 2 above, which gives you want you want without a need for an extra flag

Veykril (approved):

tshepang (changes requested):
https://github.com/ferrocene/ferrocene/pull/1073#issuecomment-2475648035

PR #1088, opened 2024-11-18
Trace ferrocene-self-test requirements to tests

Urhengulas:
This PR finds tests for each of the ferrocene-self-test requirements. The requirements can be found at https://public-docs.ferrocene.dev/main/requirements/self-test.html.

For reviewers: I would like you to check if the requirement fits to the test case and also if there are other tests that would fit that requirement, that I missed. A single test can be annotated with multiple requirements. If there is no test for a requirement, then this test has to be written.

Regarding the requirement hierachy, a requirement is considered traced if all the subrequirements are traced for all the variations (variations being e.g. the different binary tools).

This PR also makes the binaries in `ferrocene/tools` their own workspace. This is because `rust-analyzer lsif` will always analyze the whole workspace and that takes ages in the current setup. I did not yet look into what is being broken by removing the tools from the main workspace, but there certainly will be something. The best solution would be for rust-analyzer to be able to only analyze a specific package (and its dependencies), but for the short term this solution works for me.


tshepang:
this looks good... will add my thoughts to some of the questions asked

ferrocene/tools/self-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+- [ ] REQ_8TNOYG8: Check if a sample program that rustc produced can be executed.
+- [ ] REQ_B07M5S2: Check if the output of the program is as expected.
+
+
+# TODO
+
+- [ ] Make NOTEs into requirements, e.g. "The following checks apply to these Host platforms: x86-64 Linux (glibc), Armv8-A Linux (glibc)"
+- [ ] "Test" General requirements with a test that just contains a comment that explains why the requirement is covered?
+- [ ] Only use "target" or "platform", but do not use it interchangeably
```
tshepang, line 72: I think we should use "target"

ferrocene/tools/self-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+- [ ] Are we allowed to have tests that are not traced to any requirement?
```
tshepang, line 73: it would be good to identify them, and check why such are not in requirements

ferrocene/tools/self-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+- [ ] What about "negative tests", tests that test the opposite of a requirement does not work?
```
tshepang, line 74: these should be annotated with the REQ they negate

ferrocene/tools/self-test/README.md
```diff
@@ -0,0 +1,76 @@
...
+- [ ] Permission tests for windows
+- [ ] Should the requirements macro go before or after the test macro? Before or after the cfg(windows) macro?
```
tshepang, line 76: I would say after, since the annotation is secondary (so as to give more visibility to the more standard annotation, `#[test]`

PR #1090, opened 2024-11-18
Fix setup-darwin.sh pointing at a broken gcc-arm-embedded

Hoverbear:
Works around the `gcc-arm-embedded` package being a 404 on Arm's mirrors.

Since `brew install --cask` won't work with old versions, we need to manually copy the old formulae in and use it.


Hoverbear:
bors try

PR #1091, opened 2024-11-19
adding new field in the manual

Dajamante:
This is a bit over explanatory...
The field exist only in the manual as it is generated by the KP tool.


Hoverbear (approved):
bors merge

PR #1093, opened 2024-11-19
testing to reduce used space

Dajamante:


Hoverbear:
I was looking around and this might have some ideas for us: https://github.com/easimon/maximize-build-space/blob/fc881a613ad2a34aca9c9624518214ebc21dfc0c/action.yml#L45-L60

(Please do not try to use the action, it's not on our allowlist)

tshepang:
served its purpose

PR #1096, opened 2024-11-19
LLVM is not currently used in `commit-checks.yml`

tshepang:
Also, it was broken... it did not get installed.


Hoverbear (changes requested):
Please provide more context so I can better understand the change and what you're trying to accomplish. 

Please make the title something more precise.

tshepang:
closing this in that case

PR #1098, opened 2024-11-20
Correct `setup-ubuntu.sh` not recieving env variables

tshepang:


Hoverbear (approved):
Yeah this makes sense... 😬 Sad we didn't notice that before.

Hoverbear:
bors merge

PR #1099, opened 2024-11-20
Remove some packages the GHA runner already has installed

tshepang:


Hoverbear:
If we try to install, and it's already installed, it just skips it right? No harm? I wonder if it's better to keep them so folks can use the scripts, or if the runner images change (as they're known to)?

tshepang:
this way there are less logs to look at, because the more lines there are, the more easy it is to miss things

also, I just prefer that we only do what is needed, instead of do things "in case"... feels cleaner in my head

Hoverbear (approved):
bors merge

PR #1112, opened 2024-11-21
PR rotation uses a bigger runner

Hoverbear:
We were having problems with checkout on the automation, this solves it by literally just throwing more power at it. We'll likely want to take a slightly smaller runner later.

Note: This runner only works for certain branches/workflows, you can't just use it willy nilly.

### DIFF
--- a/.github/workflows/automation-pull-upstream.yml
+++ b/.github/workflows/automation-pull-upstream.yml
@@ -28,7 +28,7 @@ permissions:
 jobs:
   run:
     name: ${{ matrix.branch }}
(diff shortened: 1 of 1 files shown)


tshepang:
[we had success with standard runner](https://github.com/ferrocene/ferrocene/actions/runs/11956397464]

Hoverbear:
Seems we're ok without doing this. Closing.

> is that runner name set in settings, accessible only by admins

It's in the organization settings under "Actions" and "Runner groups". It's only configurable by admins I think.

PR #1115, opened 2024-11-21
use Python syntax known by Ubuntu 22.04

tshepang:
Backport CI runner has Python 3.10, but this syntax was only introduced in Python 3.12. This resulted in backports failing.


tshepang:
alternative to this is just using Ubuntu 24.04 runner, but I find the newer syntax less easy to read

pvdrz (approved):

pvdrz:
bors merge

PR #1116, opened 2024-11-21
bump ci timeout to 4h

tshepang:
Had a sad case on an anticipated CI run succeed, but having it take 3h14m resulted in bors fail.


skade (approved):
This should be reduced at some point, but for now, it's okay.

tshepang:
bors merge

PR #1122, opened 2024-11-22
submodules take much too long to clone

tshepang:
this is an experiment


tshepang:
this resulted in https://github.com/ferrocene/ferrocene/pull/1124, but since it does not pull in updated submodules from upstream, an not sure it would work... to be revisited in future perhaps

PR #1130, opened 2024-11-25
Set version to 24.08.1 and channel to beta

Hoverbear:
Seems this was forgotten previously. :)

This is the last step of https://public-docs.ferrocene.dev/main/qualification/internal-procedures/release/stable.html#publishing-a-stable-release


amanjeev (approved):

Hoverbear:
bors merge

tshepang (approved):
bors merge

tshepang:
failing because Ubuntu 23.10 is deprecated

jyn514:
i think we should just merge this without running the CI checks. this doesn't change compiler behavior and this is our next-to-oldest release so i don't expect any of our customers to actually be using it, especially since it doesn't have a certified libcore.

tshepang:
I would rather we deprecate the release than we accept a change without running CI. That is, this is still a supported release.

PR #1142, opened 2024-11-29
Set `PreferredToolArchitecture=x64` on windows builds

pvdrz:


Xylakant:
Not sure this is the right place to set it. This might work under the assumption that the build happens in the same shell, but depending on how the build is split into multiple stages, the build might happen in a fresh shell, loosing the env var.

Hoverbear:
Might be setting it in here is the right place?

https://github.com/ferrocene/ferrocene/blob/908fc6606f8f17f5ac0b5217d99825160543776f/.circleci/workflows.yml#L637-L641

pvdrz:
bors try

pvdrz:
no significant difference still

PR #1063, opened 2024-11-13
Update "Upstream pulls" in "Internal Procedures"

Dajamante:
This PR updates the existing procedures with current procedures.

To test:
 `x doc ferrocene/doc/internal-procedures/ --serve` 

In review:
Please check if this makes sense, or if there is something recurrent that I forgot.

Questions to reviewers:
What more to say on ignored tests? When is it good practice to ignore a test?
Do we need to describe `run completions`?
Is it `x fmt --check` for the licensing?
Shall I add structured handover?

Added @Hoverbear points:
- monitor the cricleci job and cancel eagerly on first failure.
- always try running locally with x check or x build

PR #1145, opened 2024-12-02
Remove unused figures from internal procedures

Veykril:
Added in https://github.com/ferrocene/ferrocene/pull/1063 but they are not referenced and additionally collide with other files on case insensitive filesystems.


tshepang (approved):

Veykril:
bors merge

PR #1164, opened 2024-12-13
Prevent compiler-arguments tests from polluting the repo

Veykril:


tests/ui/ferrocene/compiler-arguments/o/o_no-delim.rs
Veykril, line 1: Removed as this is a duplicate of `o_o.rs`, exact same files

tshepang: I suppose you referring to `o.rs`, and even then, it's not quiet the same... one has space in the cli (which I suppose it what is referred to as delim)

Veykril: I did mean `o.rs` right, but neither have a space in them, both just pass `-ofoo`

tshepang: oh, indeed...

tshepang (approved):
bors merge

PR #1171, opened 2024-12-20
exit was already assigned a bool

tshepang:
Also, it is advised to use sys.exit instead, which has same behavior.


amanjeev (approved):

tshepang:
bors merge

PR #1173, opened 2024-12-20
request is used by our scripts, so add it to requirements.txt

tshepang:


requirements.txt
```diff
@@ -43,6 +51,8 @@ six==1.16.0
 tomlkit==0.13.0
     # via reuse
 urllib3==2.2.1
```
amanjeev, line 53: do you need urllib3 if you have requests already?

tshepang: I think urllib3 is lower-level than requests, and it looks like that latter depends on it

amanjeev: Yes and I assumed `requests` would pull `urllib3` anyway. I am thinking on the lines of that if we use `urllib3` elsewhere we can just change that code to use `requests`.

tshepang: oh, I think this file is equivalent to Cargo.lock... it shows transitive deps in addition to direct ones

Hoverbear (approved):

Hoverbear:
bors merge

This is used in, for example, https://github.com/ferrocene/ferrocene/blob/7461f0f7e961f308eef99779a94b0636330e9947/ferrocene/tools/common/automated_prs.py#L13

PR #1182, opened 2025-01-03
std: pin libc version

japaric:
to ensure our subtree gets used


pietroalbini (changes requested):
This approach will break as soon as we upgrade our subtree. I'm working on a more resilient patch right now.

PR #1184, opened 2025-01-03
Reject PRs with unused Cargo patches

pietroalbini:
This PR adds a script to run in CI to reject PRs containing unused Cargo patches. This is likely due to a subtree falling behind.

Note that this PR will most likely fail CI for now, as the libc subtree is behind.

### DIFF
--- a/ferrocene/ci/scripts/detect-unused-cargo-patches.py
+++ b/ferrocene/ci/scripts/detect-unused-cargo-patches.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+import subprocess
+import tomllib
+from collections import defaultdict
+
+
+def find_files(name):
+    cmd = subprocess.run(
+        ["git", "ls-tree", "HEAD", "--name-only", "-r", "-z"],
+        stdout=subprocess.PIPE,
+        check=True,
+        text=True,
+    )
+
+    # The -z flag changes the output to separate file names with a null byte, preventing files with
+    # spaces or newlines from messing things up.
+    for file in cmd.stdout.split("\0"):
+        if file == name or file.endswith(f"/{name}"):
+            yield file
+
+
+def find_unused_patches(file):
+    with open(file, "rb") as f:
+        try:
+            toml = tomllib.load(f)
+        # Probably due to conflict markers:
+        except tomllib.TOMLDecodeError:
+            print(f"note: skipping {file} as it is not valid TOML")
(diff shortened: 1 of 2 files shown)


ferrocene/ci/scripts/detect-unused-cargo-patches.py
```diff
@@ -0,0 +1,64 @@
...
+    for file in cmd.stdout.split("\0"):
+        if file == name or file.endswith(f"/{name}"):
+            yield file
+
+
+def find_unused_patches(file):
+    with open(file, "rb") as f:
+        try:
+            toml = tomllib.load(f)
+        # Probably due to conflict markers:
```
Hoverbear, line 27: If conflict markers are present, this probably shouldn't pass, right?

ferrocene/ci/scripts/detect-unused-cargo-patches.py
```diff
@@ -0,0 +1,64 @@
...
+        except tomllib.TOMLDecodeError:
+            print(f"note: skipping {file} as it is not valid TOML")
+            return
+
+    try:
+        unused_patches = toml["patch"]["unused"]
+    except KeyError:
+        return
+
+    package_versions = defaultdict(list)
```
Hoverbear, line 37: I am somewhat alarmed by this line but it does seem to be valid python, somehow.

Hoverbear (approved):
Seems to work for me

ferrocene/ci/scripts/detect-unused-cargo-patches.py
pietroalbini: When you try to access a missing element, `defaultdict` creates the new element with the function provided to it. In this case the function is `list`, creating an empty list.

It's equivalent to this code, should we switch to it?

```suggestion
    package_versions = defaultdict(lambda: [])
```

ferrocene/ci/scripts/detect-unused-cargo-patches.py
pietroalbini: Conflict markers will already be caught by the step right before, so there is no risk of CI letting files with conflict markers go through. I chose not to make this an error if someone has other files called Cargo.lock that are not valid TOML for whatever reason. I can switch it to an error if you'd prefer.

ferrocene/ci/scripts/detect-unused-cargo-patches.py
Hoverbear: I think it's fine, Python just unsettles me sometimes. 😅

pietroalbini:
Rebased on top of the libc update.

tshepang (approved):
bors merge

PR #1193, opened 2025-01-08
Use `uv` to run scripts in `ferrocene/ci/scripts`

pietroalbini:
More and more of our scripts in `ferrocene/ci/scripts` are written in Python, but dependency management was fairly chaotic. Some of the scripts imported each other (like `calculate_parameters.py` importing `llvm_cache.py` importing `cache.py`), some depended on Python libraries installed with `apt-get` on the CI runners, and some required the `requirements.txt` defined at the root of the repository.

This PR creates an [uv project](https://docs.astral.sh/uv/guides/projects/) at `ferrocene/ci/scripts/utils` containing shared code and utilities the scripts reference (instead of importing each other), and moves every script in the directory to [uv scripts](https://docs.astral.sh/uv/guides/scripts/).

uv scripts are especially nice because they declare the required Python version and dependencies at the top of the file, and `uv` takes care of installing those automatically before running the script. This removes the need to setup a virtual environment, and thanks to the shebang it also removes the requirement to run the scripts with `uv run`!

Fixes https://app.clickup.com/t/8697bzp3u


pietroalbini:
Also added a commit that stops calling those scripts with `uv run`, as the shebang already handles that.

The state I want to get to is that unless you are *writing* scripts you shouldn't have to care about virtualenvs or using `uv run` at all. It should just work (thanks to the shebang pointing to uv instead of Python). Ideally at the end of this we'd have a CI check rejecting any file with a Python shebang in `ferrocene`, `.circleci` and `.github`.

pietroalbini:
An alternative implementation for this would be to create a single project for the whole `ferrocene/ci/scripts` and declare the dependencies centrally. This would have the advantage of generating the lockfile, but would have the downside of not being able to specify each script's individual dependencies. I personally prefer this approach, but let me know if you'd prefer otherwise.

ferrocene/ci/scripts/cache.py
```diff
@@ -1,90 +1,21 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S uv run
```
Hoverbear, line 1: A particular HN commentator suggests shebang approach fails if you are ever offline (such as in a sandbox), perhaps this is a concern for us? https://news.ycombinator.com/item?id=42201103

pietroalbini: At least for these scripts no.

Hoverbear (commented):
Looks good overall, locally seems to work well. Once we fix the merge conflict I'd love to see the bors result :)

pietroalbini:
Rebased.

tshepang (approved):
bors merge

Hoverbear:
Seems the CircleCI runner agent crashed? Lets try again.

bors merge

tshepang:
bors retry

another transient issue, this time a "Couldn't connect to server"

PR #1199, opened 2025-01-10
Convert documentation to use uv

pietroalbini:
This PR converts our documentation generation to use `uv` to manage its virtual environment, and packages all of the shared resources into a proper Python package. It also adds vendoring of the Python dependencies.

On the bootstrap side, this PR requires the presence of `uv` to build the documentation. It does not let uv manage the Python installation (as customers building from source wouldn't like Python being downloaded behind their backs), and instead forces uv to use the same Python used by bootstrap. This means customers building from source just have to install `uv` into their build environment, not having to switch how they obtain Python installations.

Due to how fast uv is, the virtual environment is synchronized every time a build happens.

While I was at it, I implemented vendoring of the Python dependencies: when building the source tarball we prime the uv cache, and then include that cache into the built tarball. Then, if vendoring is enabled uv will be executed in offline move, providing the vendored cache to it.

Along with this, I also implemented a single Python package containing all of our shared resources and our theme. In a future PR I'll change the mirroring code to take this into account, and update the FLS and the problems database to use the new packaged version of the shared resources.

https://app.clickup.com/t/8697bzpyx


pietroalbini:
Pushed a temporary commit that makes the spec build again. The problem is the spec was hardcoding the path to its own copy of the shared resources, instead of using the ones packaged by uv. This will go away once the spec switches to uv.

src/bootstrap/src/core/sanity.rs
```diff
@@ -199,6 +199,13 @@ than building it.
         .map(|p| cmd_finder.must_have(p))
         .or_else(|| cmd_finder.maybe_have("reuse"));
 
+    build.config.uv = build
```
Hoverbear, line 202: Should the python also be configured to use `uv`'s python?

Hoverbear (commented):
Overall looks good, I think that `uv` needs to be added to the list of required software in https://public-docs.ferrocene.dev/main/qualification/internal-procedures/setup-local-env.html. Could that be part of this PR?

Hoverbear:
I get some warnings that may be worth reviewing/resolving:

```
ana:~/git/ferrocene/ferrocene$ ./x.py doc ferrocene/doc/safety-manual  --open
Building bootstrap
    Finished `dev` profile [unoptimized] target(s) in 0.03s
WARNING: The `change-id` is missing in the `config.toml`. This means that you will not be able to track the major changes made to the bootstrap configurations.
NOTE: to silence this warning, add `change-id = 134809` at the top of `config.toml`
Updating the Sphinx virtual environment
warning: `VIRTUAL_ENV=.venv` does not match the project environment path `build/x86_64-unknown-linux-gnu/ferrocene/sphinx-venv` and will be ignored
warning: Missing version constraint (e.g., a lower bound) for `sphinx`
warning: Missing version constraint (e.g., a lower bound) for `sphinx-needs`
warning: Missing version constraint (e.g., a lower bound) for `tomli`
Resolved 51 packages in 1ms
Audited 49 packages in 0.07ms
Gathering references of ferrocene/doc/specification
```

pietroalbini:
Added the documentation and fixed most of the warnings. Note that the warning about the virtualenv is something I cannot fix, as it basically says that it's ignoring the external virtual environment you configured (which is the expected behavior).

At some point with my uv migration I'd like to completely remove the external virtualenv.

Hoverbear (approved):
bors merge

pietroalbini:
bors retry

Spurious failure.

PR #1200, opened 2025-01-10
Setup `uv` in automations

tshepang:


Hoverbear:
You beat me to it! Can you add this to the https://github.com/ferrocene/ferrocene/blob/main/.github/workflows/automation-backport.yml as well?

PR #1211, opened 2025-01-15
Show pull upstream warnings in the generated PRs

pietroalbini:
Right now the pull upstream automation displays some warnings in its output when it does something requiring manual action from the members of the PR rotation. This is far from ideal: if the person reviewing the PR wants to look at them, they'd have to find the GitHub Actions workflow that created the PR and sift through tons of log messages to find them.

This PR improves the situation by adding a new function to `pull.sh`: `automation_warning`. When the function is called, the warning will be displayed with emphasis in the logs and it will be recorded in a separate file. The automation will then include all of those warnings at the top of the description of the created PR, highlighting them for the reviewers.

Right now I only converted to this system the warnings that require explicit action (this is a partial pull, there are committed conflict markers, and the completions have to be regenerated) to keep the noise low. In the future it'll be up to the PR rotation team to figure out what warnings could be useful for them, and extend the pull script to include them.

https://app.clickup.com/t/8697gmju8

### DIFF
--- a/ferrocene/tools/pull-upstream/automation.py
+++ b/ferrocene/tools/pull-upstream/automation.py
@@ -15,6 +15,7 @@ from dataclasses import dataclass
 import os
 import sys
 import generate_pr_body
+import tempfile
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "common"))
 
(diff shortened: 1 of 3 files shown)


Dajamante (approved):
bors merge

PR #1222, opened 2025-01-17
test ARMv7E-M targets

japaric:


japaric:
bors try

japaric:
bors try-

japaric:
bors try

Hoverbear (approved):
This generally looks good to me (with the understanding that there are some tests and further fixes coming).

ferrocene/ignored-tests.toml
pietroalbini, line 1: There is duplication in the entries between aarch64 and thumbv7m. Could we deduplicate the blocks as much as possible (having a shared block, and then target-specific blocks if some tests work on some targets)?

library/core/build.rs
```diff
@@ -0,0 +1,16 @@
...
+
+    if !target.contains("-ferrocenecoretest") {
+        return;
+    }
+
+    match target.as_str() {
+        "aarch64-unknown-ferrocenecoretest" => {}
+        "thumbv7em-ferrocenecoretest-eabi" | "thumbv7em-ferrocenecoretest-eabihf" => {
+            println!("cargo::rustc-cfg=ferrocenecoretest")
+        }
```
pietroalbini, line 13: The fact that `cfg(ferrocenecoretest)` is not set for all coretest targets is really confusing. We should either set it for all of our targets (if appropriate) or find a better name for it.

Also, instead of duplicating this in all build scripts, we could add it in a single place [inside of bootstrap](https://github.com/ferrocene/ferrocene/blob/44d7fa90d6868978259eb4d0ff5f36525d51207a/src/bootstrap/src/core/build_steps/compile.rs#L451).

pietroalbini (changes requested):

japaric:
bors try

pietroalbini (approved):
bors merge

PR #1225, opened 2025-01-17
Rename experimental targets to untested targets

Hoverbear:
We discussed if "Untested" is a better name than "Experimental", this PR does the needful in case we commit.

### DIFF
--- a/ferrocene/doc/user-manual/src/targets/index.rst
+++ b/ferrocene/doc/user-manual/src/targets/index.rst
@@ -94,9 +94,11 @@ available.
 For any of the following reasons, the target is not qualified:
 
 * The target is deemed unlikely to be used in a safety critical context.
-* The target is in the process of qualification, but is not completed yet.
+* The target is in the process of qualification, but the work has not
+  been completed yet.
 
-Quality managed targets are not qualified, but can usually be qualified on request.
+Quality managed targets are not qualified, but can usually become qualified on
+request.
 
 .. list-table::
    :header-rows: 1
@@ -114,12 +116,22 @@ Quality managed targets are not qualified, but can usually be qualified on reque
      - \-
 
 Untested targets
---------------------
+----------------
 
 Untested targets cannot be used in safety-critical contexts, and there is
 no guarantee that the Ferrocene test suite is successfully executed on the
 target. They are provided as a preview, with limited support available. They
 should not be used in production.
+
+For any of the following reasons, the target is not quality managed or qualified:
+
+* The Ferrocene test suite may not successfully execute on the target.
+* Some or all of the Ferrocene test suite does not yet execute on the target.
+* An appropriate virtualization target for testing has not been identified.
+* Improved support for the target has not been requested.
+
+Untested managed targets are not quality managed or qualified, but can usually
+become quality managed or qualified on request.
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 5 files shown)


plaindocs:
My understanding was that we were **adding** untested, not renaming. Did I get that wrong?

Hoverbear:
Reading the recent discussions, it seems we might hold some targets back under "Experimental" still. I'll update.

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -5,7 +5,7 @@ Compilation targets overview
 ============================
 
 Ferrocene has support for multiple compilation targets and host platforms.
-Targets are categorized as either "supported" or "experimental" depending on
+Targets are categorized as either "supported" or "untested" depending on
```
pietroalbini, line 8: This is missing "experimental".

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -185,6 +196,27 @@ should not be used in production.
...
+provided as-is and the targets be removed from future versions without notice.
+
+For any of the following reasons, the target is experimental:
+
+* The target is deprecated and no longer produced.
+* The target has an unstable API.
+
+If your project needs support for one of these targets, consider reaching out the Ferrocene
+support team to discuss custom support contracts.
+
```
pietroalbini, line 219: This is missing the table header (`.. list-table::`, etc).

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -185,6 +196,27 @@ should not be used in production.
...
+     - Host platform
+     - Full
+     - \-
+
+Experimental Targets
+--------------------
+
+Experimental targets cannot be used in safety-critical contexts, the Ferrocene test suite
+was not executed on the target. No support is available for these targets, artifacts are
+provided as-is and the targets be removed from future versions without notice.
```
pietroalbini, line 210:
```suggestion
provided as-is and the targets be could removed from future versions without notice.
```

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -185,6 +196,27 @@ should not be used in production.
...
+
+For any of the following reasons, the target is experimental:
+
+* The target is deprecated and no longer produced.
+* The target has an unstable API.
```
pietroalbini, line 215: I might also add "the target is a proof of concept"?

ferrocene/doc/user-manual/src/targets/index.rst
Hoverbear: Also "qualified" and "quality managed", oops!

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -185,6 +196,37 @@ should not be used in production.
...
+     - Host platform
+     - Full
+     - \-
+
+Experimental Targets
+--------------------
+
+Experimental targets cannot be used in safety-critical contexts, the Ferrocene test suite
+was not executed on the target. No support is available for these targets, artifacts are
+provided as-is and the targets be could removed from future versions without notice.
```
tshepang, line 210:
```suggestion
provided as-is and the targets could be removed from future versions without notice.
```

ferrocene/doc/user-manual/src/targets/index.rst
```diff
@@ -113,15 +115,24 @@ Quality managed targets are not qualified, but can usually be qualified on reque
...
-Experimental targets cannot be used in safety-critical contexts, and there is
+Untested targets cannot be used in safety-critical contexts, and there is
 no guarantee that the Ferrocene test suite is successfully executed on the
 target. They are provided as a preview, with limited support available. They
 should not be used in production.
 
+For any of the following reasons, the target is not quality managed or qualified:
+
+* The Ferrocene test suite may not successfully execute on the target.
+* Some or all of the Ferrocene test suite does not yet execute on the target.
```
jonathanpallant, line 129: Points 1 and 2 sound the same?

Hoverbear: The intent was that point 1 communicates that the test suite has failing tests for some reason, while point 2 is that some suites cannot be run at all on the target. Let me reword it.

skade:
By and large fine, but I have the feeling that "Experimental" leans more into "this can be purchased" than "untested". I feel "untested" should more evoke the feeling "this does not go away for the time being, but the reasoning may or may not be changeable".

Hoverbear:
I think we might resurrect this later.

PR #1228, opened 2025-01-20
Publish Sphinx shared resources to PyPI

pietroalbini:
Now that our shared resources are properly packaged, this PR replaces the mirroring code and the `ferrocene/sphinx-shared-resources` repo with a workflow that publishes that package to PyPI. This will allow other projects using this (like the FLS) to just depend on the Python package instead of messing with submodules.

This will actually publish only if the version number is manually bumped, to avoid polluting PyPI.

I already ran the workflow in a separate branch and confirmed it works. Authentication is done through OIDC.


pietroalbini:
Marking as a draft since I'm pushing commits fixing bugs in the package.

.github/workflows/publish-sphinx-shared-resources.yml
```diff
@@ -0,0 +1,68 @@
...
+  publish:
+    name: Publish packages to PyPI
+    needs: [build]
+    runs-on: ubuntu-24.04
+
+    environment: publish-pypi-packages
+    permissions:
+      id-token: write
+    steps:
+      - name: Downlaod packages built by the previous step
```
Hoverbear, line 52:
```suggestion
      - name: Download packages built by the previous step
```

ferrocene/doc/sphinx-shared-resources/README.md
```diff
@@ -3,16 +3,15 @@
 
 # Ferrocene's shared Sphinx resources
 
-This repository contains the shared Sphinx resources we rely on for our
```
Hoverbear, line 6: This README.md could use a bit of words describing how users should bump the version number to publish a new version of the package. I know it's sorta in the `pyproject.toml`, but I don't think most folks would look there.

pietroalbini: This is the readme that will be published as a description on PyPI... If you want that, I can create a separate PyPI-specific readme.

pietroalbini:
This PR is ready for review.

Hoverbear (approved):

pietroalbini:
bors merge

PR #1240, opened 2025-01-22
Fix upstream pulling

pietroalbini:
This PR fixes the last issues with the pull upstream automation. I managed to fully run the script locally and generate a working PR, so no more fixes should be necessary. Example of a generated PR: https://github.com/ferrocene/ferrocene/pull/1239


Dajamante (approved):
bors merge

PR #1242, opened 2025-01-22
Note that our Armv8-R support is only for AArch32 mode.

jonathanpallant:
Armv8-R AArch64 processors exist (like the Cortex-R82) but they will require a different target, so I've changed the target description from

> Armv8-R bare-metal (hard-float)

to

> Armv8-R AArch32 bare-metal (hard-float)

Changes to other mentions of Armv8-R in the docs probably need to go upstream.

Here's how Arm refer to it: https://developer.arm.com/documentation/ddi0568/latest/


Hoverbear (approved):
bors merge

PR #1256, opened 2025-01-27
Add ui test for known problem rust-lang/rust#124364

Veykril:


tests/ui/ferrocene/problems/issue_124364.rs
```diff
@@ -0,0 +1,32 @@
+//@ run-pass
+//@ compile-flags: --O
```
pietroalbini, line 2: Why does this have the double `-`?

pietroalbini (commented):
Is this missing an stderr file too?

Veykril:
No, it runs to success on a fixed version, but panics (hits unreachable) on a faulty version (instead of segfaulting by oob). The `println` is only there to prevent llvm from inlining.

pietroalbini (approved):
bors merge

PR #1269, opened 2025-01-28
Quality Manage ARMv7E-M

Hoverbear:
We now quality manage ARMv7E-M targets. 🎉

### DIFF
--- a/ferrocene/doc/user-manual/src/targets/thumbv7em-none-eabihf.rst
+++ b/ferrocene/doc/user-manual/src/targets/thumbv7em-none-eabihf.rst
@@ -4,12 +4,29 @@
 .. _thumbv7em-none-eabihf:
 
 :target:`thumbv7em-none-eabihf`
-==============================
+===============================
 
-The ``thumbv7em-none-eabihf`` Ferrocene target provides support for
-bare-metal ARMv7E-M processors using the
-`T32 ISA <https://developer.arm.com/Architectures/T32%20Instruction%20Set%20Architecture>`_
-with hard floating point.
+.. note::
+   
+   For this target, each CPU must be tested separately. Currently, only support
+   for the Cortex-M4F CPU is Quality Managed.
+   
+   Other CPUs can become Quality Managed upon request within a short time frame.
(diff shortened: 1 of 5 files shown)


ferrocene/doc/user-manual/src/targets/thumbv7em-none-eabihf.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. _thumbv7em-none-eabihf:
+
+:target:`thumbv7em-none-eabihf`
+==============================
+
+The ``thumbv7em-none-eabihf`` Ferrocene target provides support for
+bare-metal ARMv7E-M processors using the
+`T32 ISA <https://developer.arm.com/Architectures/T32%20Instruction%20Set%20Architecture>`_
+with hard floating point.
```
jonathanpallant, line 12: The distinction between the targets is around the ABI, not the use of floating point instructions, and I think this line doesn't make that clear. 

Unsure if it's better to refer to or copy the wording at https://doc.rust-lang.org/nightly/rustc/platform-support/arm-none-eabi.html and https://doc.rust-lang.org/nightly/rustc/platform-support/thumbv7em-none-eabi.html

jonathanpallant (commented):
Left a comment.

Also, what should customers do if they want rustc to emit double precision floating point instructions for use on a Cortex-M7 with double precision FPU?

ferrocene/doc/user-manual/src/targets/thumbv7em-none-eabihf.rst
Hoverbear: I think we can clarify a bit and refer to that document.

Hoverbear:
>  rustc to emit double precision floating point instructions for use on a Cortex-M7 with double precision FPU?

Currently, this is not supported in a safety critical context as we do not allow `target-cpu` or `target-feature`. Customers would need to contact us so we could set up testing of such targets and carve out an exclusion.

jonathanpallant (approved):

Hoverbear:
bors merge

PR #1279, opened 2025-01-30
Improve the document signing UX

pietroalbini:
This PR makes two changes to the `./x sign` command, learning from today's signing procedure:

* Signing a document will now do nothing if the document is already signed by all roles, and the signatures are up to date. This behavior can be overridden by passing `--force`.
* Running `./x sign` without any argument will attempt to sign all documents.

The result of these two changes is that Florian will now need to run only `./x sign`, both during the first signature of the release cycle, and during followup signatures of the same release. In the first case the command will sign all documents, and in the second it will only sign the changed documents.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1626,6 +1626,7 @@ dependencies = [
  "anyhow",
  "asn1-rs",
  "base64 0.21.7",
+ "clap",
  "getrandom",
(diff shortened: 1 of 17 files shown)


Hoverbear (approved):
bors merge

PR #1287, opened 2025-02-03
Fix breadcrumbs not being copied if docs are built both fresh and not fresh

pietroalbini:
This was discovered on Friday during document signing. If in the same bootstrap invocation documents are built both fresh (due to a signature) and not fresh, the `BreadcrumbsAssets` step would only be executed once, resulting in one of the builds not having the breadcrumbs. To fix the problem we stop using a step to copy the breadcrumbs, and instead invoke a function.


Veykril (approved):

Veykril:
bors merge

Hoverbear (approved):
bors merge

PR #1294, opened 2025-02-04
Enforce the use of nightly Cargo when pulling from upstream

pietroalbini:
For a long time upstream has been using unstable features in their `Cargo.toml`s, which required us to change the pull script to use `RUSTC_BOOTSTRAP=1` whenever we need to invoke Cargo. That worked well until a few days ago, when upstream stabilized one of those features. Unfortunately, `RUSTC_BOOTSTRAP=1` works only when the feature is unstable, not when the feature is stabilized on the nightly channel but hasn't reached the stable channel yet.

This PR properly solves the problem by enforcing the use of a nightly toolchain when running the pull script.

### DIFF
--- a/ferrocene/tools/pull-upstream/pull.sh
+++ b/ferrocene/tools/pull-upstream/pull.sh
@@ -67,6 +67,15 @@ else
     upstream_commit="FETCH_HEAD"  # Latest commit in the branch we pull.
 fi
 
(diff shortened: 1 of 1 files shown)


amanjeev (approved):

Hoverbear:
bors merge

amanjeev:
bors retry

PR #1302, opened 2025-02-07
Show a warning banner on public-docs

pietroalbini:
This PR adds a warning banner to https://public-docs.ferrocene.dev, to clarify the documentation comes from the main branch and might be inaccurate:

image

Feel free to review the code, but the messaging will need tweaking (we'll discuss this outside the PR). This PR is best reviewed commit-by-commit.

# Testing

To get the banner on a local build (eg `./x.py doc ferrocene/doc/internal-procedures --open`, perform the step that `prepare-docs-to-github-pages.py` would do:

```bash
cp ferrocene/doc/public-docs-warning/public-docs-warning.css build/host/doc/public-docs-warning.css
```

### DIFF
--- a/ferrocene/doc/public-docs-warning/header.html
+++ b/ferrocene/doc/public-docs-warning/header.html
@@ -1,5 +1,5 @@
-{# SPDX-License-Identifier: MIT OR Apache-2.0 #}
-{# SPDX-FileCopyrightText: The Ferrocene Developers #}
+<!-- SPDX-License-Identifier: MIT OR Apache-2.0 -->
+<!-- SPDX-FileCopyrightText: The Ferrocene Developers -->
 
 <div class="ferrocene-public-docs-warning hidden">
     <div class="ferrocene-public-docs-warning-wrapper">
--- a/ferrocene/doc/public-docs-warning/header.html
+++ b/ferrocene/doc/public-docs-warning/header.html
@@ -1,0 +1,9 @@
+{# SPDX-License-Identifier: MIT OR Apache-2.0 #}
+{# SPDX-FileCopyrightText: The Ferrocene Developers #}
+
+<div class="ferrocene-public-docs-warning hidden">
+    <div class="ferrocene-public-docs-warning-wrapper">
+        <span>This is a preview from our development branch. It might be incomplete or inaccurate.</span>
+        <a href="https://docs.ferrocene.dev">Documentation access for customers &raquo;</a>
+    </div>
+</div>
--- a/ferrocene/doc/public-docs-warning/public-docs-warning.css
+++ b/ferrocene/doc/public-docs-warning/public-docs-warning.css
@@ -25,9 +25,10 @@
 
     display: flex;
     align-items: center;
-     justify-content: space-between;
+    justify-content: space-between;
 
     background: #f2bd0d;
+    font-size: 0.9em;
 }
 
 .ferrocene-public-docs-warning-wrapper a {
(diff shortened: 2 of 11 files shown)


src/bootstrap/src/ferrocene/doc.rs
```diff
@@ -188,39 +192,36 @@ impl<P: Step + IsSphinxBook> Step for SphinxBook<P> {
...
                 fs::read_to_string(&builder.src.join("src").join("version")).unwrap().trim(),
+            ))
+            .arg(format!(
+                "-Drustfmt_version={}",
+                builder.crates.get("rustfmt-nightly").unwrap().version,
             ));
 
+        // Include the public-docs warning message.
+        css_files.push(format!("{path_to_root}/../public-docs-warning.css"));
+        include_in_header.push(relative_path(&src, &public_docs_warning.join("header.html")));
```
Hoverbear, line 212: In my local copy (before copying the warning css over) I noticed this left a hanging request:
image

I'm not able to come up with a good way to make this not happen, but perhaps by default we could have this be an empty file or something, so that the request doesn't just hang?

ferrocene/doc/public-docs-warning/public-docs-warning.css
```diff
@@ -0,0 +1,48 @@
+/* SPDX-License-Identifier: MIT OR Apache-2.0 */
+/* SPDX-FileCopyrightText: The Ferrocene Developers */
+
+:root {
+    --header-bar-1-height: 2.5rem;
+}
+
+.ferrocene-public-docs-warning {
```
Hoverbear, line 8: The font size should be the same as `div.breadcrumbs` (0.9em)

ferrocene/doc/public-docs-warning/header.html
```diff
@@ -0,0 +1,9 @@
+<!-- SPDX-License-Identifier: MIT OR Apache-2.0 -->
+<!-- SPDX-FileCopyrightText: The Ferrocene Developers -->
+
+<div class="ferrocene-public-docs-warning hidden">
+    <div class="ferrocene-public-docs-warning-wrapper">
+        <span>This is a preview from our development branch. It might be incomplete or inaccurate.</span>
```
Hoverbear, line 6: My initial thought was that this should link to the "next" release notes to show the major changes in this branch compared to last stable, but I don't think I'm particularly married to the idea.

Hoverbear (commented):
Aesthetically I'd prefer the yellow warning bar to be under the blue breadcrumb bar. Does that work for you too? (If you have strong opinions I'm not terribly fussed either way)

src/bootstrap/src/ferrocene/doc.rs
pietroalbini: This is not the first time we use this approach. We also rely on it to show/hide the technical report in the documentation index, and to show/hide the breadcrumbs in the documents. You can see the 404 happening in public-docs for `breadcrumbs.js` for example.

pietroalbini:
Fixed the font size and swapped the two bars at the top. Also added a README with some documentation on the feature.

Hoverbear (approved):
Nice :) (Holding on merge since you req'd @plaindocs )

ferrocene/doc/public-docs-warning/header.html
```diff
@@ -0,0 +1,9 @@
...
+        <a href="https://docs.ferrocene.dev">Documentation access for customers &raquo;</a>
```
plaindocs, line 7: Do we want to mention 'stable' here?

plaindocs (approved):
Thanks, looks good to me.

tshepang:
bors merge

pietroalbini:
Let's check whether that was somehow a spurious failure, I'm having trouble reproducing this locally.

bors try

ferrocene/doc/public-docs-warning/header.html
pietroalbini: We have more than just stable docs.

pietroalbini:
bors try

Hoverbear (approved):
bors merge

Hoverbear:
bors merge

PR #1308, opened 2025-02-10
Add the `:upcoming:` Sphinx role

pietroalbini:
This PR adds the `:upcoming:` Sphinx role, which allows to mark a part of the document as "coming soon". The role displays a yellow badge, and can be put either inside of a paragraph/list/table, or in the header of a page/section:

```rst
This is a paragraph describing a new feature :upcoming:`25.05`.

Exciting new feature :upcoming:`25.05`
======================================
```

The badge will be shown if the documentation is built outside of a release branch (so, on the `nightly`, `pre-rolling` and `rolling` channels), or if the version of the current release channel is lower than the version specified as the argument. In other cases the badge will not be rendered. There is no difference of public-docs vs customer docs.

When put in the page title, it will also show up in the table of contents:

image

...and links to that page will include the badge too:

image

Also, for reference, here is how it looks in the middle of a paragraph:

image

### DIFF
--- a/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/upcoming.py
+++ b/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/upcoming.py
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+from docutils import nodes
+from sphinx.util.docutils import SphinxRole
+from sphinx.transforms import SphinxTransform
+from sphinx.util.logging import getLogger
+
+
+BADGE_CONTENT = "coming soon"
+
+
+class UpcomingRole(SphinxRole):
+    def run(self):
+        content = self.text
+
+        try:
+            major, minor = content.split(".")
+            role_version = (int(major), int(minor))
+        except ValueError:
+            return self.error(f"{self.text} is not a version in the YY.MM format")
+
+        if self.env.config.ferrocene_version != "rolling":
+            try:
+                current_major, current_minor, *_ = (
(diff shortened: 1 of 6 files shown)


ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/upcoming.py
```diff
@@ -0,0 +1,116 @@
...
+            f"{self.text} is not a version in the YY.MM format",
+            location=self.get_location(),
+        )
+        # Show the error message in the generated HTML.
+        problematic = nodes.problematic()
+        problematic += nodes.Text(message)
+        return [problematic], []
+
+
+# Annoyingly, Sphinx's :doc: role strips any rich formatting from the page title when generating the
```
Hoverbear, line 52: Oh, this is so annoying...

Hoverbear (approved):
bors merge

plaindocs (approved):
Caveat aside (not a blocker) this seems fine.

ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/upcoming.py
```diff
@@ -0,0 +1,116 @@
...
+BADGE_CONTENT = "coming soon"
+
+
+class UpcomingRole(SphinxRole):
+    def run(self):
+        content = self.text
+
+        try:
+            major, minor = content.split(".")
+            role_version = (int(major), int(minor))
```
Dajamante, line 19: So no point release planned?

pietroalbini: There will probably be point releases in the future, but the fixes included within them won't need the :upcoming: role.

PR #1312, opened 2025-02-11
Fix automation showing warning boilerplate when not needed

pietroalbini:
Discovered it in https://github.com/ferrocene/ferrocene/pull/1311. Turns out that `.strip("\n")` created a `[""]`, which is not coerced to false by Python's `if`.


tshepang (approved):

Dajamante (approved):
bors merge

PR #1313, opened 2025-02-11
Remove build dir on --clear build

Dajamante:
We need to remove `build/html` dir on clean build.

### DIFF
--- a/ferrocene/doc/sphinx-shared-resources/src/ferrocene_standalone_make_cli.py
+++ b/ferrocene/doc/sphinx-shared-resources/src/ferrocene_standalone_make_cli.py
@@ -11,7 +11,8 @@ from pathlib import Path
 import argparse
 import subprocess
 import sys
-
+import os
+import shutil
 
 # Automatically watch the following extra directories when --serve is used.
 EXTRA_WATCH_DIRS = ["exts", "themes"]
@@ -32,6 +33,9 @@ def build_docs(root, builder, clear, serve, debug):
         # Enable parallel builds:
         args += ["-j", "auto"]
     if clear:
+        html_path = Path(f"{dest}/html")
+        if os.path.exists():
+            shutil.rmtree(html_path)
         args.append("-E")
     if serve:
         for extra_watch_dir in EXTRA_WATCH_DIRS:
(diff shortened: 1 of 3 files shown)


ferrocene/doc/sphinx-shared-resources/src/ferrocene_standalone_make_cli.py
```diff
@@ -32,6 +33,9 @@ def build_docs(root, builder, clear, serve, debug):
         # Enable parallel builds:
         args += ["-j", "auto"]
     if clear:
+        html_path = Path(f"{dest}/html")
```
pietroalbini, line 36: I have a few comments on this:

* The path `"{dest}/html"` is not always correct, if the builder being passed to the function is not "html".
* We already define the output directory [here](https://github.com/ferrocene/ferrocene/pull/1313/files#diff-88bde951e18af58a80087724e0456631fa2b27158e084dcb6592d44387245298R59), to ensure this doesn't break when moving the output directory somewhere else we should use the same variable.
* Python's `pathlib` has a nice syntax to join paths: the division operator! You can do `dest / "html"` as an equivalent of the code you wrote, or Rust's `dest.join("html")`.

pietroalbini (changes requested):

ferrocene/doc/sphinx-shared-resources/src/ferrocene_standalone_make_cli.py
Dajamante: Thank you!
I packed all in an `output_dir` so we don't forget to modify both.

ferrocene/doc/sphinx-shared-resources/src/ferrocene_standalone_make_cli.py
```diff
@@ -32,6 +33,9 @@ def build_docs(root, builder, clear, serve, debug):
         # Enable parallel builds:
         args += ["-j", "auto"]
     if clear:
+        if os.path.exists():
+            shutil.rmtree(Path(output_dir))
```
pietroalbini, line 37:
```suggestion
        if output_dir.exists():
            shutil.rmtree(output_dir)
```

`Path` objects have an `exists()` method, and there is no need to wrap the shutil argument into a path (it already is!)

pietroalbini (commented):
Also, you will want to bump the version number in `ferrocene/doc/sphinx-shared-resources/pyproject.toml` to publish the new version of this to PyPI.

pietroalbini (approved):
bors merge

PR #1316, opened 2025-02-12
Put variable SCRIPTS nearer where it is used

Dajamante:
For clarity it is better to have `SCRIPTS` declared where they are used.

1. There is no need to create and pass the variable `SCRIPTS` to `run.sh`.

2. Also, when looking at CI last time, `workflow.yml` echoed all the scripts at once; it looked like the `traceability matrix` had failed when it was, in fact, `tidy` which was running and failing.

### DIFF
--- a/ferrocene/ci/run.sh
+++ b/ferrocene/ci/run.sh
@@ -45,8 +45,13 @@ ferrocene/ci/configure.sh
 #   inside the sources directory.
 export CARGO_HOME="$(pwd)/build/cargo-home"
 
-echo "Running:"
-echo "${SCRIPT}"
+declare -a SCRIPTS=(
+     "./x.py test tidy"
+     "./x.py check library compiler/rustc"
+     "./x.py run ferrocene/tools/traceability-matrix")
 
-# Give control to the actual CI script. $SCRIPT
-exec bash -euo pipefail -c "${SCRIPT}"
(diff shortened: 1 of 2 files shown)


ferrocene/ci/run.sh
```diff
@@ -45,8 +45,13 @@ ferrocene/ci/configure.sh
...
+declare -a SCRIPTS=(
+     "./x.py test tidy"
+     "./x.py check library compiler/rustc"
+     "./x.py run ferrocene/tools/traceability-matrix")
+
+for script in ${SCRIPTS[@]}; do
+  echo "Running $script"
+  # Give control to the actual CI script. $SCRIPT
+  exec bash -euo pipefail -c "${script}"
+done
```
tshepang, line 57: I would...
```suggestion
set +x # be verbose so that failures are more clear
./x.py test tidy
./x.py check library compiler/rustc
./x.py run ferrocene/tools/traceability-matrix
```
is more simple

Hoverbear:
I believe this change proposal may be related to a misunderstanding of how our CI works! `$SCRIPT` is used in the CI system to allow jobs to vary what is run. If you;'d like, I'd love to give you some more info in a more syncronous chat.

I'll close this for now, appreciate the initiative even still!

PR #1320, opened 2025-02-13
Add buttons to expand/collapse all elements of the traceability matrix

pietroalbini:
This PR adds buttons to expand and collapse all elements of the traceability matrix. Unfortunately clicking the buttons takes a couple of seconds to work, due to the large amount of `<details>` we have. I chose not to style the buttons since the OS provide good feedback about a button being "stuck" while it works, and recreating that was a bit overkill.

While I was at it I also tweaked the design of the page, to have the header bar and the section title be sticky at the top of the page.

[Screencast from 13-02-2025 14:52:07.webm](https://github.com/user-attachments/assets/44dafcc5-72bb-468b-bfa8-260b0a713fb7)

### DIFF
--- a/ferrocene/tools/traceability-matrix/templates/report.html
+++ b/ferrocene/tools/traceability-matrix/templates/report.html
@@ -69,7 +69,7 @@
         </table>
 
         {% if !matrix.unknown_annotations.is_empty() %}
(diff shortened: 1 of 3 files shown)


Hoverbear (approved):
bors merge

PR #1267, opened 2025-01-28
Remove x86_64 Mac

Hoverbear:
This platform is no longer experimentally supported.

(This is held for @Xylakant / marketings approval)

PR #1321, opened 2025-02-13
Add CI step to ensure `packages.toml` is up to date

pietroalbini:
As we discovered recently, we stopped releasing nightly builds for almost two weeks due to `packages.toml` requiring OxidOS even after we stopped building it. Releases stopped because the release process expects all files listed in `packages.toml` to be present, and (rightly) refuses to publish an incomplete version.

To prevent problems like this in the future, this PR adds a new CI script executed at the end of a bors job, which verifies that:

* All the tarballs referenced in `packages.toml` are actually uploaded to S3.
* All the files uploaded to S3 are either referenced by `packages.toml` or explicitly ignored in the script.

The second check was actually useful as it discovered two packages that we never referenced in `packages.toml` (even though we should've done that!), so the second commit in the PR fixes that. One of those was x86_64 macOS, which is pending removal in #1267, but until it's removed it should be added back to `packages.toml`.

cc @Hoverbear, the script will have to be called from the upcoming GHA workflow too.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -756,9 +756,15 @@ jobs:
       - run:
           name: Checkout the submodules
           command: ferrocene/ci/scripts/checkout-submodules.sh
+      # The verification has to happen before we upload the build metadata, as the release process
+      # understands build metadata being present as "this commit can be released".
+      - run:
+          name: Verify that all of the expected artifacts were uploaded
+          command: ferrocene/ci/scripts/validate-packages-toml.py ${CIRCLE_SHA1}
       - run:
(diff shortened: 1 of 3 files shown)


Hoverbear (approved):
Look good, didn't do extensive local testing but it reads correct and makes sense in terms of where it fits.

pietroalbini:
The failure was caused by CircleCI not having a GitHub token. Instead of generating a token and put it in the environment, we can just rely on the git CLI.

Veykril:
bors merge

Hoverbear (approved):
bors merge

PR #1314, opened 2025-02-11
Mark thumbv7e-m as qualification coming soon

Hoverbear:
Marks the thumbv7e-m qualification as coming soon!

PR #1322, opened 2025-02-13
(Idea) Use specific targets for ARMv7EM CPUs

Hoverbear:
This is an idea I was mulling over around how we can provide customers with the best experience for targets like `thumbv7em-none-eabi` which support different CPUs such as M4/M7/M55.

I'll add more detail here after speaking with @jonathanpallant and @pietroalbini about it. I simply wanted to see what it looks like in the end.

Based off https://github.com/ferrocene/ferrocene/pull/1314


compiler/rustc_target/src/spec/targets/thumbv7em_m4_none_eabi.rs
```diff
@@ -0,0 +1,10 @@
+use crate::spec::{Target, TargetMetadata};
+
+pub(crate) fn target() -> Target {
+    let mut target = super::thumbv7em_none_eabi::target();
```
Hoverbear, line 4: @jonathanpallant could you help me figure out what specific options we might want to set here to optimize for the M4 CPU?

compiler/rustc_target/src/spec/targets/thumbv7em_m4f_none_eabihf.rs
```diff
@@ -0,0 +1,10 @@
+use crate::spec::{Target, TargetMetadata};
+
+pub(crate) fn target() -> Target {
+    let mut target = super::thumbv7em_none_eabihf::target();
```
Hoverbear, line 4: @jonathanpallant could you help me figure out what specific options we might want to set here to optimize for the M4F CPU?

ferrocene/ci/configure.sh
```diff
@@ -193,11 +193,11 @@ if [[ is_internal ]]; then
     add --set target.aarch64-unknown-ferrocenecoretest.cc=aarch64-linux-gnu-gcc
     add --set target.aarch64-unknown-ferrocenecoretest.profiler=false # no profiling support
 
-    add --set target.thumbv7em-ferrocenecoretest-eabi.cc=arm-none-eabi-gcc
-    add --set target.thumbv7em-ferrocenecoretest-eabi.profiler=false # no profiling support
+    add --set target.thumbv7em.m4-ferrocenecoretest-eabi.cc=arm-none-eabi-gcc
```
Hoverbear, line 196: @pietroalbini Do you know how I can specify a target with a `.` in it without it becoming a big yak shave in `configure.py`?

compiler/rustc_target/src/spec/targets/thumbv7em_m4_none_eabi.rs
jonathanpallant: For the Cortex-M4, target-cpu=cortex-m4 is the only one I know of.

compiler/rustc_target/src/spec/targets/thumbv7em_m4f_none_eabihf.rs
jonathanpallant: For the Cortex-M4F, which only has one kind of FPU, you just need target-cpu=cortex-m4.

ferrocene/ci/configure.sh
pietroalbini: add --set target.\"thumbv7em.m4-ferrocenecoretest-eabi\".cc=arm-none-eabi-gcc

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -25,7 +25,7 @@ use crate::core::builder::{
 };
 use crate::core::config::{DebuginfoLevel, LlvmLibunwind, RustcLto, TargetSelection};
 use crate::ferrocene::code_coverage::ProfilerBuiltinsNoCore;
-use crate::ferrocene::secret_sauce::SecretSauceArtifacts;
+use crate::ferrocene::secret_sauce::https://www.youtube.com/shorts/zx1cfis-Oac?feature=share;
```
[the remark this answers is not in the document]
Hoverbear: 🤣

Hoverbear:
Turns out we'll do something else, closing this.

PR #1323, opened 2025-02-14
Ensure we use the correct uv shebang

pietroalbini:
While making an unrelated change I discovered that we were checking for the `#!/usr/bin/env -S uv` shebang, not `#!/usr/bin/env -S uv run` shebang. Turns out that the script still passed because it was checking the first line *starts* with that shebang, not that it matches the shebang exactly. This PR fixes that.


skade (approved):

pietroalbini:
bors merge

PR #1324, opened 2025-02-15
"development" is unclear on its own

tshepang:
I understood "development" as any change that happens, including maintenance work like fixing bugs, but others understood it differently


Hoverbear (commented):
bors merge

Hoverbear (approved):
bors merge

PR #1326, opened 2025-02-21
the component is used for both targets

tshepang:


Hoverbear (approved):
(Approving without merge because I want to slip in #1328 first)

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
```diff
@@ -138,6 +138,11 @@ Target Procedures
 
 Currently bare metal targets have a similar procedure for testing.
 
+.. note::
+
+   Currently, these targets use our *secret sauce*.
```
plaindocs, line 143: Does **these** refer to bare metal? Or to `aarch64-unknown-none`?

I saw some Zulip chat a while back about how we want to talk about 'secret sauce', but can't recall the details.

tshepang: `aarch64-unknown-none` is a bare metal target

regarding the_sauce_, this pr only moves the text around (and I don't remember what those details could be)

Hoverbear:
bors merge

PR #1328, opened 2025-02-21
[1.83] Make 1.83 a stable release

Hoverbear:
This needs to merge after or alongside #1327 .

Follows https://public-docs.ferrocene.dev/main/qualification/internal-procedures/release/stable.html#publishing-a-stable-release

PR #1327, opened 2025-02-21
[1.83] Add 1.83 technical report

Hoverbear:
Follows https://public-docs.ferrocene.dev/main/qualification/internal-procedures/release/stable.html#uploading-the-qualification-technical-report

PR #1330, opened 2025-02-21
Hoverbear/1.83 make stable with report

Hoverbear:
Joined #1327 and #1328 along with an extension to the timeout of jobs since the Windows one keeps timing out.


amanjeev (approved):
<3

PR #1342, opened 2025-02-25
Add standard qualified disclaimer to thumbv7em

Hoverbear:
While reviewing our docs I noticed we didn't have our typical disclaimer here like we do on other qualified targets.


tshepang (approved):
bors merge

tshepang:
re-adding `[merged-in:1.87]` label...

PR #1353, opened 2025-03-03
Stop pulling submodules in parallel in `pull-subtrees`

pietroalbini:
This will hopefully fix the spurious failure we've been seeing in the subtree pull job.

The error message referred to the tree being dirty, which shouldn't happen (it's a fresh checkout anyway). Also, only the first subtree fails, the others work correctly.

When I added debug logging in a job the problem stopped happening, probably because either the debugging code (a `git status`) updated some internal git data structures, or because it is a timing issue and the extra code made the job take slightly longer.

In my past experience a lot of these problems can happen with parallel subtree pulls, so this PR removes them.


tshepang (approved):
bors merge

tshepang:
bors retry

PR #1360, opened 2025-03-06
Move most of `configure.sh` to a bootstrap profile

pietroalbini:
Note: currently a draft as I needed to cherry-pick rust-lang/rust#137220 (it hasn't reached Ferrocene yet).

This PR moves most of the configuration for our CI away from `configure.sh` and into a bootstrap profile. With this change, all of the configuration about how a Ferrocene build should look like will reside in the profile, and `configure.sh` will only contain settings specific to our CI environment.

I also made a few changes:

* Added `ferrocene.channel = "auto-detect"` to load the channel from `ferrocene/ci/channel`. Before we were loading the contents of the file in the bash script, but that's not possible anymore. This mirrors the change we upstreamed in rust-lang/rust#137220.
* Added `FERROCENE_TEST_OUTCOMES_DIR` to load the test outcomes from a directory. Before this was done through the `RUST_CONFIGURE_ARGS`, but it was the only use of that environment variable. Adding `FERROCENE_TEST_OUTCOMES_DIR` removes the need to have `RUST_CONFIGURE_ARGS`.

For our CI nothing will change!

For our developers trying to reproduce CI failures locally, they will be able to set `profile = "ferrocene-dist"` at the top of their `config.toml` instead of running `configure.sh`. Note that this is not the recommended configuration to keep locally, the previous guidance of "don't use `configure.sh` day-to-day" applies to this profile as well.

For third parties building Ferrocene on their systems, instead of running `configure.sh` with the `OUTSIDE_FERROUS=1` environment variable, they'll be able to invoke `./configure --set profile=ferrocene-dist $their_flags_there`.

### DIFF
--- a/ferrocene/ci/configure.sh
+++ b/ferrocene/ci/configure.sh
@@ -16,15 +16,6 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-add() {
-    # Add each argument split by a `\t` instead of a space. This is needed to
-    # support flags with spaces in them.
-    while [[ $# -gt 0 ]]; do
-        RUST_CONFIGURE_ARGS="${RUST_CONFIGURE_ARGS-}"$'\t'"$1"
-        shift
-    done
-}
-
 if [[ -z "${CI+x}" ]]; then
     echo "error: this doesn't look like it's being executed in CI."
     echo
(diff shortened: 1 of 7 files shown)


ferrocene/ci/configure.sh
```diff
@@ -34,115 +25,45 @@ if [[ -z "${CI+x}" ]]; then
...
-# Helper function to check whether we should include things that only Ferrous
-# Systems has access to in the configuration.
-is_internal() {
-    if [[ -z "${OUTSIDE_FERROUS+x}" ]]; then
-        return 0
-    else
-        return 1
-    fi
-}
+if [[ -n "${OUTSIDE_FERROUS+x}" ]]; then
```
Hoverbear, line 33: Is this supposed to be `FERROCENE_UNPRIVILEGED_CI`?

Hoverbear (approved):
Rest looks good!

ferrocene/ci/configure.sh
pietroalbini: No. The purpose of this block is to provide a clear error and migration path for customers building Ferrocene from source using `OUTSIDE_FERROUS=1`.

pietroalbini:
bors merge

PR #1372, opened 2025-03-18
remove misleading edition args

tshepang:
This should imply that the tests only run on the specified edition.


pietroalbini (approved):
bors merge

PR #1373, opened 2025-03-18
remove dupe test

tshepang:
tests/ui/ferrocene/compiler-arguments/edition/edition_eq-delim.rs has the same exact content


pietroalbini (approved):
bors merge

PR #1375, opened 2025-03-20
add complete copyright/licensing info to source tarball

tshepang:
Upstream recently added an automated way of generating copyright files (https://github.com/rust-lang/rust/pull/135588), so we also need to do similar, because we have a different way of generating release tarballs.

### DIFF
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -108,6 +108,7 @@ impl Step for SourceTarball {
             "README.md",
             "RELEASES.md",
             "REUSE.toml",
+            "TRADEMARK.md",
             "config.example.toml",
             "Cargo.toml",
             "Cargo.lock",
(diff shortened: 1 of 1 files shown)


src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -138,6 +141,11 @@ impl Step for SourceTarball {
             subsetter.add_file(&builder.src, &builder.src.join(item));
         }
 
+        // Generate comprehensive copyright data and include the generated files
+        for item in builder.ensure(GenerateCopyright) {
+            subsetter.add_file(&builder.src, &item);
```
pietroalbini, line 146: What will be the path `COPYRIGHT.html` and `COPYRIGHT-library.html` will be copied to? They should be placed at the top level, not inside the `build/` directory in the tarball (if that's the case).

tshepang: that is what happens... will adjust

tshepang:
bors try

pietroalbini (approved):
bors merge

PR #1379, opened 2025-03-26
First batch of commits to fix code coverage gathering

pietroalbini:
This PR adds the first batch of commits needed to make code coverage gathering work again. Note that this PR does not fix the problem, the profdata files it generates are broken and I don't know why yet.

A lot of the removed complexity comes from removing the need to build `profiler_builtins` separately. On one hand, upstream changed profiler_builtins to always be `#![no_core]` (which removed the need to build a separate `profiler_builtins`), and with a few changes in the fourth and fifth commits I was able to make `-Cinstrument-coverage` work on the whole sysroot.

I also added code to process the `.profraw` files into `.profdata` (using the llvm tools), without relying on grcov.


library/sysroot/Cargo.toml
```diff
@@ -12,6 +12,10 @@ profiler_builtins = { path = "../profiler_builtins", optional = true }
 std = { path = "../std", public = true }
 test = { path = "../test", public = true }
 
+# Ferrocene addition: in bootstrap we need to set a libcore feature for code coverage testing, so we
+# need to add it as an explicit dependency here to do `--features core/$foo`.
+core = { path = "../core" }
```
Urhengulas, line 17: Why is it not optional?

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -1,90 +1,73 @@
...
-            "profiler_builtins_no_core={}",
-            profiler_builtins_no_core_path.to_str().unwrap()
-        ));
-        instrument_coverage_flags
-            .rustflags
-            .push("-Zprofiler_runtime=profiler_builtins_no_core".into());
-        instrument_coverage_flags.rustflags.push("-L".into());
-        instrument_coverage_flags.rustflags.push(lib_dir.to_str().unwrap().to_string());
+        let mut cmd = BootstrapCommand::new(self.llvm_bin_dir.join("llvm-profdata"));
+        cmd.arg("merge").arg("--sparse").arg("-o").arg(self.profdata_file);
```
Urhengulas, line 62: I _think_ you can just supply the directory instead of the individual files.

```suggestion
        cmd.arg("merge").arg("--sparse").arg("-o").arg(self.profdata_file).arg(self.profraw_dir);
```

library/sysroot/Cargo.toml
Urhengulas: Is it because it is implicitly pulled in via `std` anyways?

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -1,90 +1,73 @@
...
+use crate::builder::Builder;
+use crate::core::build_steps::llvm::Llvm;
+use crate::core::builder::Cargo;
 use crate::core::config::TargetSelection;
-use crate::{BootstrapCommand, Kind, Mode};
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub(crate) struct ProfilerBuiltinsNoCore {
-    pub(crate) target: TargetSelection,
+pub(crate) fn instrument_coverage(cargo: &mut Cargo) {
```
Urhengulas, line 9: Does not have public does it?

```suggestion
fn instrument_coverage(cargo: &mut Cargo) {
```

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -1,90 +1,73 @@
...
-
-        let profiler_builtins_no_core_dir = "profiler_builtins_no_core";
-        let target_dir =
-            builder.cargo_out(compiler, Mode::Std, target).join(profiler_builtins_no_core_dir);
+impl<'a> GatherCoverage<'a> {
+    pub(crate) fn new(
+        builder: &'a Builder<'a>,
+        cargo: &mut Cargo,
+        target: TargetSelection,
+        name: &str,
```
Urhengulas, line 25: What's the purpose of the `name` if it is always `"library"`?

src/bootstrap/src/ferrocene/code_coverage.rs
Urhengulas: We could also get rid of the function at all, since it is just one line and only called once.

Urhengulas (commented):
Mostly questions. Will test it locally now.

library/sysroot/Cargo.toml
pietroalbini: Yes, `core` should always be built when building the sysroot.

src/bootstrap/src/ferrocene/code_coverage.rs
pietroalbini: Yes that works, didn't know about that! I'll update the code to reflect this.

src/bootstrap/src/ferrocene/code_coverage.rs
pietroalbini: The function is also called from `src/core/build_steps/compile.rs` to instrument coverage for libcore itself. I'd still like to keep it around if we need to do extra code to instrument a piece of code, so that we can change it in only one place.

pietroalbini:
Addressed review comments. I also added an error when `--coverage` is passed without profiler support enabled, as we discussed on Zulip.

src/bootstrap/src/ferrocene/code_coverage.rs
Urhengulas: Maybe you can add a doc-comment how you expect it to be used?

Urhengulas (approved):

src/bootstrap/src/ferrocene/code_coverage.rs
pietroalbini: I will replace this with a different mechanism in a separate PR (for different reasons).

pietroalbini:
bors merge

Hoverbear (approved):
bors merge

PR #1381, opened 2025-03-28
"x test library/core" now also tests "library/coretests"

tshepang:
We previously changed to using `x test library/coretests` instead of `x test library/core` because unit tests were moved to a separate crate, "library/coretest". But later, a change was made so that `x test library/core` also tested unit tests in that crate.

One advantage of switching back to `x test library/core` is that doc tests are also executed.

(this should have been part of 8ae5c72a38d84d6e276ebe24196300637120bfe4)


Dajamante (approved):
bors merge

PR #1382, opened 2025-03-28
Finish the new code coverage instrumentation

pietroalbini:
This PR replaces the old coverage report instrumentation with a new implementation, based on llvm-cov. It's now possible to generate the coverage report for libcore by running:

```
./x test library/core --coverage=library --no-doc
```

As I developed the PR, I made the choice to replace the `--coverage` flag with the `--coverage=<kind>` flag, with `library` being the only supported kind. This splits the bootstrap code in "code needed for coverage in general" and "code needed for standard library coverage", making it clear what needs to be added if a new kind of coverage will be needed in the future. As an example, `-C instrument-coverage` will only be passed to libcore when the `library` coverage kind is used.

Measuring coverage is now split into three stages:

* **Instrumentation:** depending on the kind of coverage passed through the CLI, the `instrument_coverage()` function is called when compiling the object we want to measure. For example, when `--coverage=library` is passed, we pass that function when compiling the tested stage of the standard library. Crucially, we don't instrument the tests themselves.

* **Gathering:** the places that run the tests call `GatherCoverage::new`, which configures the test runner to place the raw coverage files in a temporary directory.

* **Reporting:** after the tests are executed `GatherCoverage::post_process` is called, which takes the raw coverage files and generates a report out of them.

There is only one known problem with this setup that I didn't find a solution for: absolute paths are present in the report, and it's not possible to remove them from llvm-cov as far as I can see. This means the user's home directory will be visible in the report, and it will be hard to reproduce the same exact HTML for the report.


pietroalbini:
Removed the limitation that only one step could gather coverage. Now an arbitrary number of steps can call `measure_coverage()`, and they will all be collected at the end of the build.

pietroalbini:
Also ignored the slow tests when measuring coverage.

Hoverbear:
I am unsure why 50+ tests have `mismatched data` here:

```
Merging together code coverage measurements
Generating code coverage report
warning: 55 functions have mismatched data
The coverage report is available at: file:///Users/ana/git/ferrocene/ferrocene/build/aarch64-apple-darwin/doc/coverage/library/index.html
```

You said you have 1 function like this on Linux.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -2,73 +2,261 @@ use std::path::PathBuf;
...
+        if self.report_dir.exists() {
+            builder.remove_dir(&self.report_dir);
+        }
+
+        builder.create_dir(&self.profraw_dir);
+        builder.create_dir(&self.report_dir);
     }
 }
+
+/////////////////////////////////////////////////////////
```
Hoverbear, line 204: 🙇‍♀️

pietroalbini:
bors merge

tshepang:
bors retry

src/bootstrap/src/core/config/flags.rs
```diff
@@ -32,6 +32,19 @@ pub enum Warnings {
     Default,
 }
 
+#[derive(Copy, Clone, Debug, ValueEnum, PartialEq, Eq)]
+pub enum FerroceneCoverageFor {
```
Urhengulas, line 36: Should this be non_exhaustive?

```suggestion
#[non_exhaustive]
pub enum FerroceneCoverageFor {
```

src/bootstrap/src/ferrocene/code_coverage.rs
Urhengulas: 🙇🏾

src/bootstrap/src/core/config/flags.rs
Urhengulas: No, we want it to break compilation if we add a new variant.

library/core/src/hint.rs
```diff
@@ -98,6 +98,7 @@ use crate::{intrinsics, ub_checks};
 #[stable(feature = "unreachable", since = "1.27.0")]
 #[rustc_const_stable(feature = "const_unreachable_unchecked", since = "1.57.0")]
 #[cfg_attr(miri, track_caller)] // even without panics, this helps for Miri backtraces
+#[coverage(off)] // Ferrocene addition: this function breaks llvm-cov
```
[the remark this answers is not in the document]
pietroalbini: `coverage(off)` already only applies when gathering coverage, so I don't really see a need for `cfg_attr`. We added the `cfg_attr` for `#[ignore]` because that'd also ignore those tests when executing them without coverage.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -2,73 +2,261 @@ use std::path::PathBuf;
...
         }
-        if profdata_file.exists() {
-            builder.remove(&profdata_file);
+        Some(existing) => {
+            if state != *existing {
+                eprintln!("error: cannot measure coverage in steps with different configuration!");
+                eprintln!("step 1 configuration: {state:?}");
+                eprintln!("step 2 configuration: {existing:?}");
+                exit!(1);
+            }
```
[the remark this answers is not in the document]
pietroalbini: I'll add some comments in the next PR.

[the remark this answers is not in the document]
pietroalbini: Oh, I just noticed that I already explained why that error exists in a comment block above. Is the reasoning clear after reading that comment block, or should I expand it?

PR #1388, opened 2025-04-03
Fix `./x dist ferrocene-src`

pvdrz:
This reverts commit 210ca50f4b5f12f6f417e6e29f63aef50be7ef07 and adds a permanent fix to the panic when distributing `ferrocene-src`.

This panic happened because the 2024 edition uses the 3 resolver which is version aware and requires the `rustc` binary to fetch the compiler version.

To fix it, we replicate 61339997194448cd8b18b422f32ced0126963213 in ferrocene's `dist` script.

### DIFF
--- a/src/bootstrap/Cargo.toml
+++ b/src/bootstrap/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
 name = "bootstrap"
 version = "0.0.0"
-# Ferrocene note:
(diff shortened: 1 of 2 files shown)


pvdrz:
bors try

pvdrz:
bors try

tshepang (approved):

pvdrz:
bors merge

PR #1395, opened 2025-04-08
include copyright files in docs

tshepang:
this pr has 2 main purposes
- adds copyright files to the doc tarball
- references the copyright files from Ferrocene docs index

context:
- at the moment, we only include copyright files in the ferrocene source tarball
- this pr is a convenience that lets users view the copyright files without having to first download and extract the source tarball

[internal ticket](https://app.clickup.com/t/8697wq828)

### DIFF
--- a/ferrocene/doc/index/index.html
+++ b/ferrocene/doc/index/index.html
@@ -166,6 +166,22 @@
                     </a>
                 </div>
             </section>
+            <section>
+                <h2>Licensing and copyrights</h2>
+                <p class="note">These files contain licensing and copyright information for all files distributed with Ferrocene.</p>
+                <div class="docs">
+                    <a href="COPYRIGHT.html">
+                        <h3>Ferrocene codebase</h3>
+                        <p>Open source licenses for the Ferrocene toolchain, for the tools used to develop it, and for the accompanying documentation.</p>
+                    </a>
+                    <a href="COPYRIGHT-library.html">
+                        <h3>Standard library</h3>
+                        <p>
+                            Open source licenses used by the runtime libraries (<b>core</b>, <b>alloc</b>, <b>std</b>) the compiler includes in executables it creates.
+                            All binaries produced by Ferrocene must comply with these licenses.</p>
+                    </a>
+                </div>
+            </section>
         </div>
     </body>
 </body>
--- a/src/bootstrap/src/core/builder/mod.rs
+++ b/src/bootstrap/src/core/builder/mod.rs
@@ -1071,6 +1071,7 @@ impl<'a> Builder<'a> {
                 doc::StyleGuide,
                 doc::Tidy,
                 crate::ferrocene::doc::AllSphinxDocuments,
+                crate::ferrocene::doc::CopyrightFiles,
                 // Basic Documents
                 crate::ferrocene::doc::Index,
                 crate::ferrocene::doc::Specification,
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -48,6 +48,14 @@ impl Step for Docs {
         let mut subsetter = Subsetter::new(builder, "ferrocene-docs", "share/doc/ferrocene/html");
         subsetter.add_directory(&doc_out, &doc_out);
 
+        // Generate comprehensive copyright data and include the generated files in the tarball
+        for path in builder.ensure(GenerateCopyright) {
+            if !builder.config.dry_run() {
+                // First arg gets the file placed in the root of the tarball, instead of in build/
+                subsetter.add_file(path.parent().unwrap(), &path);
+            }
+        }
+
         subsetter.into_tarballs().map(|tarball| tarball.generate()).collect()
     }
 }
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -26,6 +26,16 @@ pub(crate) struct Docs {
     pub(crate) target: TargetSelection,
 }
 
+// Generate comprehensive copyright data, and include the generated files in the tarball
+fn add_copyright_files(subsetter: &mut Subsetter<'_>, builder: &Builder<'_>) {
+    for path in builder.ensure(GenerateCopyright) {
+        if !builder.config.dry_run() {
+            // First arg gets the file placed in the root of the tarball, instead of in build/
+            subsetter.add_file(path.parent().unwrap(), &path);
+        }
+    }
+}
+
 impl Step for Docs {
     type Output = Vec<GeneratedTarball>;
     const DEFAULT: bool = true;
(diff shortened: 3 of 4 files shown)


src/bootstrap/src/core/builder/mod.rs
```diff
@@ -1069,6 +1069,11 @@ impl<'a> Builder<'a> {
                 doc::EditionGuide,
                 doc::StyleGuide,
                 doc::Tidy,
+                doc::Bootstrap,
+                doc::Releases,
+                doc::RunMakeSupport,
+                doc::BuildHelper,
+                doc::Compiletest,
```
pietroalbini, line 1076: Note that doing this will likely result in more merge conflicts going forward, as upstream will add new docs at the bottom of the list. In principle I really appreciate this change, but it might be painful for PR rotation.

ferrocene/doc/index/index.html
```diff
@@ -166,6 +166,19 @@ <h3>Compiler error codes</h3>
...
+                    <a href="COPYRIGHT.html">
+                        <h3>The Rust codebase</h3>
+                        <p>This does not include the standard libraries.</p>
+                    </a>
+                    <a href="COPYRIGHT-library.html">
+                        <h3>The Rust standard libraries</h3>
+                        <p>This is for the standard libraries (<b>core</b>, <b>alloc</b>, and <b>std</b>).</p>
+                    </a>
+                </div>
+            </section>
```
pietroalbini, line 181: @plaindocs thoughts around the wording here?

pietroalbini (commented):
Code changes looks great! Let's see if Sam has some thoughts too.

src/bootstrap/src/core/builder/mod.rs
tshepang: one solution is sorting them upstream, with the help of tidy, and then place Ferrocene items below that list

other than that, I don't mind an occassional merge conflict... we deal with them all the time, and I don't even remember the last time we had an upstream pull without conflicts

pietroalbini: I think we should still minimize the amount of merge conflicts we get: the situation is already bad as it is, if we don't keep it under control it will keep getting worse.

I don't think sorting them upstream is a viable option either, because some of the ordering in that list is load bearing, as in some steps are expected to be executed after other steps when running with no explicit arguments.

I'd keep things as they are right now.

tshepang: > I'd keep things as they are right now.

should I revert the commit

pietroalbini: Yes.

ferrocene/doc/index/index.html
plaindocs: Do you have a screenshot of the local preview? 

I'm wondering if we can avoid the repetition of the / this / the / this, but that's an idle though mostly.

skade: > This does not include the standard libraries.

That is a negative statement. What does it include!

tshepang: Screenshot 2025-04-08 at 20 59 40

tshepang: > > This does not include the standard libraries.
> 
> That is a negative statement. What does it include?

@skade glad you asked, because the text is wrong... the aim for the file is to include licensing and copyright info for all code in the repo, while *The Rust standard libraries* is a subset of that

tshepang: > I'm wondering if we can avoid the repetition of the / this / the / this, but that's an idle thought mostly.

@plaindocs an attempt https://github.com/ferrocene/ferrocene/pull/1395/commits/7ec7496ddfc79b4715fa09b6c03d165504c059ed

tshepang:
aiming to fold the last commit onto its parent before merge, since the parent made the wrong claim
> COPYRIGHT.html: This does not include (copyright and licensing info for) the standard libraries.

tshepang:
looks like a good backport candidate, so added the label @Hoverbear

ferrocene/doc/index/index.html
pietroalbini: I still don't think the text is accurate. The licensing information there is the open source licenses of all of Ferrocene, not just the Rust codebase. Yes, we are using the upstream tooling to generate it, but the upstream tooling is designed to also detect the licensing of our code (we developed it!).

Also, I think the text as it is right now is not really useful from an user's perspective. They don't care about repositories, they care about what they need to comply to. Possible wording:

## Open Source Licenses

**Ferrocene toolchain**  
Open source licenses used by the Ferrocene toolchain, including the compiler and all of the support tools used to develop it.

**Standard library**  
Open source licenses used by the runtime libraries (core, alloc, std) the compiler includes in your executable. All binaries produced by Ferrocene must comply with them.

ferrocene/doc/index/index.html
```diff
@@ -166,6 +166,23 @@ <h3>Compiler error codes</h3>
...
+                    Rust project codebase, http://github.com/rust-lang/rust.
+                </p>
+                <div class="docs">
+                    <a href="COPYRIGHT.html">
+                        <h3>Rust codebase</h3>
+                        <p>Represents all files in the repository.</p>
+                    </a>
+                    <a href="COPYRIGHT-library.html">
+                        <h3>Rust standard libraries</h3>
+                        <p>Represents only the standard libraries (<b>core</b>, <b>alloc</b>, and <b>std</b>).</p>
```
plaindocs, line 182: Couple of thoughts here (nitpicky, sorry)

- the way the links are currently it looks like you're linking to the thing, not to the license of the thing
- licences don't "represent" they apply to? 

So how about something like (in the same html structure you have now):

- [Rust license]() - applies to all files except the standard libraries below
- [Rust standard libraries license]() - applies to the standard libraries (<b>core</b>, <b>alloc</b>, and <b>std</b>

pietroalbini: Left similar feedback in https://github.com/ferrocene/ferrocene/pull/1395#discussion_r2034796737 too.

ferrocene/doc/index/index.html
tshepang: Much better text, thanks.

Note that I used *Rust* instead of *Ferrocene* because the information does not cover additional tools Ferrocene compared to upstream... we would need to regenerate license-metadata.json. Should I do that in this PR as well?

ferrocene/doc/index/index.html
tshepang: @plaindocs am happy for all the nitpicks that propose improved reading experiences

ferrocene/doc/index/index.html
pietroalbini: Yeah we should update license-metadata, and add a CI check ensuring it is up to date. Let's do that in a separate PR though.

tshepang:
Screenshot 2025-04-19 at 10 32 40

plaindocs (approved):
Looks good to me.

Hoverbear (approved):
bors merge

PR #1399, opened 2025-04-08
point releases are not guaranteed to happen

tshepang:
The text implies that each major releases is *always* accompanied by point releases, which is not true. We have in fact not had a single point release so far.


plaindocs (approved):
LGTM

tshepang:
bors merge

PR #1403, opened 2025-04-09
replace wrong/inconsistent copyright assignment

tshepang:
All open source Ferrocene work assigns copyright to "The Ferrocene Developers", but we had some files that assigns it to "Ferrous Systems GmbH".
This is also not accurate in that Ferrocene contributions are at times external to Ferrous Systems.

[internal ticket](https://app.clickup.com/t/86949v552)


pietroalbini (approved):
bors merge

PR #1409, opened 2025-04-11
Don't add coretests to core when testing crates individually

pietroalbini:
In CI we pass the `--ferrocene-test-one-crate-per-cargo-call` flag to ensure each crate shows up as a separate line in our qualification report, but a recent upstream change broke it:

image

Since core's test suite is split between core and coretests (and the alloc test suite is split between alloc and alloctests), upstream recently changed bootstrap to run both core and coretests when testing `library/core`. This is done *after* our implementation of `--ferrocene-test-one-crate-per-cargo-call`, which would result in coretests and alloctests being executed twice.

This PR fixes the problem by suppressing the implicit addition of coretests and alloctests when the flag is passed. They will still be tested standalone.

### DIFF
--- a/src/bootstrap/src/core/build_steps/test.rs
+++ b/src/bootstrap/src/core/build_steps/test.rs
@@ -2760,15 +2760,20 @@ impl Step for Crate {
         }
 
         let mut crates = self.crates.clone();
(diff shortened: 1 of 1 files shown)


Urhengulas (approved):
bors merge

PR #1415, opened 2025-04-14
Add testing categories to the CLI docs

pietroalbini:
When considering how to test CLI flags, we categorized them internally in informational, narrow impact, or high impact. We used this terminology when talking to assessors too, but we never actually wrote it down in the qualification material. This PR adds our categorization criteria to the evaluation report, and actually categorizes each CLI flag in the user manual.

Categories can be set by adding the required `:category:` option to the CLI argument directive, as it is currently done in `cli.rst`. The extension will verify the value belongs in an allowlist, and it will automatically style the list.

Along with the changes I spent a bit of time improving the styling of the CLI page:

| Before | After |
| --- | --- |
| image | image |

Other notable changes are the removal of the `-C` and `--codegen` from the page, since those are not actually flags the user can use, and the "unsupported in a qualified environment" caution snippet being added automatically by the extension.

### DIFF
--- a/ferrocene/doc/user-manual/src/rustc/cli.rst
+++ b/ferrocene/doc/user-manual/src/rustc/cli.rst
@@ -34,12 +34,6 @@
 
          $ rustc -A warnings -A absolute-paths-not-starting-with-crate
 
-   .. cli:option:: -C
-      :category: narrow
-
-      Compiler argument ``-C`` customizes code generation. The compiler
-      argument must be followed by at least one code generation option.
-
    .. cli:option:: -C codegen-units=<number>
       :category: narrow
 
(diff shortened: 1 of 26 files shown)


ferrocene/doc/user-manual/src/rustc/cli.rst
```diff
@@ -33,12 +34,8 @@
 
          $ rustc -A warnings -A absolute-paths-not-starting-with-crate
 
-   .. cli:option:: -C
```
Hoverbear, line 36: Seems like this should be marked `unqualified` not removed?

Hoverbear (approved):
Generally looks good, had one question

ferrocene/doc/user-manual/src/rustc/cli.rst
pietroalbini: `-C` and `--codegen` are not actually flags. They are just a prefix for the codegen flags like `-C opt-level=3`. In the documentation we have been treating the codegen flags as their own, so adding a standalone `-C`/`--codegen` is not really useful IMO.

Hoverbear (approved):
bors merge

PR #1417, opened 2025-04-15
Fix libcore not being tested in bare metal targets

pietroalbini:
In #1409 I changed bootstrap not to test `coretests` when requesting to test `core` *and* passing the `--ferrocene-one-crate-per-cargo-call` flag. Unfortunately that had the side effect to stop testing core on bare metal targets, since those crates were not added anymore.

This PR fixes that by implementing the flag after `coretests` is added.


Hoverbear (approved):
bors merge

tshepang:
bors retry

PR #1425, opened 2025-04-19
ci: fail when licensing info is stale

tshepang:
We do not want ot accidentally distribute Ferrocene with stale licensing information, so we automate the check in CI to help with that.


.github/workflows/commit.yml
```diff
@@ -68,9 +68,11 @@ jobs:
...
           FERROCENE_HOST: x86_64-unknown-linux-gnu
           SCRIPT: |
-            ./x.py test tidy
-            ./x.py check library compiler/rustc
-            ./x.py run ferrocene/tools/traceability-matrix
+            set -e
+            ./x test tidy
+            ./x check library compiler/rustc
+            ./x run ferrocene/tools/traceability-matrix
+            ./x test collect-license-metadata
```
Hoverbear, line 75: If this check takes more than a few seconds to run we should also probably extend the timeout to like 20min.

src/tools/collect-license-metadata/src/reuse.rs
```diff
@@ -34,7 +34,14 @@ pub(crate) fn collect(
...
     let output = Command::new(reuse_exe)
-        .args(&["--include-submodules", "spdx", "--add-license-concluded", "--creator-person=bors"])
+        // Ferrocene note: changed to assume reuse is installed via "uv" instead of directly
+        .args(&[
+            "reuse",
+            "--include-submodules",
+            "spdx",
+            "--add-license-concluded",
+            "--creator-person=bors",
+        ])
```
pietroalbini, line 44: We should not do this. The reason why `reuse` is searched on the filesystem and can be overridden in `config.toml` is that customers who build Ferrocene from source might do it from an airgapped system where uvx doesn't work. We should instead [install REUSE as a tool in CI](https://docs.astral.sh/uv/guides/tools/#installing-tools).

src/bootstrap/src/core/sanity.rs
```diff
@@ -200,7 +200,8 @@ than building it.
         .reuse
         .take()
         .map(|p| cmd_finder.must_have(p))
-        .or_else(|| cmd_finder.maybe_have("reuse"));
+        // Ferrocene note: changed to "uv" since we don't use Python directly
+        .or_else(|| cmd_finder.maybe_have("uvx"));
```
pietroalbini, line 204: Same as the other comment, we shouldn't change this.

pietroalbini (approved):

tshepang:
bors merge

PR #1443, opened 2025-05-02
Small improvement to the signing docs

pietroalbini:
Turns out a change was made while I was on leave removing the need to agree to your email address being published. Also, the third bullet point was worded in a confusing way. This PR reworks that section to address those concerns.


BriocheBerlin (commented):
Thanks for the update!
Could you add, as a first step, logging into the AWS CLI? Since this has to be done on the same day, it should be mentioned here as well, not only on the general set up page.

pietroalbini:
Done! Also, as you suggested in DMs, I changed the local env setup document to recommend installing `gnu-tar` on macOS and adding a note about bootstrap.toml vs config.toml.

BriocheBerlin (approved):
Perfect, thank you!

pietroalbini:
bors merge

PR #1445, opened 2025-05-02
Remove sphinx-needs

Hoverbear:
Related to: #786 https://github.com/ferrocene/ferrocene/commit/42c9a2d313cb818676c5d65ca75b81c908e2ac23

`sphinx-needs` is sadly non-deterministic and not currently suitable for our docs.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
+++ b/ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
@@ -18,61 +18,12 @@ documents the features of each extension.
 This extension implements "global" features that are not tied to any specific
 document, and should be imported in each document.
 
-The extension also enables ``sphinx.ext.intersphinx`` and ``sphinx-needs``
-without the need to manually enable them in ``conf.py``.
+The extension also enables ``sphinx.ext.intersphinx`` without the need to
+manually enable them in ``conf.py``.
 
-``sphinx-needs`` integration
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-The extension configures `sphinx-needs`_ to load IDs from other qualification
-documents. The following nodes are available:
-
-.. list-table::
-   :header-rows: 1
-
-   * - Sphinx directive
-     - ID prefix
-     - Purpose
-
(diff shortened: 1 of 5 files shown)


ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
```diff
@@ -18,62 +18,13 @@ documents the features of each extension.
 This extension implements "global" features that are not tied to any specific
 document, and should be imported in each document.
 
-The extension also enables ``sphinx.ext.intersphinx`` and ``sphinx-needs``
-without the need to manually enable them in ``conf.py``.
+The extension also enables ``sphinx.ext.intersphinx`` without the need to
+manually enable them in ``conf.py``.
```
tshepang, line 22:
```suggestion
manually enable it in ``conf.py``.
```

ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/__init__.py
```diff
@@ -56,8 +56,7 @@ def validate_config(app, config):
 
 
 def inject_version(app, config):
```
tshepang, line 58: is this fn used anywhere else... asking since it was introduced with addition of sphinx-needs support

tshepang:
other than lockfile fail, there are some references to sphinx-needs remaining

tshepang:
bors try

tshepang:
would squash this so we need revert just 1 commit when reverting the removal

will we also need to publish sphinx-shared-resources (to pypi) after this is merged

pietroalbini:
> will we also need to publish sphinx-shared-resources (to pypi) after this is merged

There is no manual publishing step, publishes happen if you update the version number in a PR: CI will then take care of publishing it. We need to publish new versions though only when we need the change somewhere else (known problems, fls...), and I don't think we care in the other books whether sphinx-needs is present.

Hoverbear:
We gonna squash this

pietroalbini:
Let the "unsubscribe from notifications" begin 😭

tshepang (approved):
bors merge

PR #1458, opened 2025-05-08
Update readme with new qualification

Hoverbear:
Noticed we were missing the new medical qualification in our README, so I updated it.


Hoverbear:
bors merge

PR #1441, opened 2025-05-01
Distribute `rust-analyzer-proc-macro-srv` in `rustc` package

Hoverbear:
Turns out we weren't doing this, and we should!

You can test this by doing `uv run x.py dist rustc` and expanding the `build/rustc-*.tar.xz` to see in `libexec` there is the `rust-analyzer-proc-macro-srv` binary.

Re: https://github.com/ferrocene/ferrocene/blob/466de87e29b9086f5cc5b2a4cd08bc4de1bcf4a1/src/bootstrap/defaults/bootstrap.ferrocene-dist.toml#L35

I do not *believe* we need to include this in `packages.toml`, since it's a component of `rustc`, not it's own package...

PR #1465, opened 2025-05-15
Include rust-analyzer tool when building rustc 

Hoverbear:
... so we can build `rust-analyzer-proc-macro-srv`

Follow up on #1441 . I didn't notice it wasn't working since I didn't use `split-tasks.py` during testing.

When reviewing this PR, please make sure the CI uploaded rustc package has `libexec/rust-analyzer-proc-macro-srv` and that the `rust-analyzer` package is also uploaded using `aws s3 ls` and `aws s3 cp` on the urls in the "Upload Artifacts to S3" stage of the dist jobs.

### DIFF
--- a/ferrocene/ci/split-tasks.py
+++ b/ferrocene/ci/split-tasks.py
@@ -87,7 +87,7 @@ JOBS_DEFINITION: JobsDefinition = {
 
         "oxidos": ["ferrocene-oxidos"],
 
-        "tools": ["rust-analyzer", "clippy", "rustfmt", "flip-link"],
+        "tools": ["clippy", "rustfmt", "flip-link"],
     },
 
     "test": {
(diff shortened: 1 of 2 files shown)


Hoverbear:
bors try

ferrocene/ci/split-tasks.py
```diff
@@ -87,7 +87,7 @@ def __lt__(self, other):
 
         "oxidos": ["ferrocene-oxidos"],
 
-        "tools": ["rust-analyzer", "clippy", "rustfmt", "flip-link"],
+        "tools": ["clippy", "rustfmt", "flip-link"],
```
tshepang, line 90: is it a good guess that this will build r-a in different job, one that builds rustc

PR #1479, opened 2025-05-26
doc: show the format to use when doing a manual subtree pull

tshepang:
The format of the argument to use is not obvious, so documenting it here.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/subtree-pulls.rst
+++ b/ferrocene/doc/internal-procedures/src/subtree-pulls.rst
@@ -82,9 +82,13 @@ You can also pass the name of a repository to only pull that specific repo::
 
    ferrocene/tools/pull-subtrees/pull.py <SUBTREE_REPO>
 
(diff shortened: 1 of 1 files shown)


ferrocene/doc/internal-procedures/src/subtree-pulls.rst
```diff
@@ -83,8 +83,11 @@ You can also pass the name of a repository to only pull that specific repo::
    ferrocene/tools/pull-subtrees/pull.py <SUBTREE_REPO>
 
 Note that `<SUBTREE_REPO>` must be the repository of the subtree and not its path.
+An example is ``ferrocene/tools/pull-subtrees/pull.py rust-lang/fls``,
```
Hoverbear, line 86: Could we break this into a separate paragraph? As it stands it appears kind of mooshed in with the note:

image

Hoverbear (approved):
bors merge

PR #1481, opened 2025-05-26
25.05 release notes use `target-with-tuple`

Hoverbear:
While reviewing the 25.05 release notes I realized that we weren't using `target-with-tuple` in the release notes, and that the "(Arm Cortex-M4, Arm Cortex-M7)" bit looked off, so I removed it.


Veykril (approved):

Hoverbear:
bors merge

PR #1422, opened 2025-04-17
Document coverage and add download-ci for it

pietroalbini:
:warning: This PR builds on top of https://github.com/ferrocene/ferrocene/pull/1421, commits from it are present here :warning: 

This PR adds the last piece of the puzzle for coverage, `coverage-outcomes = "download-ci";`, to download the latest coverage metadata from CI when generating a report locally. It also adds documentation for the whole coverage work.

PR #1486, opened 2025-05-28
code coverage: Instrument coretests

Urhengulas:
In the work to enable code coverage for libstd and libcore (#1422 and previous) we forgot to also instrument the test runner binaries (e.g. `coretests`). This is necessary because generic functions only get monomorphised in the test runner binaries. Without instrumenting the test runner, generic functions and methods on generic types will always show up as uncovered, even when they are tested.

As a reference, libstd/libcore is instrumented here: https://github.com/ferrocene/ferrocene/blob/ecd298803cfae0e0448de9a0d83a7d25377120d0/src/bootstrap/src/core/build_steps/compile.rs#L282:L307

### Slower

This change makes compiling and executing `coretests` slower. Therefore a few tests have to be disabled on coverage runs because otherwise they time out.   

### Higher

This change raises code coverage on libcore from 24.74 to 27.95%.

### Lower (🤔)

There is one surprising change in the report here: Instrumenting coretests raises the covered lines (as epxected), but it also (in a few modules) reduces the total number of coverable lines. Functions that are previously shown as uncovered are now not even shown as coverable in the report (e.g. Cell::is_writing).

This is related to inlining. These functions come back when inlining is disabled for those functions. So this problem will be fixed in a later PR, when solving the overall inlining problem.

### DIFF
--- a/library/coretests/tests/num/flt2dec/strategy/grisu.rs
+++ b/library/coretests/tests/num/flt2dec/strategy/grisu.rs
@@ -4,6 +4,7 @@ use super::super::*;
 
 #[test]
 #[cfg_attr(miri, ignore)] // Miri is too slow
+#[cfg_attr(ferrocene_coverage, ignore = "test too slow with coverage enabled")]
 fn test_cached_power() {
     assert_eq!(CACHED_POW10.first().unwrap().1, CACHED_POW10_FIRST_E);
     assert_eq!(CACHED_POW10.last().unwrap().1, CACHED_POW10_LAST_E);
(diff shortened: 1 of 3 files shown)


pvdrz (approved):

Hoverbear (commented):
bors merge

Hoverbear:
bors merge

Hoverbear:
bors cancel (need to land something for release, should be fine to merge this after/later https://github.com/ferrocene/ferrocene/pull/1494)

Urhengulas:
bors retry

Urhengulas:
bors merge=Hoverbear

PR #1469, opened 2025-05-16
[DO NOT MERGE] test doc signing process on Mac.

BriocheBerlin:
Contains a fix reverting https://github.com/ferrocene/ferrocene/commit/fccdce16804f9392aa049d3719256acfae8ce6d5.

PR #1494, opened 2025-05-30
Exclude `literal_block` with `source` attribute from Document ID

Hoverbear:
While testing signing (#1493, #1471, #1469) we noticed that Internal Procedures was creating a different Document ID when built locally vs CI. When we diffed the doctrees we noticed the following diff:

```diff
-            <literal_block force="False" highlight_args="{'linenostart': 1}" language="default" linenos="False" source="/home/ci/project/src/bootstrap/src/ferrocene/test_variants.rs" xml:space="preserve">("2021", &amp;[
+            <literal_block force="False" highlight_args="{'linenostart': 1}" language="default" linenos="False" source="/Users/ana/git/ferrocene/ferrocene/src/bootstrap/src/ferrocene/test_variants.rs" xml:space="preserve">("2021", &amp;[
```

This PR expands our existing exclusion to cover this as well.

### DIFF
--- a/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/document_id.py
+++ b/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/document_id.py
@@ -37,9 +37,10 @@ class Hasher:
         else:
             self.state.update(struct.pack("<Q", len(node.non_default_attributes())))
             for name, value in node.attlist():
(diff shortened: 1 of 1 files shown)


japaric (approved):
confirmed to fix repro issue when building docs in #1499 and #1493

tshepang:
bors merge

PR #1504, opened 2025-06-05
dist rust-std for `thumbv7em-ferrocenecoretest-eabihf`

japaric:


Hoverbear (approved):
This looks like enough for an exploration. We'll need to follow up with docs and consider things like naming a bit.

Hoverbear:
bors merge

PR #1194, opened 2025-01-08
WIP: Use GHA for our CI

Hoverbear:
This description is a work in progress.

PR #1505, opened 2025-06-05
Switchover to GHA

Hoverbear:
After #1194 merges, we'll run the CIs in parallel for a bit, then switch over.

This does that switchover and updates the docs.


jyn514:
this is outdated, the GHA work is happening elsewhere

PR #1508, opened 2025-06-09
ci: do not build stage 1 rustdoc

tshepang:
I do not know why we are doing this, so checking...


tshepang:
bors try

tshepang:
bors retry

tshepang:
bors try

Hoverbear:
Should also apply to #1194

tshepang:
> Post bootstrap re-org this makes sens I think https://blog.rust-lang.org/inside-rust/2025/05/29/redesigning-the-initial-bootstrap-sequence/

This is not related actually, and this PR checked if tests passed before we synced that stage0 redesign.

PR #1510, opened 2025-06-09
ci: reduce runtime of Commit workflow

tshepang:
Upstream now builds `rustc` before it can check `std`, because `std` can only be built with in-tree `rustc`. This is a lot for what is supposed to be a quick check.


Hoverbear:
Unfortunate. :( It looks like this saves about 3 minutes off the check?

tshepang:
make that 5m

Hoverbear (approved):
bors merge

PR #1511, opened 2025-06-09
compiletest: remove unused dep

tshepang:
The feature using this dep was removed before Ferrocne was open to the public


Hoverbear:
bors merge

PR #1513, opened 2025-06-10
Include rustbook in vendor tarball

Hoverbear:
A customer reported that while trying to build ferrocene from vendor tarballs they were receiving an error related to missing the dependencies of `rustbook`. This adds them.

An easy way to validate this PR is to note that `rustbook` is the only crate in the repo that depends on `mdbook-i18n-helpers` (according to `Cargo.lock`s) and is present in the result of `./x.py --stage 2 dist $(ferrocene/ci/split-tasks.py dist:src)`


tshepang (approved):
bors merge

PR #1515, opened 2025-06-10
Mark `Iterator` and `IntoIterator` as certified

pvdrz:
This is a minimal effort, meaning that none of the `Iterator` adapters are included on this PR.

Additionally `<[T]>::fill_with` is marked as certified to ensure that `for` loops work.

### DIFF
--- a/library/core/src/iter/mod.rs
+++ b/library/core/src/iter/mod.rs
@@ -475,8 +475,6 @@ pub use self::traits::FusedIterator;
 #[unstable(issue = "none", feature = "inplace_iteration")]
 #[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::InPlaceIterable;
-#[stable(feature = "rust1", since = "1.0.0")]
-pub use self::traits::Iterator;
 #[unstable(issue = "none", feature = "trusted_fused")]
 #[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::TrustedFused;
@@ -491,8 +489,10 @@ pub(crate) use self::traits::UncheckedIterator;
 #[stable(feature = "rust1", since = "1.0.0")]
 #[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::{
-    DoubleEndedIterator, ExactSizeIterator, Extend, FromIterator, IntoIterator, Product, Sum,
+    DoubleEndedIterator, ExactSizeIterator, Extend, FromIterator, Product, Sum,
 };
+#[stable(feature = "rust1", since = "1.0.0")]
+pub use self::traits::{IntoIterator, Iterator};
 
 #[cfg(not(feature = "ferrocene_certified"))]
 mod adapters;
--- a/library/core/src/iter/mod.rs
+++ b/library/core/src/iter/mod.rs
@@ -1,359 +1,360 @@
 //! Composable external iteration.
-//!
-//! If you've found yourself with a collection of some kind, and needed to
-//! perform an operation on the elements of said collection, you'll quickly run
-//! into 'iterators'. Iterators are heavily used in idiomatic Rust code, so
-//! it's worth becoming familiar with them.
-//!
-//! Before explaining more, let's talk about how this module is structured:
-//!
-//! # Organization
-//!
-//! This module is largely organized by type:
-//!
-//! * [Traits] are the core portion: these traits define what kind of iterators
-//!   exist and what you can do with them. The methods of these traits are worth
-//!   putting some extra study time into.
-//! * [Functions] provide some helpful ways to create some basic iterators.
-//! * [Structs] are often the return types of the various methods on this
-//!   module's traits. You'll usually want to look at the method that creates
-//!   the `struct`, rather than the `struct` itself. For more detail about why,
-//!   see '[Implementing Iterator](#implementing-iterator)'.
-//!
-//! [Traits]: #traits
-//! [Functions]: #functions
-//! [Structs]: #structs
-//!
-//! That's it! Let's dig into iterators.
-//!
-//! # Iterator
-//!
-//! The heart and soul of this module is the [`Iterator`] trait. The core of
-//! [`Iterator`] looks like this:
-//!
-//! ```
-//! trait Iterator {
-//!     type Item;
-//!     fn next(&mut self) -> Option<Self::Item>;
-//! }
-//! ```
-//!
-//! An iterator has a method, [`next`], which when called, returns an
-//! <code>[Option]\<Item></code>. Calling [`next`] will return [`Some(Item)`] as long as there
-//! are elements, and once they've all been exhausted, will return `None` to
-//! indicate that iteration is finished. Individual iterators may choose to
-//! resume iteration, and so calling [`next`] again may or may not eventually
-//! start returning [`Some(Item)`] again at some point (for example, see [`TryIter`]).
-//!
-//! [`Iterator`]'s full definition includes a number of other methods as well,
-//! but they are default methods, built on top of [`next`], and so you get
-//! them for free.
-//!
-//! Iterators are also composable, and it's common to chain them together to do
-//! more complex forms of processing. See the [Adapters](#adapters) section
-//! below for more details.
-//!
-//! [`Some(Item)`]: Some
-//! [`next`]: Iterator::next
-//! [`TryIter`]: ../../std/sync/mpsc/struct.TryIter.html
-//!
-//! # The three forms of iteration
-//!
-//! There are three common methods which can create iterators from a collection:
-//!
-//! * `iter()`, which iterates over `&T`.
-//! * `iter_mut()`, which iterates over `&mut T`.
-//! * `into_iter()`, which iterates over `T`.
-//!
-//! Various things in the standard library may implement one or more of the
-//! three, where appropriate.
-//!
-//! # Implementing Iterator
-//!
-//! Creating an iterator of your own involves two steps: creating a `struct` to
-//! hold the iterator's state, and then implementing [`Iterator`] for that `struct`.
-//! This is why there are so many `struct`s in this module: there is one for
-//! each iterator and iterator adapter.
-//!
-//! Let's make an iterator named `Counter` which counts from `1` to `5`:
-//!
-//! ```
-//! // First, the struct:
-//!
-//! /// An iterator which counts from one to five
-//! struct Counter {
-//!     count: usize,
-//! }
-//!
-//! // we want our count to start at one, so let's add a new() method to help.
-//! // This isn't strictly necessary, but is convenient. Note that we start
-//! // `count` at zero, we'll see why in `next()`'s implementation below.
-//! impl Counter {
-//!     fn new() -> Counter {
-//!         Counter { count: 0 }
-//!     }
-//! }
-//!
-//! // Then, we implement `Iterator` for our `Counter`:
-//!
-//! impl Iterator for Counter {
-//!     // we will be counting with usize
-//!     type Item = usize;
-//!
-//!     // next() is the only required method
-//!     fn next(&mut self) -> Option<Self::Item> {
-//!         // Increment our count. This is why we started at zero.
-//!         self.count += 1;
-//!
-//!         // Check to see if we've finished counting or not.
-//!         if self.count < 6 {
-//!             Some(self.count)
-//!         } else {
-//!             None
-//!         }
-//!     }
-//! }
-//!
-//! // And now we can use it!
-//!
-//! let mut counter = Counter::new();
-//!
-//! assert_eq!(counter.next(), Some(1));
-//! assert_eq!(counter.next(), Some(2));
-//! assert_eq!(counter.next(), Some(3));
-//! assert_eq!(counter.next(), Some(4));
-//! assert_eq!(counter.next(), Some(5));
-//! assert_eq!(counter.next(), None);
-//! ```
-//!
-//! Calling [`next`] this way gets repetitive. Rust has a construct which can
-//! call [`next`] on your iterator, until it reaches `None`. Let's go over that
-//! next.
-//!
-//! Also note that `Iterator` provides a default implementation of methods such as `nth` and `fold`
-//! which call `next` internally. However, it is also possible to write a custom implementation of
-//! methods like `nth` and `fold` if an iterator can compute them more efficiently without calling
-//! `next`.
-//!
-//! # `for` loops and `IntoIterator`
-//!
-//! Rust's `for` loop syntax is actually sugar for iterators. Here's a basic
-//! example of `for`:
-//!
-//! ```
-//! let values = vec![1, 2, 3, 4, 5];
-//!
-//! for x in values {
-//!     println!("{x}");
-//! }
-//! ```
-//!
-//! This will print the numbers one through five, each on their own line. But
-//! you'll notice something here: we never called anything on our vector to
-//! produce an iterator. What gives?
-//!
-//! There's a trait in the standard library for converting something into an
-//! iterator: [`IntoIterator`]. This trait has one method, [`into_iter`],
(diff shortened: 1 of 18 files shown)


library/core/src/num/uint_macros.rs
```diff
@@ -826,6 +826,15 @@ macro_rules! uint_impl {
...
             );
+            #[cfg(feature = "ferrocene_certified")]
+            assert_unsafe_precondition!(
+                check_language_ub,
+                "unchecked_sub cannot overflow",
+                (
+                    lhs: $SelfT = self,
+                    rhs: $SelfT = rhs,
+                ) => !lhs.overflowing_sub(rhs).1,
+            );
```
Urhengulas, line 837: Are `concat!` and `stringify!` the missing pieces here? Can they not be added to the subset?

library/core/src/iter/mod.rs
```diff
@@ -381,86 +382,122 @@ macro_rules! impl_fold_via_try_fold {
...
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use self::sources::{Successors, successors};
 #[stable(feature = "fused", since = "1.26.0")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::FusedIterator;
 #[unstable(issue = "none", feature = "inplace_iteration")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::InPlaceIterable;
-#[stable(feature = "rust1", since = "1.0.0")]
-pub use self::traits::Iterator;
```
Urhengulas, line 450: What is the reason for deleting it here and adding it in line 495?

library/core/src/iter/mod.rs
```diff
@@ -381,86 +382,122 @@ macro_rules! impl_fold_via_try_fold {
...
 pub use self::traits::TrustedLen;
 #[unstable(feature = "trusted_step", issue = "85731")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::TrustedStep;
+#[cfg(not(feature = "ferrocene_certified"))]
 pub(crate) use self::traits::UncheckedIterator;
 #[stable(feature = "rust1", since = "1.0.0")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use self::traits::{
-    DoubleEndedIterator, ExactSizeIterator, Extend, FromIterator, IntoIterator, Product, Sum,
```
Urhengulas, line 460: Can you please not delete it here, but just mark this use as not(ferrocene_certified) and mark the new use with ferrocene_certified?

```rust
pub use self::traits::{
    DoubleEndedIterator, ExactSizeIterator, Extend, FromIterator, IntoIterator, Product, Sum,
};
#[cfg(feature = "ferrocene_certified")]
pub use self::traits::IntoIter;
```
I think this should make it easier to resolve merge conflicts

library/core/src/iter/traits/collect.rs
```diff
@@ -314,11 +317,15 @@ pub trait IntoIterator {
...
+    #[cfg(not(feature = "ferrocene_certified"))]
     type Item = I::Item;
+    #[cfg(not(feature = "ferrocene_certified"))]
     type IntoIter = I;
 
     #[inline]
+    #[cfg(not(feature = "ferrocene_certified"))]
     fn into_iter(self) -> I {
         self
     }
```
Urhengulas, line 331: Is it not enough to mark the whole impl block as not(certified)? Why do the types and the method have to be marked?
```suggestion
#[cfg(not(feature = "ferrocene_certified"))]
impl<I: Iterator> IntoIterator for I {
    type Item = I::Item;
    type IntoIter = I;

    #[inline]
    fn into_iter(self) -> I {
        self
    }
```

library/core/src/iter/traits/iterator.rs
```diff
@@ -32,7 +38,7 @@ fn _assert_is_dyn_compatible(_: &dyn Iterator<Item = ()>) {}
     label = "`{Self}` is not an iterator",
     message = "`{Self}` is not an iterator"
 )]
-#[doc(notable_trait)]
+#[cfg_attr(not(feature = "ferrocene_certified"), doc(notable_trait))]
```
Urhengulas, line 41: What's the problem here? Does it need a feature or something?

library/core/src/iter/traits/mod.rs
```diff
@@ -1,24 +1,36 @@
+#[cfg(not(feature = "ferrocene_certified"))]
 mod accum;
+
```
Urhengulas, line 3:
```suggestion
```

library/core/src/iter/traits/mod.rs
```diff
@@ -1,24 +1,36 @@
...
     accum::{Product, Sum},
-    collect::{Extend, FromIterator, IntoIterator},
+    collect::{Extend, FromIterator},
     double_ended::DoubleEndedIterator,
     exact_size::ExactSizeIterator,
-    iterator::Iterator,
     marker::{FusedIterator, TrustedLen},
 };
+#[stable(feature = "rust1", since = "1.0.0")]
+pub use self::{collect::IntoIterator, iterator::Iterator};
```
Urhengulas, line 36: Same here. Please leave the original import untouched, just config it off and do the inverse config for the new import. This should make resolving merge conflicts in the future more easy.

library/core/src/lib.rs
```diff
@@ -89,7 +89,8 @@
 #![allow(internal_features)]
 #![deny(ffi_unwind_calls)]
 #![warn(unreachable_pub)]
-#![allow(unused_attributes)] // Ferrocene addition: accept coverage(off) no-ops
+#![allow(unused_attributes)]
+// Ferrocene addition: accept coverage(off) no-ops
```
Urhengulas, line 93:
```suggestion
#![allow(unused_attributes)] // Ferrocene addition: accept coverage(off) no-ops
```

library/core/src/prelude/v1.rs
```diff
@@ -53,8 +53,11 @@ pub use crate::convert::{AsMut, AsRef, From, Into};
 pub use crate::default::Default;
 #[stable(feature = "core_prelude", since = "1.4.0")]
 #[doc(no_inline)]
+pub use crate::iter::{IntoIterator, Iterator};
+#[stable(feature = "core_prelude", since = "1.4.0")]
+#[doc(no_inline)]
 #[cfg(not(feature = "ferrocene_certified"))]
-pub use crate::iter::{DoubleEndedIterator, ExactSizeIterator, Extend, IntoIterator, Iterator};
+pub use crate::iter::{DoubleEndedIterator, ExactSizeIterator, Extend};
```
Urhengulas, line 60: Same, please leave the original import untouched.

library/core/src/ptr/metadata.rs
```diff
@@ -4,9 +4,7 @@
 use crate::fmt;
 #[cfg(not(feature = "ferrocene_certified"))]
 use crate::hash::{Hash, Hasher};
-use crate::intrinsics::aggregate_raw_ptr;
-#[cfg(not(feature = "ferrocene_certified"))]
-use crate::intrinsics::ptr_metadata;
+use crate::intrinsics::{aggregate_raw_ptr, ptr_metadata};
```
Urhengulas, line 7: I assume the two imports were initially together but we split them up before and are now putting them together again? As in, now it is like it was in upstream before?

library/core/src/ptr/mod.rs
```diff
@@ -429,16 +429,16 @@ pub use crate::intrinsics::write_bytes;
 mod metadata;
 #[unstable(feature = "ptr_metadata", issue = "81513")]
 #[cfg(not(feature = "ferrocene_certified"))]
-pub use metadata::{DynMetadata, metadata};
+pub use metadata::DynMetadata;
+#[unstable(feature = "ptr_metadata", issue = "81513")]
+pub use metadata::metadata;
```
Urhengulas, line 434: Please leave the initial import untouched.

library/core/src/ptr/mut_ptr.rs
```diff
@@ -2125,6 +2179,7 @@ impl<T: ?Sized> Eq for *mut T {}
 impl<T: ?Sized> Ord for *mut T {
     #[inline]
     #[allow(ambiguous_wide_pointer_comparisons)]
+    #[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 2182: How does that work when the trait is implemented but without any method?

library/core/src/ptr/non_null.rs
```diff
@@ -1,13 +1,24 @@
...
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::ptr::Unique;
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::slice::{self, SliceIndex};
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::ub_checks::assert_unsafe_precondition;
-use crate::{fmt, hash, intrinsics, mem, ptr};
+#[cfg(not(feature = "ferrocene_certified"))]
+use crate::{fmt, hash, mem, ptr};
+use crate::{intrinsics, mem};
```
Urhengulas, line 21: Please don't touch original import

library/core/src/ptr/non_null.rs
```diff
@@ -421,40 +448,42 @@ impl<T: ?Sized> NonNull<T> {
...
-    /// ```
-    ///
-    /// [the module documentation]: crate::ptr#safety
+    // /// Returns a unique reference to the value. If the value may be uninitialized, [`as_uninit_mut`]
+    // /// must be used instead.
+    // ///
+    // /// For the shared counterpart see [`as_ref`].
+    // ///
+    // /// [`as_uninit_mut`]: NonNull::as_uninit_mut
+    // /// [`as_ref`]: NonNull::as_ref
```
Urhengulas, line 465: Would it work to use html links?

```suggestion
    /// Returns a unique reference to the value. If the value may be uninitialized, [`as_uninit_mut`]
    /// must be used instead.
    ///
    /// For the shared counterpart see [`as_ref`].
    ///
    // /// [`as_uninit_mut`]: NonNull::as_uninit_mut
    // /// [`as_ref`]: NonNull::as_ref
    /// [`as_uninit_mut`]: ./struct.NonNull.html#method.as_uninit_mut
    /// [`as_ref`]: ./struct.NonNull.html#method.as_ref
```

library/core/src/ptr/non_null.rs
```diff
@@ -1584,26 +1648,34 @@ impl<T: ?Sized> Clone for NonNull<T> {
...
 unsafe impl<T: ?Sized> PinCoerceUnsized for NonNull<T> {}
 
 #[unstable(feature = "pointer_like_trait", issue = "none")]
+#[cfg(not(feature = "ferrocene_certified"))]
 impl<T> core::marker::PointerLike for NonNull<T> {}
 
 #[stable(feature = "nonnull", since = "1.25.0")]
+#[cfg(not(feature = "ferrocene_certified"))]
 impl<T: ?Sized> fmt::Debug for NonNull<T> {
+    #[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 1669: Why does the method has to be config-offed and not only the impl block?

Urhengulas: Same for impl blocks below

library/core/src/slice/iter.rs
```diff
@@ -1,27 +1,42 @@
 //! Definitions of a bunch of iterators for `[T]`.
 
 #[macro_use] // import iterator! and forward_iterator!
+
```
Urhengulas, line 4:
```suggestion
```

library/core/src/slice/iter.rs
```diff
@@ -1,27 +1,42 @@
...
+#[cfg(not(feature = "ferrocene_certified"))]
+use crate::mem;
+use crate::mem::SizedTypeProperties;
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::num::NonZero;
-use crate::ptr::{NonNull, without_provenance, without_provenance_mut};
+#[cfg(not(feature = "ferrocene_certified"))]
+use crate::ptr::without_provenance;
+use crate::ptr::{NonNull, without_provenance_mut};
+#[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 24: Please leave the original imports untouched

library/core/src/slice/mod.rs
```diff
@@ -6,22 +6,33 @@
...
 use crate::intrinsics::{exact_div, unchecked_sub};
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::mem::{self, MaybeUninit, SizedTypeProperties};
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::num::NonZero;
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::ops::{OneSidedRange, OneSidedRangeBound, Range, RangeBounds, RangeInclusive};
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::panic::const_panic;
+use crate::ptr;
```
Urhengulas, line 21: Please leave the original import untouched.

library/core/src/slice/mod.rs
```diff
@@ -30,50 +41,76 @@ pub mod memchr;
...
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use index::{range, try_range};
 #[unstable(feature = "array_windows", issue = "75027")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub use iter::ArrayWindows;
+#[stable(feature = "rust1", since = "1.0.0")]
+#[cfg(not(feature = "ferrocene_certified"))]
+pub use iter::Iter;
+#[stable(feature = "rust1", since = "1.0.0")]
+pub use iter::IterMut;
```
Urhengulas, line 81: Please leave the original import untouched.

library/core/src/slice/mod.rs
```diff
@@ -111,7 +150,6 @@ impl<T> [T] {
     #[rustc_const_stable(feature = "const_slice_len", since = "1.39.0")]
     #[cfg_attr(not(bootstrap), rustc_no_implicit_autorefs)]
     #[inline]
-    #[must_use]
```
Urhengulas, line 114: Why remove this? If it has to be removed let's use a cfg_attr

library/core/src/num/uint_macros.rs
Urhengulas: I would like to avoid changing function bodies as much as possible.

pvdrz: yes they are. I can add them on a PR and then rebase this on top of that one.

library/core/src/iter/mod.rs
pvdrz: makes sense. I will!

library/core/src/ptr/metadata.rs
pvdrz: yes it should be that way

library/core/src/lib.rs
pvdrz: hmm that must have been `./x fmt`

library/core/src/iter/traits/collect.rs
pvdrz: It's easier for me to mark every single item as uncertified as it can be automated easily and then remove only the necessary attributes incrementally. I can change it if you want

library/core/src/iter/traits/iterator.rs
pvdrz: yes it needs a feature and it brings very little to the table imho

library/core/src/iter/mod.rs
pvdrz: formatting probably

library/core/src/iter/traits/collect.rs
Urhengulas: If it is not too much additional effort I would prefer only marking necessary items to keep the diff as small as possible.

PR #1516, opened 2025-06-10
Mark `AtomicU32` as certified

pvdrz:
cc @Urhengulas

### DIFF
--- a/library/core/src/lib.rs
+++ b/library/core/src/lib.rs
@@ -89,7 +89,8 @@
 #![allow(internal_features)]
 #![deny(ffi_unwind_calls)]
 #![warn(unreachable_pub)]
-#![allow(unused_attributes)] // Ferrocene addition: accept coverage(off) no-ops
+#![allow(unused_attributes)]
+// Ferrocene addition: accept coverage(off) no-ops
 // Do not check link redundancy on bootstraping phase
 #![allow(rustdoc::redundant_explicit_links)]
 #![warn(rustdoc::unescaped_backticks)]
@@ -136,8 +137,6 @@
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(abi_unadjusted))]
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(adt_const_params))]
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(cfg_sanitize))]
-#![cfg_attr(not(feature = "ferrocene_certified"), feature(cfg_target_has_atomic))]
-#![cfg_attr(not(feature = "ferrocene_certified"), feature(cfg_target_has_atomic_equal_alignment))]
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(const_precise_live_drops))]
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(deprecated_suggestion))]
 #![cfg_attr(not(feature = "ferrocene_certified"), feature(doc_cfg))]
@@ -166,6 +165,8 @@
 #![feature(allow_internal_unsafe)]
 #![feature(allow_internal_unstable)]
 #![feature(auto_traits)]
+#![feature(cfg_target_has_atomic)]
+#![feature(cfg_target_has_atomic_equal_alignment)]
 #![feature(cfg_ub_checks)]
 #![feature(const_trait_impl)]
 #![feature(decl_macro)]
@@ -357,7 +358,7 @@ pub mod async_iter;
 #[unstable(feature = "bstr", issue = "134915")]
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod bstr;
-#[cfg(not(feature = "ferrocene_certified"))]
+
 pub mod cell;
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod char;
@@ -385,7 +386,7 @@ pub mod random;
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod range;
 pub mod result;
-#[cfg(not(feature = "ferrocene_certified"))]
+
 pub mod sync;
 #[unstable(feature = "unsafe_binders", issue = "130516")]
 #[cfg(not(feature = "ferrocene_certified"))]
--- a/library/core/src/sync/atomic.rs
+++ b/library/core/src/sync/atomic.rs
@@ -3,235 +3,235 @@
 //! Atomic types provide primitive shared-memory communication between
 //! threads, and are the building blocks of other concurrent
 //! types.
-//!
-//! This module defines atomic versions of a select number of primitive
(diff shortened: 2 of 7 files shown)


library/core/src/lib.rs
```diff
@@ -89,7 +89,8 @@
 #![allow(internal_features)]
 #![deny(ffi_unwind_calls)]
 #![warn(unreachable_pub)]
-#![allow(unused_attributes)] // Ferrocene addition: accept coverage(off) no-ops
+#![allow(unused_attributes)]
+// Ferrocene addition: accept coverage(off) no-ops
```
Urhengulas, line 93:
```suggestion
#![allow(unused_attributes)] // Ferrocene addition: accept coverage(off) no-ops
```

Urhengulas: Or

```suggestion
// Ferrocene addition: accept coverage(off) no-ops
#![allow(unused_attributes)]
```

library/core/src/lib.rs
```diff
@@ -357,7 +358,7 @@ pub mod async_iter;
 #[unstable(feature = "bstr", issue = "134915")]
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod bstr;
-#[cfg(not(feature = "ferrocene_certified"))]
+
```
Urhengulas, line 361:
```suggestion
```

library/core/src/lib.rs
```diff
@@ -385,7 +386,7 @@ pub mod random;
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod range;
 pub mod result;
-#[cfg(not(feature = "ferrocene_certified"))]
+
```
Urhengulas, line 389:
```suggestion
```

library/core/src/sync/atomic.rs
```diff
@@ -244,9 +244,13 @@
 
 use self::Ordering::*;
 use crate::cell::UnsafeCell;
+#[cfg(not(feature = "ferrocene_certified"))]
+use crate::fmt;
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::hint::spin_loop;
-use crate::{fmt, intrinsics};
```
Urhengulas, line 248: Please leave the original import untouched.

library/core/src/sync/atomic.rs
```diff
@@ -428,7 +465,8 @@ unsafe impl<T> Sync for AtomicPtr<T> {}
 ///
 /// [nomicon]: ../../../nomicon/atomics.html
 #[stable(feature = "rust1", since = "1.0.0")]
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(not(feature = "ferrocene_certified"), derive(Debug, Eq, PartialEq, Hash))]
+#[derive(Copy, Clone)]
```
Urhengulas, line 469: Here I am not sure if it has a benefit to leave the original derive untouched, since we are inserting the cfg_attr anyways. What do you think is better to avoid merge conflicts, the way you have it now or:

```suggestion
#[cfg_attr(not(feature = "ferrocene_certified"), derive(Copy, Clone, Debug, Eq, PartialEq, Hash))]
#[cfg_attr(feature = "ferrocene_certified", derive(Copy, Clone))]
```

library/core/src/sync/atomic.rs
```diff
@@ -2554,73 +2649,73 @@ macro_rules! atomic_int {
...
+            // ///
+            // #[doc = concat!("let atomic_forty_two = ", stringify!($atomic_type), "::new(42);")]
+            // /// ```
             #[inline]
             #[$stable]
             #[$const_stable_new]
             #[must_use]
             pub const fn new(v: $int_type) -> Self {
-                Self {v: UnsafeCell::new(v)}
+                Self { v: UnsafeCell::new(v) }
```
Urhengulas, line 2665: Please no formatting
```suggestion
                Self {v: UnsafeCell::new(v)}
```

library/core/src/sync/atomic.rs
```diff
@@ -2629,50 +2724,50 @@ macro_rules! atomic_int {
                 unsafe { &*ptr.cast() }
             }
 
-
```
Urhengulas, line 2632: Please no formatting (add the line back)

library/core/src/sync/atomic.rs
```diff
@@ -2685,61 +2780,63 @@ macro_rules! atomic_int {
...
-            /// std::thread::scope(|s| {
-            ///     some_ints
-            ///         .iter()
-            ///         .enumerate()
-            ///         .for_each(|(idx, int)| {
-            ///             s.spawn(move || assert_eq!(int.load(Ordering::Relaxed), idx as _));
-            ///         })
-            /// });
-            /// ```
+            // #[doc = concat!("Get non-atomic access to a `&mut [", stringify!($atomic_type), "]` slice")]
```
Urhengulas, line 2783: This sould work with `concat!` and `stringify!`, no? They are helpful for the other PR as well, so let's add them I'd say.

library/core/src/sync/atomic.rs
```diff
@@ -2864,62 +2963,62 @@ macro_rules! atomic_int {
...
+            // /// assert_eq!(some_var.load(Ordering::Relaxed), 10);
+            // /// ```
             #[inline]
             #[$stable]
             #[deprecated(
                 since = "1.50.0",
-                note = "Use `compare_exchange` or `compare_exchange_weak` instead")
-            ]
+                note = "Use `compare_exchange` or `compare_exchange_weak` instead"
+            )]
```
Urhengulas, line 3011: Please no formatting
```suggestion
                note = "Use `compare_exchange` or `compare_exchange_weak` instead")
            ]
```

Urhengulas: Why is that module so badly formatted though? Is upstream not running tidy tests for it?

library/core/src/sync/atomic.rs
```diff
@@ -2864,62 +2963,62 @@ macro_rules! atomic_int {
...
             #[$cfg_cas]
             #[cfg_attr(miri, track_caller)] // even without panics, this helps for Miri backtraces
-            pub fn compare_and_swap(&self,
-                                    current: $int_type,
-                                    new: $int_type,
-                                    order: Ordering) -> $int_type {
-                match self.compare_exchange(current,
-                                            new,
-                                            order,
-                                            strongest_failure_ordering(order)) {
```
Urhengulas, line 2922: Please no formatting

library/core/src/sync/atomic.rs
```diff
@@ -3807,6 +3926,7 @@ fn strongest_failure_ordering(order: Ordering) -> Ordering {
 
 #[inline]
 #[cfg_attr(miri, track_caller)] // even without panics, this helps for Miri backtraces
+
```
Urhengulas, line 3929: Please not formatting (here and below)

PR #1526, opened 2025-06-13
Distribute other ferrocenecoretest for qualified platforms

Hoverbear:
Builds on #1504 by also distributing the other `ferrocenecoretest` targets.


ferrocene/tools/self-test/src/targets.rs
```diff
@@ -20,6 +20,11 @@ static SUPPORTED_TARGETS: &[TargetSpec] = &[
     TargetSpec { triple: "aarch64-unknown-linux-gnu", std: true, linker: Linker::HostCc },
     TargetSpec { triple: "aarch64-apple-darwin", std: true, linker: Linker::BundledLld },
     TargetSpec { triple: "aarch64-unknown-none", std: false, linker: Linker::BundledLld },
+    TargetSpec {
+        triple: "aarch64-unknown-ferrocenecoretest",
+        std: false,
```
tshepang: I think
```suggestion
        std: true,
```

tshepang:
needs some fmt ❤️

PR #1531, opened 2025-06-17
Add minimal CI to check libcore subset

Urhengulas:


.circleci/workflows.yml
```diff
@@ -121,6 +121,8 @@ jobs:
         ferrocene/ci/scripts/llvm_cache.py download
         ./x.py --stage 2 build library src/tools/rustdoc
         ./x.py --stage 1 build src/tools/rustdoc
+        # Build the certified subset of libcore
+        ./x.py --stage 1 build library/core --set rust.std-features='["ferrocene_certified"]'
```
Urhengulas: Should this be stage 1 or 2?

.circleci/workflows.yml
```diff
@@ -138,6 +140,8 @@ jobs:
         # to CI beforehand, for manual inspection in case of failures.
         ./x.py --stage 2 dist $(ferrocene/ci/split-tasks.py dist:docs)
         ./x.py --stage 2 test $(ferrocene/ci/split-tasks.py test:docs)
+        # Build the API docs of the certified subset of libcore
+        ./x.py doc library/core --set rust.std-features='["ferrocene_certified"]'
```
Urhengulas, line 144: Should the stage be specified as well?

.circleci/workflows.yml
tshepang: it should do the right thing if you remove the stage

.circleci/workflows.yml
tshepang: nope, no need

Urhengulas:
bors try

Urhengulas:
bors try

Urhengulas:
bors try

.circleci/workflows.yml
Urhengulas: It seems it is necessary. Otherwise it will try to build stage 1 std which will fail: https://app.circleci.com/pipelines/github/ferrocene/ferrocene/12537/workflows/b6ab1eaf-a431-4914-84ee-7450e6e93b70/jobs/75360

Urhengulas:
bors try

.circleci/workflows.yml
```diff
@@ -121,6 +121,8 @@ jobs:
         ferrocene/ci/scripts/llvm_cache.py download
         ./x.py --stage 2 build library src/tools/rustdoc
         ./x.py --stage 1 build src/tools/rustdoc
+        # Build the certified subset of libcore
+        ./x.py build --stage 1 library/core --set rust.std-features='["ferrocene_certified"]'
```
Hoverbear, line 125: This job creates a cache that all future jobs use, so I'm not sure this accomplishes what you want. I suspect that this will either damage or invalidate the cached libcore, possibly causing problems or making the build process longer.

I recommend removing this for now. Once we land custom targets we can add it correctly.

.circleci/workflows.yml
Hoverbear, line 144: I recommend removing this for now. Once we land custom targets we can add it correctly.

Hoverbear:
@Urhengulas with #1547 I think we can take a different approach

PR #1537, opened 2025-06-24
fls: fix and move impl trait precise capture text

tshepang:


ferrocene/doc/specification/src/types-and-traits.rst
```diff
@@ -2914,6 +2892,16 @@ Trait and Lifetime Bounds
    ForGenericParameterList ::=
        $$for$$ GenericParameterList
 
+   UseCaptures ::=
```
Veykril: `TypeBound` also needs to be updated to include `UseCapture` as one of its variants. Then we need to semantically forbid these bounds from appearing in type bound lists, unless the list is from an anonymous return type

ferrocene/doc/specification/src/types-and-traits.rst
```diff
@@ -2977,6 +2965,18 @@ If an :t:`outlives bound` applies to the :t:`type` of a :t:`function parameter`
 or to a :t:`return type`, then this bound also applies to the related
 :t:`function` as an :t:`implied bound`.
 
+:dp:`fls_69hqMjvNno9u`
+An :t:`use capture` is a :t:`generic parameter` referenced via keyword ``use`` within an :t:`anonymous return type`.
```
Veykril:
```suggestion
An :t:`use capture` is a :t:`generic parameter` referenced within an :t:`anonymous return type`.
```
Given this can happen implicitly we should not actually declare the explicit keyword syntax in the rule I think

ferrocene/doc/specification/src/types-and-traits.rst
```diff
@@ -2977,6 +2965,18 @@ If an :t:`outlives bound` applies to the :t:`type` of a :t:`function parameter`
...
+
+:dp:`fls_OnyR0Wsfk7KI`
+:t:`[use capture]s` shall only be used within :t:`[anonymous return type]s`.
+
+:dp:`fls_KgH6c5cC4S0G`
+An :t:`anonymous return type` that does not specify a list of :t:`[use capture]s` implicitly :t:`[use capture]s` all :t:`[type parameter]s` and :t:`[constant parameter]s` that are in :t:`scope`.
+
+:dp:`fls_iT9WCNfUZQnC`
+An :t:`anonymous return type` behaves as if it contained all its :t:`[use capture]s`.
```
Veykril: Hmm, so actually, I think we want to keep these where they were before, as this is declaring the concept of a generic capture (we call it use capture I guess), except for rule `fls_OnyR0Wsfk7KI`, which we keep down here but rephrase to be about the use capture *bounds* (the explicit syntax). It is a bit annoying as that means we get a forward jump for declarations, but that's fine.

```
:dp:`fls_OnyR0Wsfk7KI`
:t:`[use capture` :t:`[bound]s` shall only be used within :t:`[anonymous return type]s`.

```

ferrocene/doc/specification/src/types-and-traits.rst
tshepang: I assume this is what rust-lang/reference refers to as Abstract return type (`-> impl Trait`)

Veykril:
bors merge

tshepang:
bors retry

PR #1541, opened 2025-06-26
Generate docs for certified subset

pvdrz:
- **Expose `sync` and `cell`**
- **Generate documentation even if intra-doc links are broken**


Hoverbear (approved):
This does mean we need to certify some new features I think...

PR #1546, opened 2025-07-01
libcore-cert: Add doc comments back

Urhengulas:
Thanks to the `allow(rustdoc::broken_intra_doc_links)` which was added in https://github.com/ferrocene/ferrocene/pull/1541, we can just add back all doc-comments as rustdoc will not complain about broken links anymore.

Generating docs for full libcore and for the subset work (tested locally) and include the whole doc-comment:

```console
$ ./x doc library/core
$ ./x doc library/core --set rust.std-features="['ferrocene_certified']"
```

### DIFF
--- a/library/core/src/clone.rs
+++ b/library/core/src/clone.rs
@@ -1,38 +1,38 @@
 //! The `Clone` trait for types that cannot be 'implicitly copied'.
-// //!
-// //! In Rust, some simple types are "implicitly copyable" and when you
-// //! assign them or pass them as arguments, the receiver will get a copy,
-// //! leaving the original value in place. These types do not require
-// //! allocation to copy and do not have finalizers (i.e., they do not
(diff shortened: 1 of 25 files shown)


library/core/src/clone.rs
```diff
@@ -1,5 +1,4 @@
 //! The `Clone` trait for types that cannot be 'implicitly copied'.
-// FIXME(pvdrz): Fix docs
 // //!
```
pvdrz: Some of these shouldn't be removed as the docstrings are still commented.

Urhengulas: Oh, I forgot about those doc-strings

Urhengulas: Also added those doc-strings back

pvdrz (approved):

Hoverbear:
(@pvdrz please use bors)

PR #1547, opened 2025-07-01
Introduce certified targets

pvdrz:
This PR introduces two new targets with a certified `core` library: `aarch64-unknown-none.certified` and `x86_64-unknown-none.certified` (even if we don't want to certify the amd64 target, I consider it is useful for local testing as several people use amd64 as their daily driver).

We had to patch the in-tree version of `compiler-builtins` to introduce the new targets and patch the vendored dependency  on `compiler-builtins` by the in-tree version.

Other than that, this PR introduces some small improvements to the gating done over panic-related items to ensure that any code that compiles in the certified targets will compile in the uncertified targets as well.

### DIFF
--- a/compiler/rustc_target/src/spec/targets/aarch64_unknown_none_certified.rs
+++ b/compiler/rustc_target/src/spec/targets/aarch64_unknown_none_certified.rs
@@ -3,6 +3,8 @@ use crate::spec::Target;
 pub(crate) fn target() -> Target {
     let mut target = super::aarch64_unknown_none::target();
 
+    target.metadata.description =
+        target.metadata.description.map(|v| format!("{v} (certified)").into());
     target.metadata.host_tools = Some(false);
     target.metadata.tier = None;
     target.metadata.std = Some(false);
--- a/compiler/rustc_target/src/spec/targets/x86_64_unknown_none_certified.rs
+++ b/compiler/rustc_target/src/spec/targets/x86_64_unknown_none_certified.rs
@@ -3,6 +3,8 @@ use crate::spec::Target;
 pub(crate) fn target() -> Target {
     let mut target = super::x86_64_unknown_none::target();
 
+    target.metadata.description =
+        target.metadata.description.map(|v| format!("{v} (certified)").into());
     target.metadata.host_tools = Some(false);
     target.metadata.tier = None;
     target.metadata.std = Some(false);


compiler/rustc_target/src/spec/targets/aarch64_unknown_none_certified.rs
```diff
@@ -0,0 +1,9 @@
+use crate::spec::Target;
+
+pub(crate) fn target() -> Target {
+    let mut target = super::aarch64_unknown_none::target();
+
+    target.metadata = Default::default();
```
Hoverbear: I'm not sure `Default::default()` is correct, perhaps we can populate `description`, `host_tools` (false), `std` (false) for this and the other target?

compiler/rustc_target/src/spec/targets/x86_64_unknown_none_certified.rs
```diff
@@ -0,0 +1,9 @@
+use crate::spec::Target;
+
+pub(crate) fn target() -> Target {
+    let mut target = super::x86_64_unknown_none::target();
+
+    target.metadata = Default::default();
```
Hoverbear: I'm not sure `Default::default()` is correct, perhaps we can populate `description`, `host_tools` (false), `std` (false) for this and the other target?

library/Cargo.toml
```diff
@@ -52,3 +52,4 @@ rustc-std-workspace-alloc = { path = 'rustc-std-workspace-alloc' }
 rustc-std-workspace-std = { path = 'rustc-std-workspace-std' }
 
 libc = { path = '../ferrocene/library/libc' }
+compiler_builtins = { path = './compiler-builtins/compiler-builtins' }
```
Hoverbear, line 55: I believe we keep other similar subtrees in `../ferrocene/library`? (like libc)

library/core/src/panic/panic_info.rs
```diff
@@ -25,14 +32,24 @@ pub struct PanicInfo<'a> {
...
+        PanicInfo { message }
+    }
+}
+
+#[cfg(not(feature = "ferrocene_certified"))]
 impl<'a> PanicInfo<'a> {
     #[inline]
     pub(crate) fn new(
-        message: &'a fmt::Arguments<'a>,
+        message: &'a PanicFmt<'a>,
```
Hoverbear, line 55: This should probably get a `ferrocene annotation` so we don't get confused in a merge conflict

library/core/src/panicking.rs
```diff
@@ -31,9 +31,16 @@
 #[cfg(not(feature = "ferrocene_certified"))]
 use crate::fmt;
 use crate::intrinsics::const_eval_select;
+#[cfg(feature = "ferrocene_certified")]
+use crate::panic::PanicInfo;
 #[cfg(not(feature = "ferrocene_certified"))]
 use crate::panic::{Location, PanicInfo};
 
+#[cfg(feature = "ferrocene_certified")]
```
Hoverbear, line 40: Think we could get a comment for this?

src/bootstrap/src/utils/cc_detect.rs
```diff
@@ -110,23 +110,24 @@ pub fn find_target(build: &Build, target: TargetSelection) {
     // build script context (rust-lang/cc-rs#1225). Map `ferrocenecoretest` targets back to the
```
Hoverbear: We should update this comment

Hoverbear (commented):
Overall looks good, some minor feedback

compiler/rustc_target/src/spec/targets/aarch64_unknown_none_certified.rs
```diff
@@ -0,0 +1,12 @@
+use crate::spec::Target;
+
+pub(crate) fn target() -> Target {
+    let mut target = super::aarch64_unknown_none::target();
+
+    target.metadata.description = Some("Bare ARM64, hardfloat (certified)".into());
```
Hoverbear: Let's avoid a situation where this breaks if upstream changes names:

```suggestion
    target.metadata.description = target.metadata.description.map(|v| format!("{v} (certified)").into());
```

compiler/rustc_target/src/spec/mod.rs
```diff
@@ -2064,6 +2064,8 @@ supported_targets! {
     ("aarch64-unknown-ferrocenecoretest", aarch64_unknown_ferrocenecoretest),
     ("thumbv7em-ferrocenecoretest-eabi", thumbv7em_ferrocenecoretest_eabi),
     ("thumbv7em-ferrocenecoretest-eabihf", thumbv7em_ferrocenecoretest_eabihf),
+    ("x86_64-unknown-none.certified", x86_64_unknown_none_certified),
+    ("aarch64-unknown-none.certified", aarch64_unknown_none_certified),
```
Hoverbear, line 2068: How about `thumbv7em`? A future PR?

pvdrz: Yeah I'd think it would be best

library/compiler-builtins/compiler-builtins/src/lib.rs
Urhengulas, line 1: Is there any part of compiler-builtins that is not configged out?

library/core/Cargo.toml
```diff
@@ -25,8 +25,7 @@ bench = false
 # core specifically for profiling.
 ferrocene_inject_profiler_builtins = ["dep:profiler_builtins"]
 # Ferrocene addition: only include certfified code
-# requires panic_immediate_abort to allow panic while not needing fmt machinery
-ferrocene_certified = ["panic_immediate_abort"]
+ferrocene_certified = []
```
Urhengulas, line 28: Why was panic_immediate_abort removed? Is it not necessary anymore?

library/compiler-builtins/compiler-builtins/src/lib.rs
Urhengulas: If not we might be able to use `#![cfg(not(feature = "ferrocene_certified"))]`

Hoverbear:
(@pvdrz please use bors)

PR #1551, opened 2025-07-03
Add thumbv7em certified targets

pvdrz:
This PR adds the `thumbv7em-none-eabi.certified` and `thumbv7em-none-eabihf.certified` targets. Which are libcore-certified counterparts of `thumbv7em-none-eabi` and `thumbv7em-none-eabihf` respectively.


Hoverbear (approved):
bors merge

PR #1557, opened 2025-07-08
Merge `main` into `libcore-cert`

Urhengulas:
Main at adc20051308 from 04.07.2025.

### DIFF
--- a/library/core/src/ptr/non_null.rs
+++ b/library/core/src/ptr/non_null.rs
@@ -1,7 +1,5 @@
 #[cfg(not(feature = "ferrocene_certified"))]
 use crate::cmp::Ordering;
-#[cfg(feature = "ferrocene_certified")]
-use crate::marker::PointeeSized;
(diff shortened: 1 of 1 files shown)


library/core/src/cell.rs
```diff
@@ -814,34 +793,15 @@ impl Display for BorrowError {
 /// An error returned by [`RefCell::try_borrow_mut`].
 #[stable(feature = "try_borrow", since = "1.13.0")]
 #[non_exhaustive]
-<<<<<<< HEAD
 #[cfg(not(feature = "ferrocene_certified"))]
-=======
 #[derive(Debug)]
```
pvdrz, line 797: you mentioned something about leaving the `cfg(certified)` as the last attribute all the time. Should we do that here?

Urhengulas: This was more of a hunch to try to make it consistent in case we gonna try to process them later. But this does not seem to be necessary at the moment, and if we process them we probably want to use a proper parser anyways.

PR #1574, opened 2025-07-14
Use main backtrace, update subtree.

Hoverbear:
While trying to enable `miri` on Ferrocene I noted that our `backtrace-rs` subtree was out of date. Upstream uses a more recent version, so we can too: https://github.com/rust-lang/rust/tree/master/library


tshepang (approved):
bors merge

tshepang:
> After this merges, we should do an subtree pull!

https://github.com/ferrocene/ferrocene/pull/1579

PR #1577, opened 2025-07-15
Include dynamic libraries as instrumented binaries

pvdrz:
cc @hoverbear @Urhengulas


Urhengulas:
Let's not include it if it does not change the numbers.

PR #1581, opened 2025-07-15
Add `trpl` vendor to ferrocene-src

Hoverbear:
Following the ideas in #1513, we now add `trpl` as a customer who wants an offline vendor tarball now depends on it.


tshepang (approved):
bors merge

PR #1585, opened 2025-07-17
Fix relative_path in dry_run to make bootstrap tests pass locally

Hoverbear:
I was unable to run `./x.py test bootstrap` since a recent upstream pull as some of these tests use `--dry-run` and have an `out` path which is not relative to the `src`.

```
thread 'core::builder::tests::snapshot::test_exclude_kind' panicked at src/bootstrap/src/ferrocene/doc/mod.rs:876:41:
called `Result::unwrap()` on an `Err` value: StripPrefixError(())
```

Fixing this correctly is rather hard since in distribution builds Sphinx needs to be given paths relative to it's `conf.py`. Instead, we simply return `path` during `dry-run` since reproducability is not a concern during dry runs.


pvdrz (approved):
bors merge

pvdrz:
bors merge

PR #1588, opened 2025-07-18
Improve `core::array` coverage

pvdrz:
This brings line coverage for `core::array` to 100%


library/coretests/tests/array.rs
```diff
@@ -717,3 +717,10 @@ fn array_map_drops_unmapped_elements_on_panic() {
         assert_eq!(counter.load(Ordering::SeqCst), MAX);
     }
 }
+
+// This takes covers `<[T; N]>::eq` when it returns none.
```
skade, line 721: @pvdrz typo: "this takes" -> "this test"

PR #1590, opened 2025-07-20
Only report library coverage for libcore

Urhengulas:
If collecting library coverage, ignore all directories that are not part of libcore. Otherwise things like compiler-builtins, coretests etc. will be included. portable-simd and stdarch are kept as well, since parts of them are included into libcore via a path attribute.

This changes the report from this:

Screenshot from 2025-07-20 18-22-37

to this:

Screenshot from 2025-07-20 18-37-20


Hoverbear (approved):
bors merge

PR #1607, opened 2025-07-24
Note that low effort contributions are rejected

Hoverbear:
Largely follow the recent Rust project discussions, as well as our own internal team experiences, in taking a stance where we reject so-called "Low effort" changes such as those which might be "Vibe coded."


Dajamante (approved):
bors merge

PR #1618, opened 2025-07-30
Improve `core::char` coverage

pvdrz:
This brings line coverage for `core::ascii` to 99.5%

- **Cover `core::char::convert`**
- **Cover `core::char::methods`**

### DIFF
--- a/library/coretests/tests/lib.rs
+++ b/library/coretests/tests/lib.rs
@@ -14,7 +14,6 @@
 #![feature(bool_to_result)]
 #![feature(bstr)]
 #![feature(cfg_target_has_reliable_f16_f128)]
-#![feature(char_internals)]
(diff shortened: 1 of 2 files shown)


library/coretests/tests/lib.rs
```diff
@@ -14,6 +14,7 @@
 #![feature(bool_to_result)]
 #![feature(bstr)]
 #![feature(cfg_target_has_reliable_f16_f128)]
+#![feature(char_internals)]
```
Urhengulas: Let's add this into a separate section below

```
// Ferrocene additions:
//
// ... feature we add

Urhengulas (changes requested):

Hoverbear (approved):
bors merge

PR #1638, opened 2025-08-07
Fix `src/bootstrap/src/ferrocene` license

Urhengulas:
Problem: All files in `src/**` have their copyright attributed to the Rust project. Files in `src/bootstrap/src/ferrocene` should (also) have their copyright attributed to Ferrocene. Usually this is done by adding the license header. But unfortunately this was forgotten a few times and `reuse` does not detect this problem, since it only detects if a file has no copyright at all. 

Solution: Therefore we modify REUSE.toml to attribute all files in `src/bootstrap/src/ferrocene` to Ferrocene as well. We still aim to add the copyright header. But in case we forget, the copyright is still there.

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -248,19 +248,7 @@ SPDX-FileCopyrightText = "Jeremy Thomas"
 SPDX-License-Identifier = "MIT"
 
 [[annotations]]
-path = "requirements.txt"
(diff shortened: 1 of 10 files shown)


Hoverbear (approved):
bors merge

PR #1642, opened 2025-08-08
fls: update supported Rust version

tshepang:


tshepang:
done in https://github.com/ferrocene/ferrocene/pull/1645
my thinking was that, once our fls sync automation is fixed, we would get this change onto main (see https://github.com/rust-lang/fls/pull/577)

PR #1647, opened 2025-08-08
Make coverage-of-subset not introduce newlines

Hoverbear:
We noticed that the coverage lines seemed wrong in some measurements and I noticed that our coverage report in CI is using outcomes from a version of core with `coverage-of-subset` run, which [introduces new lines to the code](https://github.com/ferrocene/ferrocene/blob/cddbe05d4e0223ec24085522c63b386db674ca0f/ferrocene/tools/coverage-of-subset/src/main.rs#L118), but [the coverage report uses the actual libcore](https://github.com/ferrocene/ferrocene/blob/82e625f38c11d4ca03adbfbf230bec4ef22b5411/.circleci/workflows.yml#L138). This is a problem.

This seems to resolve it.

Before:

image

After

image


pvdrz:
bors merge

PR #1665, opened 2025-08-14
Mark `collect` macro internals as uncertified

pvdrz:
This fixes a small bug where we get the `collect` module for the certified subset coverage report even though it is gated.

### DIFF
--- a/library/core/src/iter/traits/collect.rs
+++ b/library/core/src/iter/traits/collect.rs
@@ -461,7 +461,6 @@ impl Extend<()> for () {
     fn extend_one(&mut self, _item: ()) {}
 }
 
-#[cfg(not(feature = "ferrocene_certified"))]
 macro_rules! spec_tuple_impl {
     (
         (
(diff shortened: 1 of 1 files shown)


Hoverbear (approved):
bors merge

library/core/src/iter/traits/collect.rs
```diff
@@ -461,7 +461,6 @@ impl Extend<()> for () {
     fn extend_one(&mut self, _item: ()) {}
 }
 
-#[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 464: I think we should keep this, no? Otherwise the macro can be called, which we don't want.

library/core/src/iter/traits/collect.rs
```diff
@@ -679,7 +684,6 @@ macro_rules! spec_tuple_impl {
     };
 }
 
-#[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 682: This one as well.

PR #1667, opened 2025-08-15
Add static copy of certified core library API docs

Urhengulas:
This PR adds a copy of the HTML files of running `./x doc library/core --set rust.std-features="['ferrocene_certified']"`. Currently has to updated manually whenever the certified core library subset changes. There will be automation soon.


Hoverbear:
bors cancel

Hoverbear:
We decided that it doesn't make sense to do this on main as it'll only drift. We'd rather land the automation.

PR #1668, opened 2025-08-15
Update static core assets (certified API docs, subset.tsv)

Urhengulas:


Hoverbear:
bors merge (aws failure again)

Hoverbear:
bors try

Hoverbear:
bors try-

Hoverbear:
bors try

japaric:
minor thing: the version next to the log has the "-dev" suffix. that is fine on main but on this branch which will be a stable release it should not and should read "1.88". I think if the docs are re-generated with `rust.channel = stable` then it should have the right string

image

japaric (approved):
aside from the version string in the API docs, which can be addressed in a follow-up PR, this looks good to me

Urhengulas:
bors merge

PR #1681, opened 2025-08-19
[1.88] Use LLD by default on x64 regardless of channel

Hoverbear:
Cherry pick a fix that was causing us not to use `lld` as the default on x86_64 Linux (*not* any of our qualified bare metal targets).

This should also get backported to other releases.


Hoverbear:
bors try

Hoverbear:
bors try

Urhengulas:
bors merge

PR #1682, opened 2025-08-20
Remove irrelevant libraries from lcov-*.info file

Urhengulas:
Before the libraries were in the info file but excluded when generating the report. Now they are not even in the info file. Also all removed paths are in one place now.

This is mostly a cleanup. The generated report stays the same.

### DIFF
--- a/src/bootstrap/src/ferrocene/code_coverage.rs
+++ b/src/bootstrap/src/ferrocene/code_coverage.rs
@@ -116,6 +116,7 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
     };
 
     let ignored_path_regexes: &[&str] = match state.coverage_for {
+        #[rustfmt::skip]
         FerroceneCoverageFor::Library => &[
             // Ignore Cargo dependencies:
             "\\.cargo/registry", // Without remap-path-prefix


src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -124,8 +124,11 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
             "ferrocene/library/backtrace-rs",
             "ferrocene/library/libc",
             "library/alloc",
+            "library/compiler-builtins",
+            "library/coretests",
```
Hoverbear: We found these were helpful to have in the `.info` file while debugging coverage issues, I'm not entirely sure this is a desirable change for that reason...

Urhengulas: Which ones specifically?

We could also take the fully different approach: Don't filter when generating the info file at all and include all of alloc, libc etc. and only filter at the report generation.

Hoverbear: When @pvdrz and I were debugging coverage we found we looked for tests in here to validate the CI did in fact run them and instrument them.

Urhengulas: But did it actually not run and instrument them sometimes?

jyn514:
We no longer use .info files for generating coverage reports.

PR #1624, opened 2025-08-01
Only show certified libcore in code coverage report

Urhengulas:
> [!NOTE]
> This is the third PR, of four expected ones, to merge the changes from libcore-cert into main.
>
>  Blocked on https://github.com/ferrocene/ferrocene/pull/1623.

This changes the coverage report to only show coverage for the certified subset of libcore, instead of the full uncertified libcore.

This is implemented by inserting the `#[coverage(off)]` attribute in all the places there is a `#[cfg(not(feature = "ferrocene_certified"))]` attribue. Doing this requires patching the coverage attribute to be accepted on top of all kinds of items. Before it was only allowed to add the coverage attribute on functions, modules and impl blocks.

The new report is on the left, it has way less entries than before. The old report is on the right.

new report
old report

# Testing

The changes can be tested like this:

```console
$ # gather code coverage for uncertified libcore
$ ./x test library/core --coverage=library --no-doc
$ # open the report at ./build/x86_64-unknown-linux-gnu/doc/coverage/library-x86_64-unknown-linux-gnu/index.html
$
$ # insert coverage attributes
$ ./x run coverage-of-subset
$
$ # gather code coverage for certified libcore
$ ./x test library/core --coverage=library --no-doc
$ # open the report at ./build/x86_64-unknown-linux-gnu/doc/coverage/library-x86_64-unknown-linux-gnu/index.html
```

PR #1697, opened 2025-08-25
Revert coverage everywhere

Hoverbear:
In #1624 we were a bit too liberal with changes and ended up altering compiler behavior in a way that could have impacted `#[coverage(off)]` users. This is not desirable, so we'll take a different road here.


Hoverbear:
bors try

tshepang (approved):

Hoverbear:
bosr try-

Hoverbear:
bors merge

ferrocene/tools/coverage-of-subset/src/main.rs
```diff
@@ -40,12 +56,21 @@ fn handle_dir(dir_path: impl AsRef<Path>) {
...
     fs::write(file_path, file_content).unwrap();
 }
 
-fn inject_coverage_attribute(file_content: &str) -> String {
-    file_content.replace(MARKER, &MARKER_PLUS_TO_BE_INSERTED)
+fn inject_coverage_attribute(mut file_content: String) -> String {
+    // This is, frankly, a very slow hack with a heavy hand, but it works for our purposes.
+    for applicable in APPLICABLE {
+        for i in 0..MAX_INDENTATION {
+            let valid_marker = format!("{MARKER}\n{}{}", " ".repeat(i), applicable);
```
Urhengulas, line 70: The problem with this approach is that it only catches the marker if it is the last attribute.

It will catch this:

```rust
#[no_mangle]
fn name() {}
```

but not this:

```rust
#[no_mangle]
fn name() {}
```

In the beginning we tried to mostly put the subset marker as the last attribute, but if we rely on this we should check this again.

Alternatively we could properly parse it with `syn`.

ferrocene/tools/coverage-of-subset/src/main.rs
```diff
@@ -3,16 +3,32 @@
...
+    "pub(crate) unsafe fn",
+    "pub(crate) const fn",
+    "pub(crate) const unsafe fn",
+    "impl",
+    "pub impl",
+    "pub(crate) impl",
+    "mod",
+    "pub mod",
+    "pub(crate) mod",
+];
```
Urhengulas, line 31: Let's add `unsafe impl`. At the moment this is not relevant, but it could become relevant e.g. for `unsafe impl<T: Clone> CopySpec for T` or `unsafe impl GenericRadix for $T`.
```suggestion
    "unsafe impl",
];
```

ferrocene/tools/coverage-of-subset/src/main.rs
```diff
@@ -3,16 +3,32 @@
...
+    "const unsafe fn",
+    "pub fn",
+    "pub unsafe fn",
+    "pub const fn",
+    "pub const unsafe fn",
+    "pub(crate) fn",
+    "pub(crate) unsafe fn",
+    "pub(crate) const fn",
+    "pub(crate) const unsafe fn",
+    "impl",
```
Urhengulas, line 23: I also briefly though we have to add `impl const`, but then I realised this is covered by `impl`.

ferrocene/tools/coverage-of-subset/src/main.rs
Urhengulas, line 70: Can we go in steps of four? The formatting should enforce that indentation is a multiple of four, no?

```suggestion
            let valid_marker = format!("{MARKER}\n{}{}", " ".repeat(i * 4), applicable);
```

If so we can probably reduce the size of `MAX_INDENTATION`.

Urhengulas: I just reviewed all instances of `#[cfg(not(feature = "ferrocene_certified"))]` in `library/core` on `main`. We mostly have the subset marker as the last attribute, except in two cases, but they are not relevant for coverage since it is on structs:

```diff
diff --git a/library/core/src/cell.rs b/library/core/src/cell.rs
index 75d07589d28..4df82b20632 100644
--- a/library/core/src/cell.rs
+++ b/library/core/src/cell.rs
@@ -762,8 +762,8 @@ pub struct RefCell<T: ?Sized> {
 /// An error returned by [`RefCell::try_borrow`].
 #[stable(feature = "try_borrow", since = "1.13.0")]
 #[non_exhaustive]
-#[cfg(not(feature = "ferrocene_certified"))]
 #[derive(Debug)]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub struct BorrowError {
     #[cfg(feature = "debug_refcell")]
     location: &'static crate::panic::Location<'static>,
@@ -790,8 +790,8 @@ fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 /// An error returned by [`RefCell::try_borrow_mut`].
 #[stable(feature = "try_borrow", since = "1.13.0")]
 #[non_exhaustive]
-#[cfg(not(feature = "ferrocene_certified"))]
 #[derive(Debug)]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub struct BorrowMutError {
     #[cfg(feature = "debug_refcell")]
     location: &'static crate::panic::Location<'static>,
```

library/core/src/array/mod.rs
```diff
@@ -43,7 +43,8 @@ mod iter;
 #[cfg(not(feature = "ferrocene_certified"))]
 pub(crate) use drain::drain_array_with;
 #[stable(feature = "array_value_iter", since = "1.51.0")]
-#[cfg(not(feature = "ferrocene_certified"))] /* blocked by IntoIter */
+// blocked by IntoIter
+#[cfg(not(feature = "ferrocene_certified"))]
```
Urhengulas, line 47: This re-formatting should not be necessary. In the past the previous comment style was accepted by CI.

Urhengulas:
> I just compared the numbers of the current approach and the one from this PR. It is the same, except that the new one does not insert a coverage(off) attribute here: https://github.com/ferrocene/ferrocene/blob/hoverbear/revert-coverage-everywhere/library/core/src/fmt/mod.rs#L136-L137

We need to add `trait`, `pub trait` etc.

ferrocene/tools/coverage-of-subset/src/main.rs
Hoverbear: This code takes all of like 2s to run on my machine I'm not particularly worried about perf. I'd rather just leave this as is if that's ok.

ferrocene/tools/coverage-of-subset/src/main.rs
Hoverbear: Added anyways

library/core/src/array/mod.rs
Hoverbear: It is necessary since it's not detected by the script if we do

```rust
#[cfg(not(feature = "ferrocene_certified"))] // Blah blah
mod Foo {}
```

Hoverbear:
@Urhengulas `trait` cannot have a coverage attribute

Urhengulas (approved):
bors merge

emig (approved):

Hoverbear:
bors merge

library/core/src/array/mod.rs
Urhengulas: Ah true!

PR #1707, opened 2025-08-27
do not break local users who run qnx scripts

tshepang:
REMOTE_TEST_SERVER_STAGE envvar is set in CI, and less likely to be set locally, so have a default that works for those cases. This is needed because --stage 0 no longer works, since the bootstrap change that does not allow in-tree libs to be built with downloaded rustc.

### DIFF
--- a/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
+++ b/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
@@ -113,7 +113,7 @@ cmd_prepare() {
 
     echo
     echo "===> building remote-test-server"
-    stage="${REMOTE_TEST_SERVER_STAGE-0}"
+    stage="${REMOTE_TEST_SERVER_STAGE-1}"
     ./x build src/tools/remote-test-server --target "${nto_target}" --stage "${stage}"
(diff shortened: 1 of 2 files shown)


ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
```diff
@@ -113,7 +113,10 @@ cmd_prepare() {
 
     echo
     echo "===> building remote-test-server"
-    stage="${REMOTE_TEST_SERVER_STAGE-0}"
+    stage="${REMOTE_TEST_SERVER_STAGE}"
```
Hoverbear: According to https://stackoverflow.com/a/16753536, I think you can make this

```
stage="${REMOTE_TEST_SERVER_STAGE-1}"
```
And get the same result?

tshepang: oh fun!

Hoverbear (approved):
bors merge

Hoverbear:
bors cancel

(just prioritizing)

Hoverbear:
bors merge

Hoverbear:
bors cancel

(lets hold this until the 1.88 backports happen please)

Hoverbear:
bors merge

PR #1714, opened 2025-08-29
Distribute the certified core library API docs

Urhengulas:
This PR automates building the API docs for the certified libraries, that is only the core library at the moment. This removes the need for the static HTML copy of the certified core API docs.

It introduces a new bootstrap command: `./x doc ferrocene-certified-api-docs`.

Also it makes the existing commands `./x doc ferrocene/doc` and `./x dist ferrocene-docs` also create the certified API docs.

### DIFF
--- a/src/bootstrap/src/core/build_steps/doc.rs
+++ b/src/bootstrap/src/core/build_steps/doc.rs
@@ -629,6 +629,12 @@ impl Std {
     ) -> Self {
         Std { build_compiler, target, format, crates: vec![] }
     }
+
+    // Ferrocene addition
+    pub(crate) fn with_crates(mut self, crates: Vec<String>) -> Self {
+        self.crates = crates;
+        self
+    }
 }
 
 impl Step for Std {
@@ -823,6 +829,7 @@ fn doc_std(
     // Ferrocene addition
     if target.contains("ferrocene.certified") {
         cargo.rustdocflag("--cfg=ferrocene_certified");
+        cargo.arg("--features").arg("ferrocene_certified");
     }
 
     if builder.config.library_docs_private_items {
--- a/src/bootstrap/src/core/config/target_selection.rs
+++ b/src/bootstrap/src/core/config/target_selection.rs
@@ -110,6 +110,20 @@ impl TargetSelection {
         }
     }
 
+    // Ferrocene addition
+    /// Map the target to the certified target that is based on it, if there is one
+    pub fn certified_equivalent(&self) -> Option<TargetSelection> {
+        let target_tuple = match self.triple.as_ref() {
+            "x86_64-unknown-linux-gnu" => "x86_64-unknown-ferrocene.certified",
+            "aarch64-unknown-none" => "aarch64-unknown-ferrocene.certified",
+            "thumbv7em-none-eabi" => "thumbv7em-ferrocene.certified-eabi",
+            "thumbv7em-none-eabihf" => "thumbv7em-ferrocene.certified-eabihf",
+            target if target.contains("ferrocene.certified") => target,
+            _ => return None,
+        };
+        Some(TargetSelection::from_user(target_tuple))
+    }
+
     /// Path to the file defining the custom target, if any.
     pub fn filepath(&self) -> Option<&Path> {
         self.file.as_ref().map(Path::new)
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -26,16 +26,6 @@ pub(crate) struct Docs {
     pub(crate) target: TargetSelection,
 }
 
-// Generate comprehensive copyright data, and include the generated files in the tarball
-fn add_copyright_files(subsetter: &mut Subsetter<'_>, builder: &Builder<'_>) {
-    for path in builder.ensure(GenerateCopyright) {
-        if !builder.config.dry_run() {
-            // First arg gets the file placed in the root of the tarball, instead of in build/
-            subsetter.add_file(path.parent().unwrap(), &path);
-        }
-    }
-}
-
 impl Step for Docs {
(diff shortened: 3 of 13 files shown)


Urhengulas:
bors try

Urhengulas:
bors try

Urhengulas:
bors try

Urhengulas:
bors try

src/bootstrap/src/core/build_steps/doc.rs
```diff
@@ -604,6 +604,15 @@ impl Std {
     ) -> Self {
         Std { build_compiler, target, format, crates: vec![] }
     }
+
+    pub(crate) fn new(
```
Hoverbear: While this is tempting, I don't think this is a divergence which benefits us to carry.

Urhengulas: How else can I set the `crates` field then?

Urhengulas:
bors try

src/bootstrap/src/core/build_steps/doc.rs
Urhengulas: I added `with_crates(mut self, crates: Vec<String>)` instead.

Urhengulas:
bors try

Hoverbear:
You can fix those tests by running them, then running `cargo insta review --manifest-path src/bootstrap/Cargo.toml` to update the snapshots. (Or manually apply the diff)

src/bootstrap/src/ferrocene/mod.rs
```diff
@@ -202,3 +202,15 @@ fn download_and_extract_ci_outcomes(builder: &Builder<'_>, kind: &str) -> PathBu
 
     extracted_dir.join("share").join("ferrocene").join(&name)
 }
+
+/// Map the target to the certified target that is based on it, if there is one
+pub(crate) fn target_to_certified_target(target: &TargetSelection) -> Option<TargetSelection> {
+    let target_tuple = match target.triple.as_ref() {
+        "x86_64-unknown-linux-gnu" => "x86_64-unknown-ferrocene.certified",
```
Hoverbear: Do you think it makes sense to do like this? So users can still pass `x86_64-unknown-ferrocene.certified` directly if they want?
```suggestion
        "x86_64-unknown-linux-gnu" | "x86_64-unknown-ferrocene.certified" => "x86_64-unknown-ferrocene.certified",
```

Also, I wonder if this is better in `TargetSelection` as `.certified_equivalent()` or something? It could go alongside the secret_sauce check: https://github.com/ferrocene/ferrocene/blob/19d6dde068ef192ff00f7f916ff8b1f322549eb7/src/bootstrap/src/core/config/target_selection.rs#L100

Urhengulas: I like it, I moved it to TargetSelection::certified_equivalent.

But I did not add the certified targets, because imo that's not the intended use.

src/bootstrap/src/core/config/target_selection.rs
```diff
@@ -110,6 +110,19 @@ impl TargetSelection {
         }
     }
 
+    // Ferrocene addition
+    /// Map the target to the certified target that is based on it, if there is one
+    pub fn certified_equivalent(&self) -> Option<TargetSelection> {
+        let target_tuple = match self.triple.as_ref() {
+            "x86_64-unknown-linux-gnu" => "x86_64-unknown-ferrocene.certified",
```
Hoverbear, line 117: It might avoid some confusion to map the certified targets to themselves here so folks can do `x.py dist ferrocene-certified-api-docs --target aarch-unknown-ferrocene.certified` and it still work.

```suggestion
            "x86_64-unknown-linux-gnu" | "x86_64-unknown-ferrocene.certified" => "x86_64-unknown-ferrocene.certified",
```

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -53,11 +44,19 @@ impl Step for Docs {
...
         builder.default_doc(&[]);
-        let doc_out = builder.out.join(&self.target.triple).join("doc");
+        let doc_out = builder.doc_out(self.target);
+
+        // Build the documentation for certified crates
+        //
+        // NOTE: must be called before .add_directory, since it places the
+        // certified API docs in the doc_out
+        if self.target.certified_equivalent().is_some() {
+            builder.ensure(CertifiedApiDocs { target: self.target });
```
Hoverbear, line 54: Likely we should include a log that we're ignoring like https://github.com/ferrocene/ferrocene/blob/b3308ea7b597f4cf1c2c0f8f8fc44e37c59dfe34/src/bootstrap/src/ferrocene/install.rs#L173-L175

Urhengulas:
Bors try

src/bootstrap/src/core/config/target_selection.rs
Urhengulas: Added a `target if target.contains("ferrocene.certified") => target` arm

src/bootstrap/src/core/build_steps/doc.rs
```diff
@@ -629,6 +629,11 @@ impl Std {
     ) -> Self {
         Std { build_compiler, target, format, crates: vec![] }
     }
+
```
Hoverbear, line 632:
```suggestion

// Ferrocene addition
```

src/bootstrap/src/core/build_steps/doc.rs
```diff
@@ -629,6 +629,12 @@ impl Std {
...
+// Ferrocene addition
```
Hoverbear:
```suggestion
    // Ferrocene addition
```

Hoverbear (approved):
bors merge

Urhengulas:
bors try

Hoverbear (approved):
bors merge

PR #1721, opened 2025-09-02
Add the core cert to the release notes

Hoverbear:
Seems we forgot this before. 😅

PR #1723, opened 2025-09-02
Use spaces, not dashes, in standards

Hoverbear:
@thejpster pointed out that *technically* it's 'IEC 61508', not 'IEC-61508' etc, so let's up our game here.

SHould wait for #1721, since it includes it


skade (approved):

ferrocene/doc/release-notes/src/25.08.0.rst
tshepang, line 1: do we want this change, given https://github.com/ferrocene/ferrocene/pull/1721

Hoverbear: This PR is atop that so this should come after and be ok

pvdrz (approved):
bors merge

PR #1732, opened 2025-09-08
Add error if out-of-tree dependencies are not found

Hoverbear:
Add a simple sanity check to ensure that we actually populate the out of tree deps and aren't impacted by #1491


tshepang (approved):

Hoverbear:
bors merge

PR #1733, opened 2025-09-08
Validate `-fuse-ld=lld` is passed, enable on aarch64

Hoverbear:
Also needed to edit the test harness to allow non-strict args.

[I am upstreaming the aarch64 enablement as well.](https://github.com/rust-lang/rust/pull/146604). -- I do not expect it to merge as is, as Upstream is changing some stuff to make this story better. :)

This is manually backported to 1.90/25.11: https://github.com/ferrocene/ferrocene/pull/1804

A basic smoke test:

```bash
#! /usr/bin/env bash

# Run from repo root
set +eux

SCRATCH=/tmp/dummy-scratch
DIST=build/dist

rm -rf $SCRATCH
mkdir $SCRATCH

rm -rf $DIST

RUST_BACKTRACE=1 ./x.py --stage 2 dist rustc cargo llvm-tools rust-std ferrocene-self-test

for folder in build/dist/*.tar.xz; do
    tar xvf $folder -C $SCRATCH
done
cd $SCRATCH
./bin/ferrocene-self-test
```

### DIFF
--- a/ferrocene/tools/self-test/src/targets.rs
+++ b/ferrocene/tools/self-test/src/targets.rs
@@ -441,10 +441,7 @@ mod tests {
         assert_fail("libcore-0123456789abcdef"); // No extension
     }
 
-    #[cfg(all(target_os = "linux", target_arch = "x86_64"))] // Only on x86_64 Linux since the test is specific to that.
-    #[test]
-    fn test_check_target_default_link_args() -> Result<(), Error> {
-        let tuple = "x86_64-unknown-linux-gnu";
+    fn check_target_default_link_args(tuple: &'static str) -> Result<(), Error> {
         let target = TargetSpec { tuple, std: true, linker: Linker::HostCc };
 
         let utils = TestUtils::new();
@@ -455,5 +452,17 @@ mod tests {
             .stdout("-fuse-ld=lld")
             .create();
         check_default_link_args(utils.sysroot(), &target)
+    }
+
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+    #[test]
+    fn test_check_target_default_link_args_x86_64() -> Result<(), Error> {
+        check_target_default_link_args("x86_64-unknown-linux-gnu")
+    }
+
+    #[cfg(all(target_os = "linux", target_arch = "aarch64"))]
+    #[test]
+    fn test_check_target_default_link_args_x86_64() -> Result<(), Error> {
+        check_target_default_link_args("aarch64-unknown-linux-gnu")
     }
 }
--- a/ferrocene/tools/self-test/src/targets.rs
+++ b/ferrocene/tools/self-test/src/targets.rs
@@ -4,10 +4,12 @@
 use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::path::Path;
+use std::process::Command;
 
 use crate::error::Error;
 use crate::linkers::Linker;
 use crate::report::Reporter;
+use crate::utils::run_command;
 
 static SUPPORTED_TARGETS: &[TargetSpec] = &[
     // Targets with architecture specific tuning
@@ -127,9 +129,50 @@ fn check_target(
         false => ["core", "alloc"].as_slice(),
     };
     check_libraries(target, &target_dir, expected_libs)?;
+    check_default_link_args(sysroot, target)?;
 
     reporter.success(&format!("target installed correctly: {}", target.tuple));
     Ok(CheckTargetOutcome::Found)
+}
+
+/// Check if the default link args for the target are what is expected
+fn check_default_link_args(sysroot: &Path, target: &TargetSpec) -> Result<(), Error> {
+    match target.linker {
+        Linker::HostCc => (),
+        Linker::BundledLld | Linker::CrossCc(_) => return Ok(()), // No default link args expected
+    }
+
+    let rustc = sysroot.join("bin").join("rustc");
+    let temp_dir = tempfile::tempdir()
+        .map_err(|error| Error::TemporaryCompilationDirectoryCreationFailed { error })?;
+    let temp_main = temp_dir.path().join("main.rs");
+    std::fs::write(&temp_main, "fn main() {}").map_err(|error| {
+        Error::WritingSampleProgramFailed {
+            name: "link_arg_example".into(),
+            dest: temp_main.clone(),
+            error,
+        }
+    })?;
+
+    let mut command = Command::new(rustc);
+    command.arg("--target");
+    command.arg(target.tuple);
+    command.arg("--print");
+    command.arg("link-args");
+    command.arg(temp_main);
+    let output = run_command(&mut command)
+        .map_err(|error| Error::sample_program_compilation_failed("link-args", error))?;
+
+    // All our `HostCc` targets require `-fuse-ld=lld` to be passed
+    let fuse_ld_arg = "-fuse-ld=lld";
+    if !output.stdout.contains(fuse_ld_arg) {
+        Err(Error::TargetDefaultLinkArgMissing {
+            target: target.tuple.into(),
+            link_arg: fuse_ld_arg.into(),
+        })
+    } else {
+        Ok(())
+    }
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


ferrocene/tools/self-test/src/test_utils/bin_builder.rs
```diff
@@ -21,6 +21,8 @@ pub(crate) struct BinBuilder<'a> {
     stderr: Option<String>,
     exit: Option<i32>,
     expected_args: Option<&'a [&'a str]>,
+    /// If true test for all args, if false only check for presence
```
Urhengulas: If `expected_args_strict` is false, we check that all expected args are present.

If `expected_args_strict` is true, we additionally check that there are no additional args found that are not expected (and we check that the order matches).

Correct?

If so we can maybe update the doc-comment to something like:

```suggestion
    /// If true test that all expected args and no additional ones are present.
    /// If false only check expected args are present, but there can be additional ones.
```

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -109,11 +111,55 @@ fn check_target(
...
+        return Err(Error::TargetDefaultLinkArgMissing {
+            target: target.tuple.into(),
+            link_arg: fuse_ld_arg.into(),
+        });
+    } else {
+        println!("Contains {fuse_ld_arg}!");
+        println!("{}", output.stdout);
+    }
+
+    Ok(())
```
Urhengulas:
```suggestion
    if !output.stdout.contains(fuse_ld_arg) {
        Err(Error::TargetDefaultLinkArgMissing {
            target: target.tuple.into(),
            link_arg: fuse_ld_arg.into(),
        })
    } else {
        println!("Contains {fuse_ld_arg}!");
        println!("{}", output.stdout);
        Ok(())
    }


```

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
```diff
@@ -161,7 +171,13 @@ fn main() {
...
         let expected = expected_args.split("\t").collect::<Vec<_>>();
         let found = std::env::args().skip(1).collect::<Vec<_>>();
-        assert_eq!(expected, found);
+        if Some("true") == option_env!("EXPECTED_ARGS_STRICT") {
+            assert_eq!(expected, found);
+        } else {
+            for item in expected {
+                assert!(found.contains(&item.to_string()));
+            }
+        }
```
Urhengulas: Iiuc the only arg that is not going to be expected is the `/tmp/.../main.rs` at the end. So another option would be to:

```suggestion
        if Some("true") == option_env!("LAST_ARG_IS_PATH") {
            assert_eq!(expected, found);
        } else {
            assert_eq!(&expected, &found[..found.len() - 1]);
        }
```

Imo this would simplify things a bit.

Urhengulas (approved):
LGTM, just left a few ideas for improvements

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -109,11 +111,55 @@ fn check_target(
         false => ["core", "alloc"].as_slice(),
     };
     check_libraries(target, &target_dir, expected_libs)?;
+    check_default_link_args(sysroot, target)?;
```
Urhengulas, line 132: Removing that line does not lead to any failing tests. Maybe we can add a test that verifies it fails when it should, in this or a future PR.

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
Hoverbear: That's true for this test yes, I was trying to write a solution that was more generally applicable though. Do you feel strongly about this?

Urhengulas: No. It just took me some time to understand why we need it. And part of why it took me long was that it is more general. But probably it makes sense for future use cases.

Urhengulas (approved):
bors merge

Urhengulas (approved):
bors merge

Hoverbear:
bors cancel

Hoverbear:
bors merge

Hoverbear:
bors try

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
```diff
@@ -161,7 +172,13 @@ fn main() {
     if let Some(expected_args) = option_env!("EXPECTED_ARGS") {
         let expected = expected_args.split("\t").collect::<Vec<_>>();
         let found = std::env::args().skip(1).collect::<Vec<_>>();
-        assert_eq!(expected, found);
+        if Some("true") == option_env!("EXPECTED_ARGS_STRICT") {
```
Hoverbear: @jyn514 said:
> the setup here scares me but this specific change looks ok. have you double checked that this fails if it gets a contains-but-not-exact match?

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -362,4 +440,20 @@ mod tests {
         assert_fail("libcore-0123456789abcdef.so"); // Different extension
         assert_fail("libcore-0123456789abcdef"); // No extension
     }
+
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))] // Only on x86_64 Linux since the test is specific to that.
```
Hoverbear: @jyn514 said:

> this test and the one below are for x86. Did you intend to test aarch64? or should we parameterize it over the target?

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
Hoverbear: Since `found` is a `Vec<String>` then the later `found.contains(item)` line should indeed be checking for exact matches. I suppose I could do `Vec::contains(found, &item.to_string())` and add a comment?

Hoverbear: Hmmm looking more this does not check *order* of args. Let me make sure it does that.

jyn514: to be clear: the code here looks fine, but I am asking you to manually test the self-test tool itself.

jyn514 (approved):
This seems ok to me. We've discussed internally that self-test is in flux, but I'm comfortable merging this for 1.90 and main.

I want to note explicitly: this new test only runs when self-check is running on the *host*, when the host has dynamically linked libc. If you are cross-compiling x64 -> aarch64, the test does not run. If you cross-compile aarch64 -> x64, it does not run. If you host-compile on ferrocene.facade or -musl, it does not run. Only if you are on aarch64-gnu or x86_64-gnu does the test do anything.

Hoverbear:
bors merge

Hoverbear:
bors merge-
bors cancel

jyn514:
@Hoverbear i think this is ready to merge and just waiting on you to rebase?

Hoverbear:
Yup, let me get it in

Hoverbear:
Upstream landed `default-linker-linux-override`, so adapted to that.

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
```diff
@@ -21,6 +21,10 @@ pub(crate) struct BinBuilder<'a> {
     stderr: Option<String>,
     exit: Option<i32>,
     expected_args: Option<&'a [&'a str]>,
+    /// If true test that all expected args and no additional ones are present.
+    /// If false only check expected args are present and in order,
+    /// but there can be additional arguments before, between, or after.
+    expected_args_strict: bool,
```
jyn514, line 27: non-blocking nit: i think this would be nicer as an enum so you can write the doc-comments on each variant

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
jyn514: did you ever get a chance to manually test this?

jyn514 (commented):
This seems ok to me. We've discussed internally that self-test is in flux, but I'm comfortable merging this for 1.90 and main.

I want to note explicitly: this new test only runs when self-check is running on the *host*, when the host has dynamically linked libc. If you are cross-compiling x64 -> aarch64, the test does not run. If you cross-compile aarch64 -> x64, it does not run. If you host-compile on a ferrocene.facade or -musl, it does not run. Only if you are on aarch64-gnu or x86_64-gnu does the test do anything.

jyn514:
lol github duplicated my comment https://github.com/ferrocene/ferrocene/pull/1733#pullrequestreview-3311515345
that was not intentional

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
Hoverbear: I think I understand what you mean now. I'll do this.

jyn514 (approved):
LGTM

jyn514:
bors merge

Hoverbear:
bors merge

(An ssl cert had expired)

Hoverbear:
bors merge

Hoverbear:
bors cancel (seems to be failing)

src/bootstrap/defaults/bootstrap.ferrocene-dist.toml
```diff
@@ -170,6 +170,10 @@ channel = "auto-detect"
 
 
 [target]
+# We default to the self-contained lld linker on Linux GNU targets.
+x86_64-unknown-linux-gnu.default-linker-linux-override = 'self-contained-lld-cc'
+aarch-unknown-linux-gnu.default-linker-linux-override = 'self-contained-lld-cc'
```
Hoverbear, line 175: For some reason this doesn't seem to be sufficient? I'm rebasing and cleaning up the musl issues to isolate this.

Hoverbear:
bors try

jyn514:
latest error is 
```
===> running the self-test tool
    Info: using sysroot /home/ci/project/tmp.Ttx2RLBrnE/sysroot
 Success: binary cargo is valid
 Success: libexec binary rust-analyzer-proc-macro-srv is valid
   Error: default link arg -fuse-ld=lld is missing from target aarch64-unknown-linux-gnu
    Note: the error code is FST_027
```

src/bootstrap/defaults/bootstrap.ferrocene-dist.toml
Xylakant: leaving this from the future: the target name is `aarch64-unknown-linux-gnu`

jyn514:
i *believe* the state of this is that the config was enabled in 7d525905cd3409d20605f3207965447c7ba2c645, but we don't have any tests for it.

Xylakant:
> i _believe_ the state of this is that the config was enabled in [7d52590](https://github.com/ferrocene/ferrocene/commit/7d525905cd3409d20605f3207965447c7ba2c645), but we don't have any tests for it.

I believe rebasing the PR on current main would indeed fix the issues - we did independently figure out the reason for the underlying issue - the target name in the bootstrap.toml was wrong, so the linker config was never set https://github.com/ferrocene/ferrocene/pull/1733#discussion_r3133650825

PR #1734, opened 2025-09-09
Fix lima setup in MacOs section of the "testing other targets" doc

mhatzl:
The lima command to create a VM on MacOS had wrong indentation,
so lima was ignoring it and used the default settings.

This PR also adds a ssh host configuration that works in combination with 1Password as identity agent and agent forwarding for the VM.

The Linux section is moved below MacOS and Windows to make it more obvious that the Linux packages must also be installed in the created VMs on MacOS and Windows. Having the Linux section at the top breaks reading flow for MacOS and Windows, because it required to scroll down to the respective section, then up again to the Linux one, and then down again to the target procedures.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/testing-other-targets.rst
+++ b/ferrocene/doc/internal-procedures/src/testing-other-targets.rst
@@ -23,17 +23,6 @@ Host Setup
 Unless otherwise noted, all bare-metal targets are tested via QEMU on a Linux host.
 On macOS, a tool like Lima or Docker must be used. On Windows, WSL2 must be used.
 
-:target-with-triple:`x86_64-unknown-linux-gnu`
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-You need to have all the normal prerequisites from :doc:`internal-procedures:setup-local-env`
-installed, as well as a few extras:
-
-.. code-block:: bash
-
-   sudo apt install qemu-user-static binfmt-support
-
-
 :target-with-triple:`aarch64-apple-darwin`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -186,6 +175,16 @@ Currently bare metal targets have a similar procedure for testing.
 
    Currently, these targets use our *secret sauce*.
    This will eventually be an open source component, but for now, it's our little bit of arcane magic.
+
+:target-with-triple:`x86_64-unknown-linux-gnu`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You need to have all the normal prerequisites from :doc:`internal-procedures:setup-local-env`
+installed, as well as a few extras:
+
+.. code-block:: bash
+
+   sudo apt install qemu-user-static binfmt-support
 
 :target-with-triple:`aarch64-unknown-none`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ferrocene/doc/internal-procedures/src/testing-other-targets.rst
+++ b/ferrocene/doc/internal-procedures/src/testing-other-targets.rst
@@ -47,7 +47,7 @@ You can create a guest with the following:
 
 .. code-block:: yaml
 
-    cat <<EOF | limactl create --name=ferrocene -
+    cat <<EOF | limactl create --name=test -
     minimumLimaVersion: "1.0.0"
     images:
       - location: "https://cloud-images.ubuntu.com/releases/24.10/release-20241212/ubuntu-24.10-server-cloudimg-arm64.img"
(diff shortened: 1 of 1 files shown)


jonathanpallant:
I tried this and got:

```console
$ limactl shell ferrocene
bash: line 1: cd: /Users/jonathan: No such file or directory
bash: line 1: cd: /Users/jonathan: No such file or directory
```

mhatzl:
> I tried this and got:
> 
> ```
> $ limactl shell ferrocene
> bash: line 1: cd: /Users/jonathan: No such file or directory
> bash: line 1: cd: /Users/jonathan: No such file or directory
> ```

~~I fixed the vm name in my latest commit. was still at "test" before~~

I also got similar warnings, but the VM worked otherwise.
not sure why it shows those errors

jonathanpallant:
I think `qemu-system-arm` is redundant too. You need `qemu-user-static`

mhatzl:
> I think `qemu-system-arm` is redundant too. You need `qemu-user-static`

went through the setup again and it worked without `qemu-system-arm`, so removed it from the docs

mhatzl:
not sure why this got cancelled

Hoverbear:
Sometimes the GHA mirrors of ubuntu repos hang so we have a timeout, it's safe to restart

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
```diff
@@ -23,47 +23,46 @@ Host Setup
...
-        images:
-        - location: "https://cloud-images.ubuntu.com/releases/24.10/release-20241212/ubuntu-24.10-server-cloudimg-arm64.img"
-            arch: "aarch64"
-            digest: "sha256:fb39312ffd2b47b97eaef6ff197912eaa3e0a215eb3eecfbf2a24acd96ee1125"
-        - location: "https://cloud-images.ubuntu.com/releases/24.10/release/ubuntu-24.10-server-cloudimg-arm64.img"
-            arch: "aarch64"
-        mounts:
-        - location: "~"
-        - location: "/tmp/lima"
+    cat <<EOF | limactl create --name=ferrocene -
```
jyn514, line 40:
```suggestion
    mkdir -p /Users/jyn/lima/shared
    cat <<EOF | limactl create --name=ferrocene -
```
otherwise it gives this warning when you run `limactl create`:
```
WARN[0000] field `mounts[1].location` refers to a non-existent directory: "/Users/jyn/lima/shared":
```

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
```diff
@@ -133,6 +166,20 @@ Finally, ensure the guest is configured according to :doc:`internal-procedures:s
...
+You need to have all the normal prerequisites from :doc:`internal-procedures:setup-local-env`
+installed, as well as a few extras:
+
+.. code-block:: bash
+
+   sudo apt install qemu-user-static binfmt-support
+
+.. Note::
+
+    These packages must also be installed in the VMs used on MacOS and Windows.
```
jyn514, line 187: I wonder if we could at some point create a docker image that's just base ubuntu plus these packages. No need to change anything in this PR, though.

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
```diff
@@ -133,6 +166,20 @@ Finally, ensure the guest is configured according to :doc:`internal-procedures:s
...
 
+:target-with-tuple:`x86_64-unknown-linux-gnu`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You need to have all the normal prerequisites from :doc:`internal-procedures:setup-local-env`
+installed, as well as a few extras:
+
+.. code-block:: bash
+
+   sudo apt install qemu-user-static binfmt-support
```
jyn514, line 183: this also needs `zlib1g-dev`, otherwise rustc_driver fails to link.
```suggestion
   sudo apt install qemu-user-static binfmt-support zlib1g-dev
```

jyn514:
LGTM modulo comments, thanks for fixing this :)

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
mhatzl: The comment above refers to the "setup-local-env" section, so I only added the zlib package there.

maybe we should highlight this comment better?

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
mhatzl: that is the "shared" folder mount point defined in the lima settings below.

I could add `mkdir -p ~/lima/shared` and add a comment that this can be changed to another path if wanted and synced with the lima setting?

jyn514: sounds good to me :)

ferrocene/doc/internal-procedures/src/testing-other-targets.rst
jyn514: you are absolutely right, it's listed in https://public-docs.ferrocene.dev/main/qualification/internal-procedures/setup-local-env.html#x86-64-linux-glibc-x86-64-unknown-linux-gnu. i could have sworn i ran that line ... not sure what broke here, no need for changes on your end.

mhatzl: yeah it is a bit easy to miss.
maybe if more people keep missing this, we change it to a bigger note :D

Hoverbear (approved):
bors merge

PR #1757, opened 2025-09-19
[1.88] Mark core cert as pre-release

Hoverbear:
Unfortunately, we aren't confident in our code coverage tooling, so we're gonna spend some more time with it and provide a pre-release of the certification docs instead.


Hoverbear:
bors try

PR #1760, opened 2025-09-22
Downgrades Ubuntu version in lima setup for testing other targets.

BriocheBerlin:


mhatzl:
switching to Ubuntu 24.04 also affects the "setup-local-dev" document, because `awscli` is not available.

I had to use `sudo snap install aws-cli --classic` (see: https://forum.virtualmin.com/t/ubuntu-24-04-does-not-have-awscli/130361)

we should likely adapt the "setup-local-dev" section to refer to Ubuntu 24.04 in this PR as well. Haven't checked if other docs also refer to 24.10.

BriocheBerlin:
> we should likely adapt the "setup-local-dev" section to refer to Ubuntu 24.04 in this PR as well.

It does say 

> Note
> On versions earlier than Ubuntu 24.10 the awscli package is not newer than 2.9.0, and should not be used.

PR #1766, opened 2025-09-23
Always submodule update before signing

Hoverbear:
We caught this bug in https://github.com/ferrocene/ferrocene/pull/1763

### DIFF
--- a/src/bootstrap/src/ferrocene/sign/mod.rs
+++ b/src/bootstrap/src/ferrocene/sign/mod.rs
@@ -42,12 +42,11 @@ impl<S: Step<Output = PathBuf> + IsSphinxBook> Step for SignDocument<S> {
             _ => panic!("only the sign command supports signing documents"),
         };
 
-        // Ensure we user have all submodules updated, otherwise the signatures
+        // Ensure we always have all submodules updated, otherwise the signatures
         // produced locally may be different than what CI produces.
         // While this is not strictly required, it has caught us during release and
         // caused an abundance of stress.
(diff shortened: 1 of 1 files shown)


tshepang:
maybe this should be done by bootstrap, where this is ran before cosign is invoked (which means we don't have to remember to do it, or document it)

src/bootstrap/src/ferrocene/sign/mod.rs
```diff
@@ -42,6 +42,13 @@ impl<S: Step<Output = PathBuf> + IsSphinxBook> Step for SignDocument<S> {
             _ => panic!("only the sign command supports signing documents"),
         };
 
+        // Ensure we user have all submodules updated, otherwise the signatures
```
jyn514:
```suggestion
        // Ensure we always have all submodules updated, otherwise the signatures
```

src/bootstrap/src/ferrocene/sign/mod.rs
```diff
@@ -42,6 +42,13 @@ impl<S: Step<Output = PathBuf> + IsSphinxBook> Step for SignDocument<S> {
...
+        // produced locally may be different than what CI produces.
+        // While this is not strictly required, it has caught us during release and
+        // caused an abundance of stress.
+        builder.require_and_update_all_submodules();
+
```
tshepang, line 50:
```suggestion
```

tshepang (approved):

Hoverbear:
bors merge

PR #1769, opened 2025-09-23
Enable code coverage on core doctests, std

Hoverbear:
This PR enables code coverage measurements on both `core`'s doctests and `std` by ensuring we always build `std`/`core` with the same features when doing code coverage.

This works around a number of issues with duplicated `core`s and missing `profiler_builtins` by ensuring wherever we build `std`/`core` we call `instrument_coverage` if appropriate. It also unifies the changes made around code coverage to all live inside `instrument_coverage`, to ensure it's always the same features.

I expect this code may need a bit of further work, but it's already an improvement on the state of things.

### DIFF
--- a/src/bootstrap/src/core/build_steps/compile.rs
+++ b/src/bootstrap/src/core/build_steps/compile.rs
@@ -30,7 +30,7 @@ use crate::core::config::flags::FerroceneCoverageFor;
 use crate::core::config::{
     CompilerBuiltins, DebuginfoLevel, LlvmLibunwind, RustcLto, TargetSelection,
 };
-use crate::ferrocene::code_coverage::{Paths, instrument_coverage};
+use crate::ferrocene::code_coverage::instrument_coverage;
 use crate::ferrocene::secret_sauce::SecretSauceArtifacts;
 use crate::utils::build_stamp;
 use crate::utils::build_stamp::BuildStamp;
@@ -716,12 +716,6 @@ pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Car
     cargo.rustdocflag(&html_root);
 
     cargo.rustdocflag("-Zcrate-attr=warn(rust_2018_idioms)");
-
-    if builder.config.cmd.ferrocene_coverage_for() == Some(FerroceneCoverageFor::Library) {
-        let paths = Paths::find(builder, target, FerroceneCoverageFor::Library);
-        cargo.rustdocflag(&format!("--persist-doctests={}", paths.doctests_bins_dir.display()));
-        instrument_coverage(builder, cargo);
-    }
 
     // ferrocene addition: `cfg` used to adapt libstd to our "secret sauce" libc
     if target.contains("facade") {
--- a/src/bootstrap/src/core/builder/cargo.rs
+++ b/src/bootstrap/src/core/builder/cargo.rs
@@ -6,7 +6,8 @@ use super::{Builder, Kind};
 use crate::core::build_steps::test;
 use crate::core::build_steps::tool::SourceType;
 use crate::core::config::SplitDebuginfo;
-use crate::core::config::flags::Color;
+use crate::core::config::flags::{Color, FerroceneCoverageFor};
+use crate::ferrocene::code_coverage::{Paths, instrument_coverage};
 use crate::utils::build_stamp;
 use crate::utils::helpers::{self, LldThreads, check_cfg_arg, linker_args, linker_flags};
 use crate::{
@@ -125,6 +126,14 @@ impl Cargo {
             _ => {
                 cargo.configure_linker(builder);
             }
+        }
+
+        if compiler.stage == builder.top_stage
+            && builder.config.cmd.ferrocene_coverage_for() == Some(FerroceneCoverageFor::Library)
+        {
+            let paths = Paths::find(builder, target, FerroceneCoverageFor::Library);
+            cargo.rustdocflag(&format!("--persist-doctests={}", paths.doctests_bins_dir.display()));
+            instrument_coverage(builder, &mut cargo);
         }
 
         cargo


src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -25,6 +25,26 @@ pub(crate) fn instrument_coverage(builder: &Builder<'_>, cargo: &mut Cargo) {
...
+    // placed in the `deps` directory inside of Cargo's target directory. In theory this
+    // would result in Cargo picking it up automatically, but in practice it doesn't.
+    //
+    // Turns out that Cargo passes `-L dependency=$target_dir/deps` to rustc instead of
+    // just `-L $target_dir/deps`. The `dependency=` prefix causes rustc to only load
+    // explicit dependencies from that directory, not implicitly injected crates.
+    //
+    // To fix the problem, we add our own `-L` flag to the Cargo invocation, pointing to
+    // the location of profiler_builtins without the `dependency=` prefix.
+    let compiler = builder.compiler(1, builder.host_target);
```
jyn514:
```suggestion
    let compiler = builder.compiler(builder.top_stage, builder.host_target);
```

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -25,6 +25,26 @@ pub(crate) fn instrument_coverage(builder: &Builder<'_>, cargo: &mut Cargo) {
...
+    let target_dir = builder.cargo_out(compiler, Mode::Std, builder.config.host_target).join("deps");
+    cargo.rustflag(&format!("-L{}", target_dir.to_str().unwrap()));
```
jyn514, line 55: do we need `-L` here? can we use `--extern profiler_builtins=...` directly instead? that requires us to get the hash from somewhere in bootstrap, but I would expect that to be possible.

src/bootstrap/src/ferrocene/code_coverage.rs
Hoverbear: I also tried this. This would create a cycle:

```
Cycle in build detected when adding Assemble { target_compiler: Compiler { stage: 2, host: x86_64-unknown-linux-gnu, forced_compiler: false } }
	Std { target: x86_64-unknown-linux-gnu, build_compiler: Compiler { stage: 1, host: x86_64-unknown-linux-gnu, forced_compiler: false }, crates: [], force_recompile: false, extra_rust_args: [], is_for_mir_opt_tests: false }
	Rustc { target: x86_64-unknown-linux-gnu, build_compiler: Compiler { stage: 1, host: x86_64-unknown-linux-gnu, forced_compiler: false }, crates: [] }
	Assemble { target_compiler: Compiler { stage: 2, host: x86_64-unknown-linux-gnu, forced_compiler: false } }
```

I actually expected `./x.py test --stage 1 --coverage=library library/core -vv --doc` to break with this but it did not. 🤔

src/bootstrap/src/ferrocene/code_coverage.rs
Hoverbear: Initial testing suggests yes! The following also appears to work:

```rust
    // TODO: Hack! Do this properly!
    let mut needle = None;
    for hay in std::fs::read_dir(target_dir).unwrap() {
        let hay = hay.unwrap().path();
        let hay_file = hay.file_name().unwrap().to_str().unwrap();
        if !hay_file.ends_with(".rlib") {
            continue
        }
        if hay_file.starts_with("libprofiler_builtins") {
            needle = Some(hay);
        }
    }
    let Some(needle) = needle else {
        panic!("No libprofiler_builtins found for `--extern libprofiler_builtins");
    };
    println!("NEEDLE {needle:?}");

    cargo.rustflag("--extern");
    cargo.rustflag(&format!("profiler_builtins={}", needle.display()));
``` 

I'm not sure if that's code I'd merge, but it's a proof that this is valid and probably less dangerous.

src/bootstrap/src/ferrocene/code_coverage.rs
jyn514: oh, you are not actually running the compiler, just using it to determine the out-dir. try `Compiler::new(top_stage, host_target)`.

jyn514: actually wait hold on. why is instrument_coverage being called here at all? i would expect the caller to pass in the appropriate stage.

Hoverbear: Okay so this change works if you use stage 1, but if you try with stage 2 it breaks.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -96,7 +124,16 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
         FerroceneCoverageFor::Library => {
             let mut instrumented_binaries = vec![];
             let out_dir = builder.cargo_out(state.compiler, Mode::Std, state.target).join("deps");
```
Hoverbear, line 126: @pvdrz has a fix coming with `builder.doc_tests != DocTests::No` for an error he hit with the directory not existing.

pvdrz:
bors try

pvdrz:
bors try

pvdrz:
bors cancel

pvdrz:
bors try

pvdrz:
bors try

Hoverbear:
bors try-

Hoverbear:
bors try

Hoverbear:
bors try

Hoverbear:
bors try-

Hoverbear:
bors try

pvdrz:
bors try-

Hoverbear:
bors try

pvdrz:
bors try

library/std/src/net/tcp.rs
```diff
@@ -9,6 +9,9 @@
         target_os = "trusty",
     ))
 ))]
+// Ferrocene addition: disabled temporarily as these fail when running the test suite with coverage
+// instrumentation enabled.
+#[cfg(not(test))]
 mod tests;
```
jyn514, line 15: this is fine short-term. long term i would like the following:
- `x run generate-coverage` (or any similar name) so everyone gets a "canonical" report which is what we put in the certification docs. that runs the equivalent of `x test --coverage=library --no-doc library/core` today (possibly with doctests and std after this PR), and omits things that are known to be slow or broken with coverage instrumentation.
- `x test --coverage [... arbitrary args]` which lets you expand coverage reports by adding more paths to the x.py invocation.

i think the current situation is in some sense the worst of both worlds: you have to know exactly how to invoke x.py to get the right report, but we also have hard-coded stuff like this that doesn't allow people to run with additional tests.

src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -3052,7 +3028,8 @@ impl Step for Crate {
             }
 
             measure_coverage(builder, cargo.as_mut(), build_compiler, target, coverage_for);
-            cargo.rustflag("--cfg=ferrocene_coverage");
+            // cargo.rustflag("--cfg=ferrocene_coverage");
+            // cargo.arg("--features=core/ferrocene_inject_profiler_builtins");
```
jyn514:
```suggestion
```

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -736,6 +717,12 @@ pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Car
 
     cargo.rustdocflag("-Zcrate-attr=warn(rust_2018_idioms)");
 
+    if builder.config.cmd.ferrocene_coverage_for() == Some(FerroceneCoverageFor::Library) {
+        let paths = Paths::find(builder, target, FerroceneCoverageFor::Library);
+        cargo.rustdocflag(&format!("--persist-doctests={}", paths.doctests_bins_dir.display()));
+        instrument_coverage(builder, cargo);
+    }
```
jyn514: This is set unconditionally when building core. I would expect it to only be set when we are compiling tests for this stage. I think you can do that with a `kind == Kind::Test && stage == builder.top_stage` check: https://doc.rust-lang.org/nightly/nightly-rustc/bootstrap/core/builder/enum.Kind.html. Unfortunately, you don't have a `kind` available in `std_cargo`, which makes me think this is the wrong place for it. I remember you tried putting it in `builder.cargo` — is it possible to move it back there?

> I would expect it to only be set when we are compiling tests for this stage

just to expand on this more — if you set it unconditionally, and then try and do `x test --stage 2 --coverage=library`, it will take several hours to rebuild the compiler. so we really do want to only set it when we're actually going to run tests on the built core.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -24,7 +24,35 @@ pub(crate) fn instrument_coverage(builder: &Builder<'_>, cargo: &mut Cargo) {
...
+    // To fix the problem, we add our own `-L` flag to the Cargo invocation, pointing to
+    // the location of profiler_builtins without the `dependency=` prefix.
+    let compiler = builder.compiler(
+        // Note that for the standard library, stage 1 is tested when either --stage 1 or
+        // --stage 2 are passed.
+        1,
+        builder.host_target,
+    );
+    let target_dir =
+        builder.cargo_out(compiler, Mode::Std, builder.config.host_target).join("deps");
```
jyn514, line 53: you have changed this from `target` to `host_target`. why that change?

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -24,7 +24,35 @@ pub(crate) fn instrument_coverage(builder: &Builder<'_>, cargo: &mut Cargo) {
...
+    // Turns out that Cargo passes `-L dependency=$target_dir/deps` to rustc instead of
+    // just `-L $target_dir/deps`. The `dependency=` prefix causes rustc to only load
+    // explicit dependencies from that directory, not implicitly injected crates.
+    //
+    // To fix the problem, we add our own `-L` flag to the Cargo invocation, pointing to
+    // the location of profiler_builtins without the `dependency=` prefix.
+    let compiler = builder.compiler(
+        // Note that for the standard library, stage 1 is tested when either --stage 1 or
+        // --stage 2 are passed.
+        1,
```
jyn514, line 49: This really does not seem correct to me. Is this to avoid the hour-long recompiles I mentioned above? I would rather fix those directly than hard-code stages.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -96,7 +124,23 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
         FerroceneCoverageFor::Library => {
             let mut instrumented_binaries = vec![];
             let out_dir = builder.cargo_out(state.compiler, Mode::Std, state.target).join("deps");
-            for res in std::fs::read_dir(out_dir).expect("cannot read deps directory") {
+
+            let res_doctest_bins = std::fs::read_dir(&paths.doctests_bins_dir);
+
+            if builder.doc_tests != DocTests::No && res_doctest_bins.is_err() {
+                panic!("cannot read doctests bins directory")
```
jyn514: put the original error here please
```suggestion
                t!(res_doctest_bins, "cannot read doctests bins directory")
```

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -96,7 +124,23 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
...
+            }
+
+            let doctests_bins = res_doctest_bins.ok().into_iter().flat_map(|read_dir| {
```
jyn514: you append these bins even if `Doctests::No` is set. i worry this will run into caching bugs. please explicitly do an `if Doctests { read_dir() } else { vec![] }` rather than the after-the-fact error checking.

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -96,7 +124,23 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
...
+                read_dir.flat_map(|res| {
+                    let path = res.expect("cannot inspect doctest bin directory").path();
```
jyn514: ditto for the other `expect`s here
```suggestion
                    let path = t!(res, "cannot inspect doctest bin directory").path();
```

library/std/src/net/tcp.rs
pvdrz: yeah I'd also like to open an issue so we can investigate why only these tests fail when running with coverage instrumentation enabled.

src/bootstrap/src/core/build_steps/compile.rs
pvdrz: I tried to put it inside `instrument_coverage` I think :thinking:. I remember I got an issue about the flag being added twice. I'll investigate

src/bootstrap/src/ferrocene/code_coverage.rs
pvdrz: cc @Hoverbear

src/bootstrap/src/ferrocene/code_coverage.rs
pvdrz: cc @Hoverbear

library/std/src/net/tcp.rs
Hoverbear: I'm worried about forgetting this or missing a bug. :( Can we only ignore during coverage somehow?

src/bootstrap/src/ferrocene/code_coverage.rs
Hoverbear: I do not know.

src/bootstrap/src/ferrocene/code_coverage.rs
Hoverbear: This is a comment that exists elsewhere in the codebase as well. Setting it to `top_stage` results in cycles on stage 2.

src/bootstrap/src/core/build_steps/compile.rs
Hoverbear: I am fairly sure rustdoc needs to be using the covered libcore in order to be able to import it without getting a duplicate error.

library/std/src/net/tcp.rs
pvdrz: done

src/bootstrap/src/core/build_steps/compile.rs
pvdrz: oh wait this is about the whole block and not only the rustdocflag, isn't it?

pvdrz: Hmmm... I tried to put this on `Cargo::new` but it fails due to coverage data missing for the `sync` crate:
```patch
diff --git a/src/bootstrap/src/core/build_steps/compile.rs b/src/bootstrap/src/core/build_steps/compile.rs
index 49f6680f7a8..dfff0a9d78f 100644
--- a/src/bootstrap/src/core/build_steps/compile.rs
+++ b/src/bootstrap/src/core/build_steps/compile.rs
@@ -30,7 +30,7 @@
 use crate::core::config::{
     CompilerBuiltins, DebuginfoLevel, LlvmLibunwind, RustcLto, TargetSelection,
 };
-use crate::ferrocene::code_coverage::{Paths, instrument_coverage};
+use crate::ferrocene::code_coverage::instrument_coverage;
 use crate::ferrocene::secret_sauce::SecretSauceArtifacts;
 use crate::utils::build_stamp;
 use crate::utils::build_stamp::BuildStamp;
@@ -717,12 +717,6 @@ pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Car
 
     cargo.rustdocflag("-Zcrate-attr=warn(rust_2018_idioms)");
 
-    if builder.config.cmd.ferrocene_coverage_for() == Some(FerroceneCoverageFor::Library) {
-        let paths = Paths::find(builder, target, FerroceneCoverageFor::Library);
-        cargo.rustdocflag(&format!("--persist-doctests={}", paths.doctests_bins_dir.display()));
-        instrument_coverage(builder, cargo);
-    }
-
     // ferrocene addition: `cfg` used to adapt libstd to our "secret sauce" libc
     if target.contains("facade") {
         match &*target.triple {
diff --git a/src/bootstrap/src/core/builder/cargo.rs b/src/bootstrap/src/core/builder/cargo.rs
index a1209b42d02..9c63e6e9512 100644
--- a/src/bootstrap/src/core/builder/cargo.rs
+++ b/src/bootstrap/src/core/builder/cargo.rs
@@ -6,7 +6,8 @@
 use crate::core::build_steps::test;
 use crate::core::build_steps::tool::SourceType;
 use crate::core::config::SplitDebuginfo;
-use crate::core::config::flags::Color;
+use crate::core::config::flags::{Color, FerroceneCoverageFor};
+use crate::ferrocene::code_coverage::{instrument_coverage, Paths};
 use crate::utils::build_stamp;
 use crate::utils::helpers::{self, LldThreads, check_cfg_arg, linker_args, linker_flags};
 use crate::{
@@ -127,6 +128,12 @@ pub fn new(
             }
         }
 
+        if cmd_kind == Kind::Test && compiler.stage == builder.top_stage && builder.config.cmd.ferrocene_coverage_for() == Some(FerroceneCoverageFor::Library) {
+            let paths = Paths::find(builder, target, FerroceneCoverageFor::Library);
+            cargo.rustdocflag(&format!("--persist-doctests={}", paths.doctests_bins_dir.display()));
+            instrument_coverage(builder, &mut cargo);
+        }
+
         cargo
     }
 
```

This is the error:

```console
Merging together code coverage measurements
Generating lcov dump of the code coverage measurements
error: failed to load coverage: '/home/christian/Workspace/ferrous-systems/ferrocene/ferrocene/build/x86_64-unknown-linux-gnu/stage2-std/x86_64-unknown-linux-gnu/release/deps/sync-9f93eb890811e624': no coverage data found
error: could not load coverage information
```

src/bootstrap/src/ferrocene/code_coverage.rs
pvdrz: wait, did I do this?

jyn514: if we're not sure, let's revert it back.

src/bootstrap/src/core/build_steps/compile.rs
jyn514: yes ok, so I think what's happening is that we're no longer unconditionally building all binaries with instrumentation, so the scripts that try and load it need to be smarter. there's an existing TODO about this you wrote in `bootstrap::ferrocene::code_coverage::generate_coverage_report`.

take a look at `stream_cargo` -- it has a list of the binaries that get built (you may have to modify it to actually pass in that list).

I expect this to be at least a couple hours of work, maybe more, so I'm ok if this isn't in the initial PR that lands, but in that case please open a follow-up ticket.

jyn514: oh, but I think you can remove `Kind == Test` and only gate on the stage and that still gets us 80% of the benefit for 20% of the work.

src/bootstrap/src/ferrocene/code_coverage.rs
pvdrz: Ok, git blames me because I formatted the line. The original change was introduced in [the first commit of this PR](https://github.com/ferrocene/ferrocene/pull/1769/commits/b64252ef7f715b3aebc2f784612e75ce0d46fb90#diff-e43e82037f7cd7b78019b520e58e109a06fbddaf6a6d42ae185e608a9fce8510R46).

The thing is that this section of code was moved from `impl Step for Std` (`Std` has a `target` field) to the `instrument_coverage` function which doesn't have an `Std` value in scope afaik.

src/bootstrap/src/core/build_steps/compile.rs
pvdrz: ok nice, done in d9e3f89730d74c9f4899ba585764a74f4e0d67b1

PR #1783, opened 2025-09-29
Disable rustdoc-js-std tests in compiletest job

Hoverbear:
These tests are very slow and we frequently see them failing in pulls and CI for inscrutable reasons. Since `rustdoc` is not qualified, it's OK to not run these.

### DIFF
--- a/ferrocene/ci/split-tasks.py
+++ b/ferrocene/ci/split-tasks.py
@@ -117,6 +117,9 @@ JOBS_DEFINITION: JobsDefinition = {
                 # `rustdoc-gui` requires nodejs, npm and chromium to be installed
                 # in the container, which is currently not the case.
                 "rustdoc-gui",
+                # `rustdoc-js-std` tests are frequently unstable and this is not part of our
+                # qualified tooling.
+                "rustdoc-js-std",
             ],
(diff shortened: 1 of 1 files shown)


ferrocene/ci/split-tasks.py
```diff
@@ -117,6 +117,9 @@ def __lt__(self, other):
                 # `rustdoc-gui` requires nodejs, npm and chromium to be installed
                 # in the container, which is currently not the case.
                 "rustdoc-gui",
+                # `rustdoc-js-std` tests are frequently unstable and this is not part of our
+                # qualified tooling.
+                "rustdoc-js-std",
```
jyn514, line 122: do you want to disable `rustdoc-js` also? we could probably get away with disabling all the rustdoc tests tbh but -js and -js-std are the ones that fail the most often.

Hoverbear: `rustdoc-js` failures generally indicate a stringdex problem when they fail, and are desirable.

jyn514 (approved):

Hoverbear:
bors merge

jyn514:
bors merge

PR #1784, opened 2025-09-29
do not run some of the snapshot tests on non-linux

tshepang:


src/bootstrap/src/core/builder/tests.rs
```diff
@@ -1106,16 +1108,16 @@ mod snapshot {
     }
 
     #[test]
+    // Avoids a fail due to not having this line in the snapshot of non-amd64 linux:
+    // [doc] rustc 2 <host> -> std 2 <x86_64-unknown-ferrocene.certified> crates=[core]
+    #[cfg(target_os = "linux")]
```
jyn514: oh, i missed that these tests still run on some platforms. in that case i'm happy to merge. but this cfg is wrong: the thing you care about is a x86-64 host, not whether it's linux or not. 
```suggestion
    #[cfg(target_arch = "x86_64")]
```
(ditto for other cfgs below)

src/bootstrap/src/core/builder/tests.rs
```diff
@@ -1230,9 +1235,6 @@ mod snapshot {
                 "--set",
                 "rust.lld=true",
             ])
-                // Ferrocene annotation: Our snapshot differs from upstream, as our process
-                // is somewhat different.
-                // `--bless` does not fix these. You may need to manually update this shapsnot.
```
jyn514, line 1235: ~~let's keep this comment, since we're still running the test.~~ never mind, I see that you moved it to the top of the test block. You removed the comment about `--bless` - does bless work on these snapshots now?

src/bootstrap/src/core/builder/tests.rs
tshepang: oh, indeed... didn't think that far

src/bootstrap/src/core/builder/tests.rs
tshepang: the comment at the top (by upstream) does point to how to bless... it points to the readme, which show what command to run

Hoverbear:
> Do we have those tests?

Yes, you can take a look at the `self-test` tool/jobs. We check the compiler works. :)

jyn514:
r=me with https://github.com/ferrocene/ferrocene/pull/1784/files#r2391819036 addressed

tshepang:
bors try

Hoverbear (approved):

Hoverbear:
bors merge

PR #1785, opened 2025-09-29
use separate workspace for custom Ferrocene tools

tshepang:


tshepang:
bors try

tshepang:
bors try

tshepang:
bors try-

tshepang:
bors retry

tshepang:
bors try

tshepang:
bors retry

tshepang:
bors try

tshepang:
that was flaky ci... worked when I re-ran the job

ferrocene/tools/blanket/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blanket"
-edition = "2024"
+edition.workspace = true
```
jyn514, line 3: oh this is extremely cute, i love that we can do this now :) a future PR could extend this to the dependencies too

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -315,6 +315,19 @@ for prefix in "${DIRECTORIES_CONTAINING_LOCKFILES[@]}"; do
...
+if [ "${upstream_branch}" == "master" ]; then
+    prefix="ferrocene/tools"
+    lock="${prefix}/Cargo.lock"
+    manifest="${prefix}/Cargo.toml"
+    echo "pull-upstream: ensure ${lock} has latest semver-compatible crates"
+    cargo update --manifest-path "${manifest}"
+    if git status --porcelain=v1 | grep "^ M ${lock}$" >/dev/null; then
+        git add "${lock}"
+        git commit -m "update ${lock} to latest semver-compatible crates"
+    fi
```
jyn514: nit:
```suggestion
    commit_if_modified "${lock}" "update ${lock} to latest semver-compatible crates"
```

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -315,6 +315,19 @@ for prefix in "${DIRECTORIES_CONTAINING_LOCKFILES[@]}"; do
     commit_if_modified "$lock" "update ${lock} to match ${manifest}"
 done
 
+# We keep lockfile for Ferrocene tools fresh
```
jyn514, line 316: There is a comment on line 224 that reads:
```
    # There could be conflicts between our Cargo.lock and upstream's, as we
    # have our own crates with our own dependencies in the workspace.
    # Automatically resolve any conflict involving Cargo.lock to prefer our own
    # copy of the lockfile rather than upstream's.
```
I believe that whole block of code can be removed now?

There is a separate comment on line 297 that reads:
```
# Occasionally, changes made upstream require the lockfile to be regenerated,
# otherwise CI with its --locked flag will fail. This is **not** updating the
# versions of the packages we use, but ensuring the lockfile stays consistent.
```
Do you know if it's possible to remove that now? I think it should be, now that we don't share upstream's cargo.lock.

src/bootstrap/src/utils/proc_macro_deps.rs
jyn514, line 1: noting that I manually verified this matches the upstream version 👍 all the changes here are reverting previous ferrocene edits to the file.

src/tools/tidy/src/deps.rs
jyn514, line 1: noting that i manually verified this new file matches upstream 👍

Cargo.lock
jyn514, line 1: noting that I manually verified this matches upstream (except that we missed a bump to annotate-snippets, which I think is landing in the next upstream pull and so doesn't require changes here).

tshepang: you probably checked against a different upstream checkout than the one this pull is based on...

ferrocene/tools/pull-upstream/pull.sh
tshepang: > Do you know if it's possible to remove that now? I think it should be, now that we don't share upstream's cargo.lock.

was thinking that this only applies to main lockfile, assuming that we have custom changes to bootstrap manifest, but it so happens that is not the case, so I made 21b6567fc2c1caf8bf586d6d77a817f382b7ff45 and bc44faa05ac0d4dd72dbca76d15c23d9f0646906

jyn514: Hm, so my one concern is that if we add things to build-helper, they will end up in a bunch of different workspaces -- this is going to happen in https://github.com/ferrocene/ferrocene/pull/1872, for example, we'll add the `similar` crate. So I would prefer to only remove this for the top-level Cargo.lock, not for bootstrap.

Actually I see that the top-level Cargo.lock is used by build-helper indirectly through `collect-license-metadata` :/ So I am not sure we can remove either of these. Sorry to mislead you.

We might be able to convince upstream to move license-metadata to the bootstrap workspace, in which case we could go back to only merging bootstrap's workspace, not the top-level workspace.

tshepang: fascinating that the moment I realize we don't touch manifest files is around the moment we do... reverted

tshepang:
bors try

tshepang:
flaky ci...
bors retry

tshepang:
flaky ci...
bors retry

tshepang:
lockfiles got stale, so had to rebase, and then also made them match those of upstream

tshepang:
bors try

jyn514:
```
+ choco install -y zstandard --version=1.5.6
Chocolatey v2.5.1
Installing the following packages:
zstandard
By installing, you accept licenses for the packages.

Too long with no output (exceeded 10m0s): context deadline exceeded
```
well that's concerning. i'll open an issue.

PR #1787, opened 2025-09-30
Custom core coverage aggregator (codename: `blanket`)

jyn514:
Summary:

We know which symbols we care about covering (those in https://public-docs.ferrocene.dev/main/certification/core/subset.html), and we want to ensure we cover exactly those rather than "whichever functions make it into the binary past LTO".

1. Write a rustc_driver which collects a list of those fully-qualified symbols. Emit them as JSON to `build/host/stage1-std/${certified-target}/release/symbol-report.json`.
2. Rather than using llvm-cov to turn our .profdata files into .info, use our new `blanket` tool to generate an in-memory structured data aggregation.
3. Rather than using grcov to display the aggregation, use the HTML reports in our new `blanket` tool to generate `build/${certified-target}/doc/certified-coverage-report.html.

This also fixes the main original bug, which was that we were filtering out `libstd.so` when aggregating coverage reports. That is not correct; without that .so the coverage tool doesn't know how to map the binary info back to source info, and reports all the sources as uncovered. Before, using llvm-cov, it would also be reported as "ignored". Now, using blanket, it's correctly reported as uncovered.

Doing things in this way allows us to customize the reporting as well as to add integration tests that we actually have all the coverage we expect.

This PR does not currently modify `ferrocene/doc/core-certification/src/spreadsheets/subset.tsv`. Long term, I would like to delete that file, along with `ferrocene/tools/coverage-of-subset`, and use the generated HTML report in `blanket` for everything.

---

Additionally, this extends bootstrap to support running std's tests and doctests with `-C instrument-coverage`. Unfortuately, `blanket` tends to crash while parsing the .profdata for these because of an upstream bug in `llvm-profparse`. I am going to investigate why that happens in the next few weeks. In the meantime, if you see "Parsing Failed" while running `blanket`, try again but with fewer tests.

Doctests work in the following way:
- Pass `--persist-doctest-binaries` to rustdoc for std tests whenever bootstrap is passed `--coverage=library`
- Load those binaries when generating coverage, so that blanket can parse their DWARF info

Std tests are mostly the same as before, except that passing `-L deps` for profiler_builtins has moved to `instrument_coverage` so that we're passing it consistently instead of only for compile::Std.

---

Some things to note while reviewing:
- This includes all commits from https://github.com/ferrocene/blanket/. Those commits were pulled in as a subtree, and the private repo will be deleted shortly.
- This includes two bugfixes for pre-existing issues in bootstrap, which have upstream PRs: https://github.com/rust-lang/rust/pull/147168 and https://github.com/rust-lang/rust/pull/147167
- Upstream has no way to get a fully-qualified crate name, and when [I asked about upstreaming it](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/print.20a.20fully.20qualified.20path.20name.3F/with/541561323) they didn't seem interested. So https://github.com/ferrocene/ferrocene/commit/7bd83400c45a4420bb817c2f9e7a020f2bca821e forks the compiler to add a way to do that. I don't *believe* this affects our spec, but Ana may know more.
- In #1769 Ana and Christian added various changes for doctests and std. https://github.com/ferrocene/ferrocene/commit/ed186f8f9557cb953b3ff197bc5744840b94a320 adds to those changes so that we can cross-compile std. That commit is not strictly part of symbol-report, but symbol-report needs it in order to not break.
- As part of a previous approach, I tried use rustdoc-json in `symbol-report`. I ripped that out because using a rustc_driver was more reliable. But the changes to make it possible to run `x doc certified-api-docs --json` are still present in 265c76c159e0ef3dda8b1f41cdd6faea79ea0165, because that seems useful. I can rip them out if desired.

We caught some pre-existing bugs while working on this:
- `-C instrument-coverage` breaks most of the TCP and UDP tests, because they can't bind to the socket. It's unclear why this happens. We'll investigate this later.
- We pass `-L deps/` to load profiler_builtins. This can cause caching bugs or incorrect behavior if there are multiple versions of profiler_builtins in the deps directory. We plan to switch to `--extern profiler_builtins=/exact/path.rlib` in the future to avoid this.

---

To test this, you can run `x test --coverage=library --no-doc library/core --stage 1`, which should print the path to an HTML file that looks like this:
image

### DIFF
--- a/src/bootstrap/src/ferrocene/code_coverage.rs
+++ b/src/bootstrap/src/ferrocene/code_coverage.rs
@@ -164,17 +164,33 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
         target: builder.config.host_target,
     });
 
-    builder.ensure(CoverageReport {
+    let html_report = builder.ensure(CoverageReport {
         certified_target: builder.config.host_target.certified_equivalent().unwrap(),
         profdata: paths.profdata_file,
         instrumented_binaries,
         symbol_report,
     });
 
+    let dist_report = builder
+        .out
+        .join("ferrocene")
+        .join("coverage")
+        .join(html_report.file_name().expect("No coverage report filename determined."));
+    builder.info(&format!("Saving coverage report to {}", dist_report.display()));
+    builder.copy_link(&html_report, &dist_report, crate::FileType::Regular);
+
     if builder.doc_tests != DocTests::No {
         // Remove the doctest binaries so they're not distributed afterwards.
         t!(std::fs::remove_dir_all(&paths.doctests_bins_dir));
     }
+}
+
+fn coverage_dir(builder: &Builder<'_>, t: TargetSelection) -> PathBuf {
+    builder.doc_out(t).join("coverage")
+}
+
+pub(super) fn coverage_file(builder: &Builder<'_>, t: TargetSelection) -> PathBuf {
+    coverage_dir(builder, t).join("certified-coverage-report.html")
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -244,7 +260,8 @@ impl Step for CoverageOutcomesDir {
                 Some(download_and_extract_ci_outcomes(builder, "coverage"))
             }
             FerroceneCoverageOutcomes::Local => {
-                Some(builder.out.join("ferrocene").join("coverage"))
+                let certified_target = builder.host_target.certified_equivalent().unwrap();
+                Some(coverage_dir(builder, certified_target))
             }
             FerroceneCoverageOutcomes::Custom(path) => Some(path.clone()),
         }
--- a/src/bootstrap/src/ferrocene/run.rs
+++ b/src/bootstrap/src/ferrocene/run.rs
@@ -9,6 +9,7 @@ use crate::builder::{Builder, Cargo, RunConfig, ShouldRun, Step, crate_descripti
 use crate::core::build_steps::compile::{run_cargo, std_cargo};
 use crate::core::build_steps::tool::{SourceType, Tool};
 use crate::core::config::{FerroceneTraceabilityMatrixMode, TargetSelection};
+use crate::ferrocene::code_coverage::coverage_file;
 use crate::ferrocene::doc::{Specification, SphinxMode, UserManual};
 use crate::ferrocene::test_outcomes::TestOutcomesDir;
 use crate::ferrocene::tool::{Blanket, SYMBOL_PATH, SymbolReport};
@@ -219,8 +220,7 @@ impl Step for CoverageReport {
 
     fn run(self, builder: &Builder<'_>) -> Self::Output {
         let src = to_str(&builder.src);
-        let html_out =
-            builder.doc_out(self.certified_target).join("certified-coverage-report.html");
+        let html_out = coverage_file(&builder, self.certified_target);
         let sha_buf;
         let sha = match builder.rust_info() {
             GitInfo::Absent => panic!(
--- a/ferrocene/doc/core-certification/src/code-coverage.rst
+++ b/ferrocene/doc/core-certification/src/code-coverage.rst
@@ -4,4 +4,4 @@
 Code coverage report
 ====================
 
-`Code coverage report <../../coverage/library-x86_64-unknown-linux-gnu/index.html>`_  of the certified core library.
+`Code coverage report <../../coverage/index.html>`_  of the certified core library.
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -15,6 +15,7 @@ use crate::core::build_steps::run::GenerateCopyright;
 use crate::core::config::TargetSelection;
 use crate::ferrocene::code_coverage::CoverageOutcomesDir;
 use crate::ferrocene::doc::certified_api_docs::CertifiedApiDocs;
+use crate::ferrocene::doc::code_coverage::AllCoverageReports;
 use crate::ferrocene::doc::ensure_all_xml_doctrees;
 use crate::ferrocene::test_outcomes::TestOutcomesDir;
 use crate::ferrocene::uv_command;
@@ -52,6 +53,7 @@ impl Step for Docs {
         // certified API docs in the doc_out
         if self.target.certified_equivalent().is_some() {
             builder.ensure(CertifiedApiDocs { target: self.target });
+            builder.ensure(AllCoverageReports { target: self.target });
         } else {
             builder.info(&format!("skipping Build certified-core-docs ({})", self.target));
         }
--- a/src/bootstrap/src/ferrocene/doc/code_coverage.rs
+++ b/src/bootstrap/src/ferrocene/doc/code_coverage.rs
@@ -1,13 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: The Ferrocene Developers
 
-use std::error::Error;
-use std::path::PathBuf;
-
-use build_helper::exit;
-use serde_derive::{Deserialize, Serialize};
-
-use crate::core::build_steps::tool::Tool;
+use crate::FileType;
 use crate::core::builder::{Builder, RunConfig, ShouldRun, Step};
 use crate::core::config::{FerroceneCoverageOutcomes, TargetSelection};
 use crate::ferrocene::code_coverage::CoverageOutcomesDir;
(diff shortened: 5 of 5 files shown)


jyn514:
bors try

tshepang:
this is much too much history... I woud squash most of it, remove the extra license file, and rid of the merge commit (hm, is it a merge commit)

a follow-up can replace (deprecated) structopt with clap

ferrocene/tools/symbol-report/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "symbol-report"
+version = "0.1.0"
```
tshepang: would only have this if it was a released tool
```suggestion
```

ferrocene/tools/blanket/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "blanket"
+version = "0.1.0"
```
tshepang: would only have this if it was a released tool
```suggestion
```

src/tools/collect-license-metadata/src/main.rs
tshepang, line 1: just a note that this an upstream tool, so we should not forget to upstream the changes

Hoverbear: Likely worth upstreaming this

jyn514:
https://github.com/ferrocene/ferrocene/pull/1802

ferrocene/tools/blanket/assets/html_report.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+if (document.readyState === "loading") {
+  // Loading hasn't finished yet
+  document.addEventListener("DOMContentLoaded", main);
+} else {
+  // `DOMContentLoaded` has already fired
+  main();
+}
```
Hoverbear:
```suggestion
}

```

This is also needed in the nearby `css`. :(

ferrocene/tools/symbol-report/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+            let end = lines.lines.last().unwrap().line_index;
+            symbols.push(Function(qualified_name, filename, start + 1, end + 1));
+        }
+        serde_json::to_writer(out, &Symbols(symbols)).expect("failed to serialize symbols");
+        Compilation::Stop
+    }
+}
+
+fn main() {
+    rustc_driver::install_ice_hook("https://github.com/ferrocene/comprehensive/issues/new", |_| ());
```
Hoverbear:
```suggestion
    rustc_driver::install_ice_hook("https://github.com/ferrocene/ferrocene/issues/new", |_| ());
```

src/bootstrap/src/core/build_steps/check.rs
```diff
@@ -83,16 +83,12 @@ impl Step for Std {
             Kind::Check,
         );
 
-        std_cargo(builder, target, &mut cargo);
+        std_cargo(builder, target, &mut cargo, &self.crates);
```
Hoverbear, line 86: We should annotate this as it deviates from upstream.
```suggestion
        // Ferrocene annotation: We modified `std_cargo` to take a list of crates
        // instead of having a `for krate in &*self.crates {}` loop.
        std_cargo(builder, target, &mut cargo, &self.crates);
```

src/bootstrap/src/core/build_steps/check.rs
```diff
@@ -135,14 +131,7 @@ impl Step for Std {
...
 
-        std_cargo(builder, target, &mut cargo);
-
-        // Explicitly pass -p for all dependencies krates -- this will force cargo
-        // to also check the tests/benches/examples for these crates, rather
-        // than just the leaf crate.
-        for krate in &*self.crates {
-            cargo.arg("-p").arg(krate);
-        }
+        std_cargo(builder, target, &mut cargo, &self.crates);
```
Hoverbear, line 134: We should annotate this as it deviates from upstream.
```suggestion
        // Ferrocene annotation: We modified `std_cargo` to take a list of crates
        // instead of having a `for krate in &*self.crates {}` loop.
        std_cargo(builder, target, &mut cargo, &self.crates);
```

src/bootstrap/src/core/build_steps/clippy.rs
```diff
@@ -195,11 +195,7 @@ impl Step for Std {
             Kind::Clippy,
         );
 
-        std_cargo(builder, target, &mut cargo);
-
-        for krate in &*self.crates {
-            cargo.arg("-p").arg(krate);
-        }
+        std_cargo(builder, target, &mut cargo, &self.crates);
```
Hoverbear, line 198: We should annotate this as it deviates from upstream.
```suggestion
        // Ferrocene annotation: We modified `std_cargo` to take a list of crates
        // instead of having a `for krate in &*self.crates {}` loop.
        std_cargo(builder, target, &mut cargo, &self.crates);
```

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -269,37 +269,18 @@ impl Step for Std {
                 target,
                 Kind::Build,
             );
-            std_cargo(builder, target, &mut cargo);
-            for krate in &*self.crates {
-                cargo.arg("-p").arg(krate);
-            }
+            std_cargo(builder, target, &mut cargo, &self.crates);
```
Hoverbear, line 272: We should annotate this as it deviates from upstream.
```suggestion
        // Ferrocene annotation: We modified `std_cargo` to take a list of crates
        // instead of having a `for krate in &*self.crates {}` loop.
        std_cargo(builder, target, &mut cargo, &self.crates);
```

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -541,7 +522,12 @@ fn compiler_rt_for_profiler(builder: &Builder<'_>) -> PathBuf {
 
 /// Configure cargo to compile the standard library, adding appropriate env vars
 /// and such.
-pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Cargo) {
+pub fn std_cargo(
+    builder: &Builder<'_>,
+    target: TargetSelection,
+    cargo: &mut Cargo,
+    crates: &[String],
```
Hoverbear, line 529: We should annotate this as it deviates from upstream.

```suggestion
    // Ferrocene annotation: We take a list of crates which are used in a `for krate in crates {}` loop below
    crates: &[String],
```

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -658,6 +644,13 @@ pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Car
         cargo.env("CFG_DISABLE_UNSTABLE_FEATURES", "1");
     }
 
+    // Explicitly pass -p for all dependencies crates -- this will force cargo
+    // to also check the tests/benches/examples for these crates, rather
+    // than just the leaf crate.
```
Hoverbear, line 654:
```suggestion
    // than just the leaf crate.
    // Ferrocene annotation: See `crates` argument
```

src/bootstrap/src/core/build_steps/compile.rs
```diff
@@ -670,8 +663,10 @@ pub fn std_cargo(builder: &Builder<'_>, target: TargetSelection, cargo: &mut Car
             features += " ferrocene_certified";
         }
         // for no-std targets we only compile a few no_std crates
+        if crates.is_empty() {
```
Hoverbear, line 671: We should annotate this as it deviates from upstream.

src/bootstrap/src/core/build_steps/doc.rs
```diff
@@ -797,7 +797,7 @@ fn doc_std(
         Kind::Doc,
     );
 
-    compile::std_cargo(builder, target, &mut cargo);
+    compile::std_cargo(builder, target, &mut cargo, &requested_crates);
```
Hoverbear, line 804: We should annotate this as it deviates from upstream.

```suggestion
    // Ferrocene annotation: We modified `std_cargo` to take a list of crates
    // instead of having a `for krate in requested_crates {}` loop below.
    compile::std_cargo(builder, target, &mut cargo, &requested_crates);
```

src/bootstrap/src/ferrocene/code_coverage.rs
```diff
@@ -92,84 +120,61 @@ pub(crate) fn generate_coverage_report(builder: &Builder<'_>) {
     //
     // A possible improvement would be to capture `cargo test` stderr and fetch the path of every
     // binary that cargo ran or get the build plan and fetch the paths of the binaries from there.
+    #[allow(unused_variables)]
```
Hoverbear: This might be easily removable.

src/bootstrap/src/ferrocene/run.rs
```diff
@@ -111,3 +115,159 @@ impl Step for TraceabilityMatrix {
...
+            builder,
+            build_compiler,
+            Mode::Std,
+            SourceType::InTree,
+            certified_target,
+            Kind::Check,
+        );
+        let crates = vec!["core".to_owned()]; // currently, only core is certified
+        for krate in &*crates {
+            cargo.arg("-p").arg(krate);
```
Hoverbear: Isn't this already done in `std_cargo` now?

src/bootstrap/src/utils/cc_detect.rs
```diff
@@ -75,7 +75,7 @@ pub fn fill_compilers(build: &mut Build) {
 
         _ => {
             // Ferrocene addition: Load the matching certified target
-            let certified_target = build.host_target.certified_equivalent();
+            // let certified_target = build.host_target.certified_equivalent();
```
Hoverbear: Let's remove these dead comments. :)

src/bootstrap/src/utils/cc_detect.rs
```diff
@@ -86,7 +86,8 @@ pub fn fill_compilers(build: &mut Build) {
                 .cloned()
                 .chain(iter::once(build.host_target))
                 // Ferrocene addition: see above
-                .chain(certified_target)
+                .flat_map(|t| std::iter::once(t).chain(t.certified_equivalent()))
+                // .chain(certified_target)
```
Hoverbear: Let's remove these dead comments. :)

Hoverbear (commented):
Just some small nits! Looks good and it was such a fun hack. :)

src/tools/collect-license-metadata/src/main.rs
jyn514: https://github.com/rust-lang/rust/pull/147422/

jyn514:
bors try

jyn514:
This latest force-push has quite a large diff, but I didn't make any changes other than modifying the history a lot. I'll address review comments in follow-up commits.

jyn514:
Actually, let me rebase over main to make the diff smaller.

jyn514:
ok, you can look at https://github.com/ferrocene/ferrocene/compare/d5b0330ad7a90cd0f7d562078e21511182651418..c30c8691cf25fd3af11184ea08af08f6affe25c9 to see what changed (mostly a rebase over an upstream-pull).

src/bootstrap/src/core/build_steps/check.rs
jyn514: This is being merged upstream: https://github.com/rust-lang/rust/pull/147168. Do you still want me to add an annotation? Perhaps just a link to that PR?

src/bootstrap/src/ferrocene/run.rs
jyn514: Yes, good catch.

src/bootstrap/src/core/build_steps/check.rs
Hoverbear: Oh I'm a big silly, let's not add one then! I didn't notice!

Hoverbear:
There is a "File not found" error in the coverage job and you have a merge conflict, but otherwise seems good.

jyn514:
@Hoverbear does this error mean anything to you? if I run reuse locally it at least prints a message.
```
   Built reuse==6.0.0
Installed 11 packages in 10ms

Exited with code exit status 1
```

jyn514:
squashed the history a bit more but otherwise no changes. this should be ready for final review.

ferrocene/tools/symbol-report/LICENSE
tshepang, line 1: would remove this as it is redundant

jyn514:
Updated this to rip out https://github.com/ferrocene/ferrocene/pull/1787/commits/8aadaf45e58d205088cf2720d79118c394e77e26.

jyn514:
Updated this to:
- rebase over main
- update Cargo.lock
- remove the `git subtree` history so it's easier to backport

ferrocene/tools/blanket/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "blanket"
+edition = "2024"
+
+[dependencies]
+anyhow = { version = "1.0.100", features = ["backtrace"] }
+llvm_profparser = "0.8.3"
```
tshepang, line 7: for a future pr (or this one)
```suggestion
llvm_profparser = { version = "0.8.3", default-features = false }
```

Hoverbear:
bors merge

jyn514:
bors try

Hoverbear:
bors try

PR #1788, opened 2025-09-30
Extend libcore subset

Urhengulas:


library/core/src/slice/index.rs
```diff
@@ -34,6 +34,7 @@ where
 #[cfg_attr(not(panic = "immediate-abort"), inline(never), cold)]
 #[cfg_attr(panic = "immediate-abort", inline)]
 #[track_caller]
+#[allow(unused_variables)]
```
Hoverbear: Let's only allow the unused with `ferrocene_certified`? Otherwise, we need to at least annotate this.

Urhengulas: https://github.com/ferrocene/ferrocene/pull/1788/commits/01d857946d1fca0cc86b2b5af308aac5fe307c75

library/core/src/any.rs
```diff
@@ -705,8 +715,8 @@ impl dyn Any + Send + Sync {
 ///     std::mem::forget(fake_one_ring);
 /// }
 /// ```
-#[derive(Copy, PartialOrd, Ord)]
-#[derive_const(Clone, Eq)]
+#[cfg_attr(not(feature = "ferrocene_certified"), derive(Copy, PartialOrd, Ord))]
```
Hoverbear, line 721: Copy is part of the subset is it not? Can we maintain it's impl here?

Urhengulas: Unfortunately it is a little bit involved. So unless necessary I would avoid it now.

```
error[E0204]: the trait `marker::Copy` cannot be implemented for this type
   --> library/core/src/any.rs:720:52
    |
720 |   #[cfg_attr(feature = "ferrocene_certified", derive(Copy))]
    |                                                      ^^^^ in this derive macro expansion
...
731 |       pub(crate) data: [*const (); 16 / size_of::<*const ()>()],
    |       --------------------------------------------------------- this field does not implement `marker::Copy`
    |
   ::: library/core/src/marker.rs:480:1
    |
480 | / pub macro Copy($item:item) {
481 | |     /* compiler built-in */
482 | | }
    | |_- in this expansion of `#[derive(Copy)]`

error[E0277]: the trait bound `[*const (); 2]: clone::Clone` is not satisfied
   --> library/core/src/any.rs:731:5
    |
721 |   #[cfg_attr(feature = "ferrocene_certified", derive_const(Clone))]
    |                                                            ----- in this derive macro expansion
...
731 |       pub(crate) data: [*const (); 16 / size_of::<*const ()>()],
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `clone::Clone` is not implemented for `[*const (); 2]`
    |
   ::: library/core/src/clone.rs:258:1
    |
258 | / pub macro Clone($item:item) {
259 | |     /* compiler built-in */
260 | | }
    | |_- in this expansion of `#[derive(Clone)]`
    |
```

Hoverbear (commented):
Great. :) I don't see any changes to https://public-docs.ferrocene.dev/main/certification/core/requirement-review.html, is that going to be included here?

Or do we need to redo that process in some next step? If so, how will you determine which bits need to be reviewed later?

Hoverbear:
bors try

Hoverbear:
bors ping

jyn514 (commented):
I don't see `subset.tsv` modified in this PR, is that intentional? My understanding is that if we extend the cfgs without extending the TSV, we are technically not telling the assessor that the new functions are certified.

Urhengulas:
Updating both requirement reviews and the subset.tsv will be done in a separate PR.

Urhengulas:
> Or do we need to redo that process in some next step? If so, how will you determine which bits need to be reviewed later?

The violations are automatically detected by the libcore-statistics tool

Urhengulas:
> ## try
> 
> Build failed:
> 
>     * full

It succeeded after "re-run from failed".

Hoverbear:
(Note: Please do not merge this quite yet)

Hoverbear:
We're gonna land this after #1787 to ensure easy backporting and good measurement of affects.

Urhengulas:
bors try

Urhengulas:
bors try

Urhengulas:
> ## try
> 
> Build failed:
> 
>     * full

Failed in job [aarch64-darwin-test](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/14111/workflows/675b702a-631a-4cbe-8223-84b676681b50/jobs/97163) in step "Setup Darwin Environment" with:

> ==> Downloading from https://armkeil.blob.core.windows.net/developer/Files/downloads/gnu/14.3.rel1/binrel/arm-gnu-toolchain-14.3.rel1-darwin-arm64-arm-none-eabi.pkg
#########                                                                 12.8%curl: (56) Recv failure: Operation timed out
>
> Error: Download failed on Cask 'gcc-arm-embedded' with message: Download failed: https://developer.arm.com/-/media/Files/downloads/gnu/14.3.rel1/binrel/arm-gnu-toolchain-14.3.rel1-darwin-arm64-arm-none-eabi.pkg

Clicked "Re-run from failed".

Urhengulas:
> Clicked "Re-run from failed".

It succeeded

jyn514:
here is a diff of the functions added in this PR: https://gist.github.com/Urhengulas/060408f5c27dab63f64cd397a0104733

we are working long-term towards generating that diff automatically so it's easier to review.

library/core/src/iter/traits/mod.rs
```diff
@@ -6,9 +6,7 @@ mod double_ended;
 #[cfg(not(feature = "ferrocene_certified"))]
 mod exact_size;
 mod iterator;
-#[cfg(not(feature = "ferrocene_certified"))]
 mod marker;
-#[cfg(not(feature = "ferrocene_certified"))]
 mod unchecked_iterator;
```
pvdrz, line 10: The `UncheckedIterator::next_unchecked` method has a default implementation that is not used by any function inside the subset in this PR.

I suggest we gate that default implementation as well because this trait is not public, meaning that we cannot test it in any way:

```diff
--- a/library/core/src/iter/traits/unchecked_iterator.rs
+++ b/library/core/src/iter/traits/unchecked_iterator.rs
@@ -27,6 +27,9 @@ pub(crate) trait UncheckedIterator: TrustedLen {
     /// point you might want to implement this manually instead.
     #[unstable(feature = "trusted_len_next_unchecked", issue = "37572")]
     #[inline]
+    #[cfg(feature = "ferrocene_certified")]
+    unsafe fn next_unchecked(&mut self) -> Self::Item;
+    #[cfg(not(feature = "ferrocene_certified"))]
     unsafe fn next_unchecked(&mut self) -> Self::Item {
         let opt = self.next();
         // SAFETY: The caller promised that we're not empty, and
```

Urhengulas: https://github.com/ferrocene/ferrocene/pull/1788/commits/203ba085f97db97e229a71b0ab81f72d57a78746

pvdrz:
I think you don't need the new range API either as it is an unstable feature: 

<details>
<summary>Diff</summary>

```diff
diff --git a/library/core/src/lib.rs b/library/core/src/lib.rs
index 1e5dfceeb09..86a03038b97 100644
--- a/library/core/src/lib.rs
+++ b/library/core/src/lib.rs
@@ -350,6 +350,7 @@ pub mod autodiff {
 #[cfg(not(feature = "ferrocene_certified"))]
 pub mod random;
 #[unstable(feature = "new_range_api", issue = "125687")]
+#[cfg(not(feature = "ferrocene_certified"))]
 pub mod range;
 pub mod result;
 pub mod sync;
diff --git a/library/core/src/slice/index.rs b/library/core/src/slice/index.rs
index 3c9753cf70f..b65dedce0dd 100644
--- a/library/core/src/slice/index.rs
+++ b/library/core/src/slice/index.rs
@@ -3,7 +3,10 @@
 use crate::intrinsics::slice_get_unchecked;
 use crate::panic::const_panic;
 use crate::ub_checks::assert_unsafe_precondition;
+#[cfg(not(feature = "ferrocene_certified"))]
 use crate::{ops, range};
+#[cfg(feature = "ferrocene_certified")]
+use crate::ops;
 
 #[stable(feature = "rust1", since = "1.0.0")]
 #[rustc_const_unstable(feature = "const_index", issue = "143775")]
@@ -102,7 +105,10 @@ const fn slice_index_fail(start: usize, end: usize, len: usize) -> ! {
 }
 
 mod private_slice_index {
+    #[cfg(not(feature = "ferrocene_certified"))]
     use super::{ops, range};
+    #[cfg(feature = "ferrocene_certified")]
+    use super::ops;
 
     #[stable(feature = "slice_get_slice", since = "1.28.0")]
     pub trait Sealed {}
@@ -125,12 +131,16 @@ impl Sealed for ops::RangeToInclusive<usize> {}
     impl Sealed for (ops::Bound<usize>, ops::Bound<usize>) {}
 
     #[unstable(feature = "new_range_api", issue = "125687")]
+#[cfg(not(feature = "ferrocene_certified"))]
     impl Sealed for range::Range<usize> {}
     #[unstable(feature = "new_range_api", issue = "125687")]
+#[cfg(not(feature = "ferrocene_certified"))]
     impl Sealed for range::RangeInclusive<usize> {}
     #[unstable(feature = "new_range_api", issue = "125687")]
+#[cfg(not(feature = "ferrocene_certified"))]
     impl Sealed for range::RangeToInclusive<usize> {}
     #[unstable(feature = "new_range_api", issue = "125687")]
+#[cfg(not(feature = "ferrocene_certified"))]
     impl Sealed for range::RangeFrom<usize> {}
 
     impl Sealed for ops::IndexRange {}
@@ -455,6 +465,7 @@ fn index_mut(self, slice: &mut [T]) -> &mut [T] {
 
 #[unstable(feature = "new_range_api", issue = "125687")]
 #[rustc_const_unstable(feature = "const_index", issue = "143775")]
+#[cfg(not(feature = "ferrocene_certified"))]
 unsafe impl<T> const SliceIndex<[T]> for range::Range<usize> {
     type Output = [T];
 
@@ -579,6 +590,7 @@ fn index_mut(self, slice: &mut [T]) -> &mut [T] {
 
 #[unstable(feature = "new_range_api", issue = "125687")]
 #[rustc_const_unstable(feature = "const_index", issue = "143775")]
+#[cfg(not(feature = "ferrocene_certified"))]
 unsafe impl<T> const SliceIndex<[T]> for range::RangeFrom<usize> {
     type Output = [T];
 
@@ -715,6 +727,7 @@ fn index_mut(self, slice: &mut [T]) -> &mut [T] {
 
 #[unstable(feature = "new_range_api", issue = "125687")]
 #[rustc_const_unstable(feature = "const_index", issue = "143775")]
+#[cfg(not(feature = "ferrocene_certified"))]
 unsafe impl<T> const SliceIndex<[T]> for range::RangeInclusive<usize> {
     type Output = [T];
 
@@ -793,6 +806,7 @@ fn index_mut(self, slice: &mut [T]) -> &mut [T] {
 /// The methods `index` and `index_mut` panic if the end of the range is out of bounds.
 #[stable(feature = "inclusive_range", since = "1.26.0")]
 #[rustc_const_unstable(feature = "const_index", issue = "143775")]
+#[cfg(not(feature = "ferrocene_certified"))]
 unsafe impl<T> const SliceIndex<[T]> for range::RangeToInclusive<usize> {
     type Output = [T];
```

</details>

Urhengulas:
@pvdrz said:
> I think you don't need the new range API either as it is an unstable feature:

It was not needed, removed it.

jyn514:
[updated diff](https://gist.github.com/jyn514/d1d3819e97d7eedc4bce60f8325e01a5)

jyn514:
I see that requirements-review.csv was deleted in https://github.com/ferrocene/ferrocene/pull/1810/files#diff-bcee36893fe1d2f8b6c5e0582a8c9597cd5527dc9613f8835186163caacfb68e. What is the plan for that? To use the inline `// Ferrocene annotation`s that Christian adds as he adds more coverage?

library/core/src/mem/manually_drop.rs
```diff
@@ -152,7 +152,11 @@ use crate::ptr;
 /// [`MaybeUninit`]: crate::mem::MaybeUninit
 #[stable(feature = "manually_drop", since = "1.20.0")]
 #[lang = "manually_drop"]
-#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(
+    not(feature = "ferrocene_certified"),
+    derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)
+)]
```
jyn514: let's please keep the same formatting as upstream to reduce merge conflicts. it's ok to use `#[rustfmt::skip]` if you need to.

library/core/src/panic/location.rs
```diff
@@ -39,7 +39,7 @@ use crate::ptr::NonNull;
 /// See [`Location::file`]'s documentation for more discussion.
 #[lang = "panic_location"]
 #[stable(feature = "panic_hooks", since = "1.10.0")]
-#[cfg_attr(feature = "ferrocene_certified", allow(dead_code))]
+#[cfg_attr(feature = "ferrocene_certified", expect(dead_code))]
```
jyn514, line 42: just noting that this is a semantic change. i think it's a good one 👍 but not related to expanding the subset that i can tell.

library/core/src/ptr/alignment.rs
```diff
@@ -1,14 +1,12 @@
 #![allow(clippy::enum_clike_unportable_variant)]
 
+#[cfg(feature = "ferrocene_certified")]
+use crate::mem;
```
jyn514: non-blocking nit: would be nice to move this to line 9 so it's more clear it's just a subset of the upstream import.

library/core/src/panic/panic_info.rs
```diff
@@ -15,7 +15,7 @@ use crate::panicking::PanicFmt;
 #[stable(feature = "panic_hooks", since = "1.10.0")]
 #[cfg_attr(not(feature = "ferrocene_certified"), derive(Debug))]
 pub struct PanicInfo<'a> {
-    #[cfg_attr(feature = "ferrocene_certified", allow(dead_code))]
+    #[cfg_attr(feature = "ferrocene_certified", expect(dead_code))]
```
jyn514, line 18: ditto, noting this seems unrelated to the expansion

library/core/src/time.rs
```diff
@@ -76,7 +80,11 @@ const DAYS_PER_WEEK: u64 = 7;
 /// compatibility, you may wish to format `Duration` objects yourself or use a
 /// crate to do so.
 #[stable(feature = "duration", since = "1.3.0")]
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[cfg_attr(
+    not(feature = "ferrocene_certified"),
+    derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)
+)]
```
jyn514: ditto, let's avoid reformatting upstream's code.

library/core/src/option.rs
```diff
@@ -980,6 +979,21 @@ impl<T> Option<T> {
         }
     }
 
+    /// Ferrocene addition: `msg` is type `&'static str` instead of `&str`.
```
jyn514: it's actually possible to put `cfg`s directly on function arguments: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=22369350a86e04ed53b5b8fdc3dcfb9c

could you do that here please?

library/core/src/panic.rs
```diff
@@ -97,6 +97,17 @@ pub macro unreachable_2015 {
...
+#[doc(hidden)]
+#[unstable(feature = "edition_panic", issue = "none", reason = "use unreachable!() instead")]
+#[allow_internal_unstable(panic_internals)]
+#[rustc_macro_transparency = "semitransparent"]
+#[cfg(feature = "ferrocene_certified")]
+pub macro unreachable_2021 {
+    ($($t:tt)*) => (
+        $crate::panicking::panic("internal error: entered unreachable code")
+    ),
+}
```
jyn514, line 111: why was this new macro necessary?

library/core/src/ptr/alignment.rs
Urhengulas: I'd love this as well! But the formatting is really aggressive and I found no way to keep it separate.

Urhengulas: rustfmt::skip probably does the trick...

Urhengulas:
> I see that requirements-review.csv was deleted in https://github.com/ferrocene/ferrocene/pull/1810/files#diff-bcee36893fe1d2f8b6c5e0582a8c9597cd5527dc9613f8835186163caacfb68e. What is the plan for that? To use the inline `// Ferrocene annotation`s that Christian adds as he adds more coverage?

This seems to be related to https://github.com/ferrocene/ferrocene/pull/1810 and not this PR?

jyn514:
yes, you're right -- i'll comment there.

library/core/src/panic.rs
Urhengulas: Because the second and third match arm pull depend in fmt. We do the same for panic_2021.

Urhengulas: I am just wondering why we only do it for 2021 and not the other editions? 🤔

jyn514: the other editions probably print it as a static string instead of going through fmt

jyn514 (approved):

jyn514:
bors merge

PR #1792, opened 2025-10-02
Add more tests to `coretests`

pvdrz:


library/core/src/intrinsics/mod.rs
```diff
@@ -2785,6 +2788,8 @@ pub const unsafe fn const_deallocate(_ptr: *mut u8, _size: usize, _align: usize)
 #[rustc_nounwind]
 #[rustc_intrinsic]
 #[miri::intrinsic_fallback_is_spec]
+// Ferrocene annotation: This function is only used as part of the `ThinBox` inner implementation.
+// It is also a noop in runtime so we don't cover it currently.
```
Hoverbear:
```suggestion
// It is also a noop in runtime so we can't cover it currently.
```

library/coretests/tests/not_doctests.rs
```diff
@@ -0,0 +1,221 @@
...
+    assert!(None::<&mut i32>.cloned().is_none());
+    assert!(None::<&mut i32>.copied().is_none());
+
+    assert_eq!(Some(Some(0i32)).flatten(), Some(0));
+    assert_eq!(Some(None::<i32>).flatten(), None);
+    assert_eq!(None::<Option<i32>>.flatten(), None);
+}
+
+#[test]
+fn result_methods() {
```
Hoverbear, line 117: When you upstream these are you planning to break them down a bit?

Hoverbear:
Rest looks good, lets get this merged so we can backport. :)

library/coretests/tests/not_doctests.rs
pvdrz: yes, I added these long tests functions because most of the things tested here have doctests. So I'd like to get doctests working first and then check which parts can actually be upstreamed

library/core/src/ops/drop.rs
```diff
@@ -237,5 +237,7 @@ pub const trait Drop {
     /// [`mem::drop`]: drop
     /// [`ptr::drop_in_place`]: crate::ptr::drop_in_place
     #[stable(feature = "rust1", since = "1.0.0")]
+    // Ferrocene annotation: This method runs for basically every value that doesn't implement
+    // `Copy` once the value is no longer used. It is unclear why this line is marked as uncovered.
```
Hoverbear: "We don't know" isn't a good reason. Lets figure out a better one. Is the reason because this has no default implementation?

library/core/src/sync/atomic.rs
```diff
@@ -3966,6 +3966,8 @@ atomic_int_ptr_sized! {
 
 #[inline]
 #[cfg(target_has_atomic)]
+/// Ferrocene annotation: we don't cover this function because its only use is in the deprecated
```
Hoverbear: Don't think this is a good enough reason, `compare_and_swap` is in the subset, no?

library/core/src/clone.rs
```diff
@@ -624,6 +624,8 @@ mod impls {
     #[stable(feature = "rust1", since = "1.0.0")]
     impl<T: PointeeSized> Clone for *const T {
         #[inline(always)]
+        // Ferrocene annotation: This function is thoroughly tested inside the `test_clone`
```
Hoverbear, line 627: Similarly, "We don't know" isn't good enough. :(

library/core/src/sync/atomic.rs
pvdrz: yeah but I think we should remove it. It cannot be used by the users anyway

library/core/src/ops/drop.rs
pvdrz: no, it is not that. Methods that don't have a default implementation, like `Clone::clone`, are marked as covered. I can write a panicking drop test but I suspect the drop machinery won't like it

library/core/src/sync/atomic.rs
Hoverbear: As far as I can see it is a public function and part of our subset. Is it impossible to cover this?

Hoverbear: #1816

Hoverbear: Please remove the annotation with #1816 merging.

PR #1804, opened 2025-10-07
[1.90] Enable LLD on aarch64-unknown-linux-gnu

Hoverbear:
We're doing this in https://github.com/ferrocene/ferrocene/pull/1733 which upstream has looked at and decided they'd like to do some greater changes to accomodate.

That said, we'd like the `lld` validation check on 1.90 and the aarch64 change needs to come with it if we want CI to pass.

### DIFF
--- a/compiler/rustc_target/src/spec/targets/aarch64_unknown_linux_gnu.rs
+++ b/compiler/rustc_target/src/spec/targets/aarch64_unknown_linux_gnu.rs
@@ -1,8 +1,37 @@
 use crate::spec::{
-    FramePointer, SanitizerSet, StackProbeType, Target, TargetMetadata, TargetOptions, base,
+    Cc, FramePointer, LinkerFlavor, Lld, SanitizerSet, StackProbeType, Target, TargetMetadata,
+    TargetOptions, base,
 };
 
 pub(crate) fn target() -> Target {
+    let mut base = TargetOptions {
+        features: "+v8a,+outline-atomics".into(),
+        // the AAPCS64 expects use of non-leaf frame pointers per
+        // https://github.com/ARM-software/abi-aa/blob/4492d1570eb70c8fd146623e0db65b2d241f12e7/aapcs64/aapcs64.rst#the-frame-pointer
+        // and we tend to encounter interesting bugs in AArch64 unwinding code if we do not
+        frame_pointer: FramePointer::NonLeaf,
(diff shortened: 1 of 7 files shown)


compiler/rustc_target/src/spec/targets/aarch64_unknown_linux_gnu.rs
```diff
@@ -1,8 +1,37 @@
...
+            | SanitizerSet::CFI
+            | SanitizerSet::KCFI
+            | SanitizerSet::LEAK
+            | SanitizerSet::MEMORY
+            | SanitizerSet::MEMTAG
+            | SanitizerSet::THREAD
+            | SanitizerSet::HWADDRESS,
+        supports_xray: true,
+        ..base::linux_gnu::opts()
+    };
```
jyn514, line 26: Recording for posterity: I validated that this has no code changes other than moving it above the `link_self_contained` block.

ferrocene/tools/self-test/src/test_utils/bin_builder.rs
```diff
@@ -161,7 +172,13 @@ fn main() {
     if let Some(expected_args) = option_env!("EXPECTED_ARGS") {
         let expected = expected_args.split("\t").collect::<Vec<_>>();
         let found = std::env::args().skip(1).collect::<Vec<_>>();
-        assert_eq!(expected, found);
+        if Some("true") == option_env!("EXPECTED_ARGS_STRICT") {
```
jyn514, line 175: the setup here scares me but this specific change looks ok. have you double checked that this fails if it gets a contains-but-not-exact match?

ferrocene/tools/self-test/src/targets.rs
```diff
@@ -210,13 +253,48 @@ mod tests {
...
+            .create();
+
         assert_eq!(
             CheckTargetOutcome::Found,
             check_target(utils.reporter(), utils.sysroot(), &target).unwrap()
         );
         utils.assert_report_success("target installed correctly: x86_64-unknown-linux-gnu");
     }
 
+    #[cfg(all(target_os = "linux", target_arch = "x86_64"))] // Only on x86_64 Linux since the test is specific to that.
```
jyn514, line 270: this test and the one below are for x86. Did you intend to test aarch64? or should we parameterize it over the target?

PR #1805, opened 2025-10-07
Don't generate a metrics report if the build fails

jyn514:
Bootstrap doesn't generate metrics.json when it fails. Don't try uselessly to copy it.

See https://app.circleci.com/pipelines/github/ferrocene/ferrocene/13965/workflows/58a0c496-4178-4351-a27e-7a94028276ba/jobs/95493/parallel-runs/0/steps/0-110 for an example failure this would have avoided.


Hoverbear:
bors merge

PR #2281, opened 2026-04-08
Add `merged-in:VERSION` to PRs that were backported

jyn514:
I have no idea how to test this.

### DIFF
--- a/.github/workflows/backport-labels.yml
+++ b/.github/workflows/backport-labels.yml
@@ -5,6 +5,13 @@
 # to drive the automation. This workflow is responsible for updating them
 # whenever something relevant happens.
 
+# For documentation of github-script, see [its
+# README](https://github.com/actions/github-script?tab=readme-ov-file).
+# For documentation of the `github` object, see
+# [octokit](https://octokit.github.io/rest.js/v22/).
+# For documentation of the `context` object, see [the source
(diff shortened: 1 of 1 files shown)


.github/workflows/backport-labels.yml
```diff
@@ -83,6 +92,12 @@ jobs:
                   repo: context.payload.repository.name,
                   name: "backport:" + version,
                 });
+                github.rest.issues.addLabels({
+                  issue_number: pr,
+                  owner: context.payload.repository.owner.login,
+                  repo: context.payload.repository.name,
+                  labels: ["merged-in:" + version],
```
Hoverbear, line 99: Hmmm... I'm wondering if we should make this like `backported-to:1.95`  🤔

jyn514: I'm not opposed but I don't know how github deals with this if the label doesn't exist already. Given that I have no way of testing this I would really prefer to make this change as minimal as possible.

jyn514: We'll also be able to filter these with `merged-in:1.95 backport:1.95`, if that helps.

Hoverbear (approved):
bors merge

Hoverbear:
bors merge

jyn514:
bors retry

PR #2284, opened 2026-04-08
bump Tests job timeout

tshepang:
60m limit has been reached rather too often

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -109,7 +109,7 @@ jobs:
   tests:
     name: "Tests"
     runs-on: ubuntu-24.04
-    timeout-minutes: 60
+    timeout-minutes: 100
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/commit.yml
```diff
@@ -109,7 +109,7 @@ jobs:
   tests:
     name: "Tests"
     runs-on: ubuntu-24.04
-    timeout-minutes: 60
+    timeout-minutes: 100
     steps:
       - name: Checkout
         uses: actions/checkout@v4
```
jyn514, line 116: can you also set this to only run on upstream pulls? I think there might be a GHA-specific "if branch matches automation/pull-upstream-HEAD/*" or something like that.

tshepang: a pr that inspired this one is not an upstream pull pr, and it hit the limit

jyn514: Right. I'm saying the only time I've seen this job fail, rather than timing out, is in upstream pulls. I would rather not run it at all on most PRs than wait an extra hour before each merge just to get a green check every time.

tshepang: https://github.com/ferrocene/ferrocene/pull/2284/commits/23f8310ad9a20652bd83181fe0c22125546e68ab

jyn514 (approved):
bors merge

PR #2282, opened 2026-04-08
Qualify proc macros

Urhengulas:
This PR updates the Ferrocene qualification documentation to include the `proc-macro` crate type and the `proc_macro` library into its scope.

It documents the results of a HazOp of the usecase of building a proc macro crate and using it in another crate and traces the UI tests related to proc macros to the `rustc --crate-type proc-macro` CLI arg. See `um_rustc_crate_type_proc_macro` in the traceability matrix.

PR #2286, opened 2026-04-09
Various cleanups related to our qualification docs

Urhengulas:
While working on #2282 I noticed a few things to fix/improve and decided to split them out into this PR.

### DIFF
--- a/ferrocene/doc/evaluation-plan/src/qualification-scope.rst
+++ b/ferrocene/doc/evaluation-plan/src/qualification-scope.rst
@@ -17,6 +17,9 @@ they are used in their code.
 
 Additionally a subset of the ``core`` library is certified for use in end-use code.
 
+Qualified tools
+---------------
+
 Qualified tools are:
 
 * ``rustc``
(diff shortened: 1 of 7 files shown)


ferrocene/tools/.gitignore
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+   
+target/
```
tshepang: this is more safe in that it only matches ./target and none that could be in children dirs (I think)
```suggestion
/target/
```

ferrocene/doc/evaluation-plan/src/qualification-scope.rst
```diff
@@ -17,6 +17,9 @@ they are used in their code.
 
 Additionally a subset of the ``core`` library is certified for use in end-use code.
 
+Qualified tools
```
tshepang, line 20: I think this heading is not worth it, seeing that it describes just one tool

Urhengulas: I added this because in #2282 I add a bunch of text before that heading, so it becomes easier to keep an overview with that heading.

tshepang (approved):
bors merge

PR #2291, opened 2026-04-10
backport: squash and use `cherry-pick`, not rebase

jyn514:
This hits many fewer conflicts when backporting. Now, instead of requiring every commit to backport cleanly, we only require the diff between `main` and the PR to backport cleanly. This decreases the fraction of PRs that need to be submitted manually, and also reduces the manual work needed to backport PRs when a manual backport is needed.

There are also various other changes to make the backporting scripts more maintainable and testable; see each individual commit for details. Unlike before, this does not mess with the current git state; running `git cherry-pick --abort` puts you back in exactly the same state you were in before you ran `one.py`.

See https://github.com/ferrocene/ferrocene/pull/2301 for an example PR opened by this automation, and https://github.com/ferrocene/ferrocene/actions/runs/24351997960/job/71108663815 for a sample job output.

---

## Testing

Here is a sample session using `one.py` to do a manual backport for the upcoming release:
```console
$ git checkout origin/release/1.95
HEAD is now at a31dd5ca4b4d Merge #2258
$ git checkout jyn/backport-automation ferrocene/tools/backport ferrocene/tools/automations-common
$ ferrocene/tools/backport/one.py 2245
ERROR: this script cannot be run when you have staged changes; try running `git reset`
$ git reset
Unstaged changes after reset:
M	ferrocene/tools/automations-common/src/automations_common/automated_prs.py
M	ferrocene/tools/backport/all.py
M	ferrocene/tools/backport/one.py
$ ferrocene/tools/backport/one.py 2245                                                                                                                                                                                                         17:45
warning: if no API token is set in the GITHUB_TOKEN env var, requests may be rate-limited
INFO: backporting ferrocene/ferrocene#2245 to a31dd5ca4b4dfee431d6c5be7b58a8ce56e54b00 (please do not CTRL-C)

Auto-merging .circleci/workflows.yml
Auto-merging ferrocene/ci/scripts/setup-uv.sh
CONFLICT (content): Merge conflict in ferrocene/ci/scripts/setup-uv.sh
error: could not apply d3faae633305... Backport ferrocene/ferrocene#2245

ERROR: Encountered a merge conflict backporting ferrocene/ferrocene#2245

INFO: attempting to auto-solve conflicts with mergiraf
WARN Mergiraf: Could not find a supported language for ferrocene/ci/scripts/setup-uv.sh
HELP: Fix the conflicts, run 'git add/rm <filepath>', then run 'git cherry-pick --continue'.
NOTE: Running 'git cherry-pick --abort' will put you back in your original git state.

NOTE: conflicts exist in the following locations:
ferrocene/ci/scripts/setup-uv.sh:7-17
NOTE: 'HEAD' in the conflicted files refers to your current branch, a31dd5ca4b4dfee431d6c5be7b58a8ce56e54b00

$ git diff --name-status --cached
M       .circleci/workflows.yml
U       ferrocene/ci/scripts/setup-uv.sh
$ git status | head -n5
HEAD detached at origin/release/1.95
You are currently cherry-picking commit d3faae633305.
  (fix conflicts and run "git cherry-pick --continue")
  (use "git cherry-pick --skip" to skip this patch)
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)
```

(After this lands on beta, the `git checkout ferrocene/tools && git reset` lines will no longer be needed.)


jyn514:
this does not in fact work; see https://github.com/ferrocene/ferrocene/pull/2292

jyn514:
need to install mergiraf in the backports container

ferrocene/tools/backport/all.py
```diff
@@ -13,11 +13,18 @@
...
 
 
-TEMP_BRANCH = "backport-temp--do-not-use-for-real-code"
+def restore_python_scripts(from_rev):
+    # NOTE: we've checked out base_branch, which might *not* be the same branch as where
+    # we're running this script from. Sync them together.
+    # NOTE: we can't use `self.cmd` because python might have lazily imported it.
+    subprocess.run(["git", "restore", "--source", from_rev,
+                    "ferrocene/tools/automations-common"])
+    subprocess.run(["git", "restore", "--source", from_rev, "ferrocene/tools/backport"])
```
Urhengulas, line 28: Can we catch and handle the error if this command fails due to changes to those files?

ferrocene/tools/backport/one.py
```diff
@@ -18,86 +21,172 @@
...
-        },
-    )
-    exit(result.returncode)
+    authors = git_output("log", "--format=%aN <%aE>", f"{base}..{head}").splitlines()
+    primary_author = authors[0]
+    co_authors = set(authors)
+    co_authors.remove(primary_author)
+    formatted_authors = ""
+    if co_authors:
+        formatted_authors = "\n".join("Co-authored-by: {author}" for author in co_authors)
```
Urhengulas: I was a bit confused because the primary_author is not included in the formatted_authors but only the co_authors and first thought you forgot to add the primary_author.

Now I think that it is correct because the author is passed separately when committing. But then I'd prefer to rename the variable:

```suggestion
    formatted_co_authors = ""
    if co_authors:
        formatted_co_authors = "\n".join("Co-authored-by: {author}" for author in co_authors)
```

Note that the later usage of the variable has to be updated as well.

Urhengulas (changes requested):
The last commit says it wants to be removed as well

ferrocene/tools/backport/all.py
jyn514: Restore overwrites files in the working directory, it doesn't error out. In general `all.py` is not safe to run if you have local changes, I could add something to `run` to check that?

Urhengulas: No, I think it is fine. I though it will try to merge the current and the past files. But if it just resets then its fine.

Urhengulas (approved):
bors merge

PR #2305, opened 2026-04-14
Fix python lints and enable them in CI

jyn514:
I highly recommend reviewing this commit-by-commit. This caught some real bugs. I tried to separate out the changes by severity.

### DIFF
--- a/ferrocene/doc/specification/exts/ferrocene_spec/utils.py
+++ b/ferrocene/doc/specification/exts/ferrocene_spec/utils.py
@@ -9,7 +9,7 @@ def section_id_and_anchor(section):
         try:
             id = [name for name in section["names"] if name.startswith("fls_")][0]
         except IndexError:
-            raise NoSectionIdError()
+            raise NoSectionIdError() from None
     else:
         raise NoSectionIdError()
 
--- a/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/intersphinx_support.py
+++ b/ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/intersphinx_support.py
@@ -8,11 +8,7 @@
 #
 # It also enables intersphinx by default.
 
-from sphinx.builders import Builder
-from sphinx.builders.html import StandaloneHTMLBuilder
 import json
-import sphinx
-import sphinx.ext.intersphinx
 
 
 def inject_intersphinx_mappings(app, config):
--- a/ferrocene/tools/pull-subtrees/pull.py
+++ b/ferrocene/tools/pull-subtrees/pull.py
@@ -37,9 +37,11 @@ import yaml
 class PullLatestTag:
     pass
 
+
 @dataclass
 class PullTag:
     name: str
+
 
 @dataclass
 class PullBranch:
@@ -149,9 +151,9 @@ def fetch_latest_commit(subtree):
         else:
             raise RuntimeError(f"no suitable tags found in {subtree.repo}")
     elif isinstance(subtree.pull, PullBranch):
-        ref = 'refs/heads/' + subtree.pull.name
+        ref = "refs/heads/" + subtree.pull.name
     elif isinstance(subtree.pull, PullTag):
-        ref = 'refs/tags/' + subtree.pull.name
+        ref = "refs/tags/" + subtree.pull.name
     else:
         raise RuntimeError(f"unknown subtree.pull: {subtree.pull}")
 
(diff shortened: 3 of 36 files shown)


ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/intersphinx_support.py
```diff
@@ -8,11 +8,7 @@
 #
 # It also enables intersphinx by default.
 
-from sphinx.builders import Builder
-from sphinx.builders.html import StandaloneHTMLBuilder
 import json
-import sphinx
-import sphinx.ext.intersphinx
```
Urhengulas, line 15: I am unsure if this change is correct. In the internal procedures it is documented ([see](https://public-docs.ferrocene.dev/main/qualification/internal-procedures/docs/sphinx-extensions.html#ferrocene-qualification-extension)):
> The extension also enables `sphinx.ext.intersphinx` without the need to manually enable it in `conf.py`.

And I assume this is due to the import here.

ferrocene/tools/pull-subtrees/pull.py
```diff
@@ -303,17 +305,21 @@ def update_subtree(repo_root, subtree):
...
+                    path = line.split(" ", 1)[1].strip()
+                    who = "upstream"
                 if path:
-                    header = "<<<PULL-UPSTREAM>>> file deleted " + who + "; move the Ferrocene annotations if any, and delete this file"
-                    with open(path, 'r') as original:
+                    header = (
+                        "<<<PULL-UPSTREAM>>> file deleted "
+                        + who
+                        + "; move the Ferrocene annotations if any, and delete this file"
+                    )
```
Urhengulas, line 318: Not related to the changes in this PR but I think this should be:

```suggestion
                    header = f"<<<PULL-UPSTREAM>>> file deleted {who}; move the Ferrocene annotations if any, and delete this file"
```

Urhengulas (changes requested):

ferrocene/doc/specification/exts/ferrocene_spec/utils.py
```diff
@@ -9,7 +9,7 @@ def section_id_and_anchor(section):
         try:
             id = [name for name in section["names"] if name.startswith("fls_")][0]
         except IndexError:
-            raise NoSectionIdError()
+            raise NoSectionIdError() from None
```
Urhengulas, line 12: What is this doing?

ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/intersphinx_support.py
jyn514: I think this is actually done by `app.setup_extension` on line 26 below. See https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.setup_extension:
> Import and setup a Sphinx extension module.

> Load the extension given by the module name. Use this if your extension needs the features provided by another extension. No-op if called twice.

ferrocene/doc/specification/exts/ferrocene_spec/utils.py
jyn514: See https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/:
> When raising a new exception from within an except clause, it's recommended to include a from clause to explicitly set the exception's cause. Without it, Python will implicitly chain from the current exception (setting __context__), but the __cause__ attribute won't be set, which may make debugging slightly more difficult.

Here `IndexError` isn't important to show, so we're hiding it and only showing the NoSectionIdError.

In practice this isn't too important because NoSectionID is always caught earlier in the call stack, but the lint seems useful so I made this change so I didn't need to silence it globally.

ferrocene/tools/pull-subtrees/pull.py
jyn514: this causes the line to go over `ruff`'s line length limit 🥲

ferrocene/doc/specification/exts/ferrocene_spec/utils.py
Urhengulas: TIL, thank you

ferrocene/doc/sphinx-shared-resources/src/ferrocene_qualification/intersphinx_support.py
Urhengulas: I verified that `./x doc ferrocene/doc` still works, therefore I agree the change seems fine.

Urhengulas (approved):
bors merge

PR #2311, opened 2026-04-15
Make some `//@ only-x86` tests target-independent

jyn514:
This makes it easier to test them when doing upstream pulls, without having to be on an x86 host.


Urhengulas (approved):
LGTM. I checked that `./x test tests/ui/ferrocene/compiler-arguments/` still succeeds on Linux.

bors merge

PR #2312, opened 2026-04-15
use self-hosted runners in all easy places

Xylakant:
this uses self-hosted runners in all of the places where it's easily possible.

The only job that needs modification beyond the resource class is the standard library test - circles docker runners don't support ipv6, but ours do. So we can switch this job to a docker executor instead of a vm.

This includes the changes from https://github.com/ferrocene/ferrocene/pull/2340

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1124,10 +1124,6 @@ workflows:
           test-variant: "2021"
           requires:
             - x86_64-linux-build
-      # - x86_64-linux-test-library-std:
-      #     test-variant: "2021"
-      #     requires:
-      #       - x86_64-linux-build
       - x86_64-linux-library-coverage:
           requires:
             - x86_64-linux-build
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -267,28 +267,6 @@ jobs:
           path: build/x86_64-unknown-linux-gnu/ferrocene/traceability-matrix.html
           destination: reports
 
-  # x86_64-linux-test-library-std:
-  #   executor:  docker-x86_64-ubuntu-20
-  #   resource_class: ferrocene/k8s-amd64-medium # 4-core
-  #   parameters:
-  #     test-variant:
-  #       type: string
-  #   environment:
-  #     FERROCENE_HOST: x86_64-unknown-linux-gnu
-  #     # Standard library tests need IPv6, which is not available in container
-  #     # jobs. Because of that we need to run *just* standard library tests in a
-  #     # virtual machine.
-  #     # See ferrocene/ci/split-tasks.py for a list of tasks executed by this.
-  #     SCRIPT: ./x.py --stage 2 test $(ferrocene/ci/split-tasks.py test:library-std) --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
-  #   steps:
-  #     - aws-oidc-auth
-  #     - ferrocene-checkout:
-  #         llvm-subset: true
-  #     - run:
-  #         name: Restore files from the x86_64-linux-build job
-  #         command: ferrocene/ci/scripts/persist-between-jobs.sh restore x86_64-linux-build
-  #     - ferrocene-ci
-
   x86_64-linux-library-coverage:
     executor: docker-x86_64-ubuntu-20
     resource_class: ferrocene/k8s-amd64-large
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1140,7 +1140,7 @@ workflows:
           requires:
             - x86_64-linux-build
       - x86_64-linux-generic-test-container:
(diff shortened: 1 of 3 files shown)


Xylakant:
bors try

Xylakant:
bors try

Xylakant:
bors retry

Xylakant:
bors retry

Xylakant:
bors try

Xylakant:
This includes the changes from #2340

.circleci/workflows.yml
```diff
@@ -195,7 +205,7 @@ jobs:
 
   x86_64-linux-dist-targets:
     executor: docker-x86_64-ubuntu-20
-    resource_class: large # 4-core
+    resource_class: ferrocene/k8s-amd64-large
```
jyn514, line 208: I note that everywhere else you changed 4-core runners to `-medium`, but here you've changed it to large. Is that intentional?

.circleci/workflows.yml
```diff
@@ -208,7 +218,7 @@ jobs:
 
   x86_64-linux-dist-src:
     executor: docker-x86_64-ubuntu-20
-    resource_class: medium # 2-core
+    resource_class: ferrocene/k8s-amd64-medium
```
jyn514, line 221: Elsewhere you changed 2-core runners to `-small`, but here you've changed it to medium. Is that intentional?

.circleci/workflows.yml
```diff
@@ -1042,28 +1080,35 @@ workflows:
...
+          name: x86_64-linux-test-library-std
+          job: test:library-std
+          resource-class: ferrocene/k8s-amd64-medium
           test-variant: "2021"
           requires:
             - x86_64-linux-build
+      # - x86_64-linux-test-library-std:
+      #     test-variant: "2021"
+      #     requires:
+      #       - x86_64-linux-build
```
jyn514:
```suggestion
```

.circleci/workflows.yml
```diff
@@ -257,29 +267,9 @@ jobs:
...
-    parameters:
-      test-variant:
-        type: string
-    environment:
-      FERROCENE_HOST: x86_64-unknown-linux-gnu
-      # Standard library tests need IPv6, which is not available in container
-      # jobs. Because of that we need to run *just* standard library tests in a
-      # virtual machine.
-      # See ferrocene/ci/split-tasks.py for a list of tasks executed by this.
-      SCRIPT: ./x.py --stage 2 test $(ferrocene/ci/split-tasks.py test:library-std) --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
```
jyn514, line 272: Where is this `library-std` job run now? I don't see changes to any other steps, other than `resource_class`.

jyn514: Right I see, this is modified in https://github.com/ferrocene/ferrocene/pull/2343. Presumably it doesn't make sense to merge this PR on its own then, we should just merge 2343?

.circleci/workflows.yml
jyn514: ah, this was fixed in https://github.com/ferrocene/ferrocene/pull/2343/changes/161fec5c944bb70a6141e982855662a40940f23a.

.circleci/workflows.yml
Xylakant: CircleCI has differentiates between steps and jobs. A job is a reusable item that gets used in steps. In this PR, the job is removed and replaced with a step that uses the generic `x86_64-linux-generic-test-container`. #2343 modifies it further to make the entire build step go away.

The matching change to the build step is https://github.com/ferrocene/ferrocene/pull/2312/changes#diff-7bf091450b804ea25dc3d11ac28c50202d19178f7b53f54a02ad59c946afa6a7R1101

.circleci/workflows.yml
Xylakant: We have more resources on the amd64 cluster, so I wanted to give this job a little more power to make it complete faster.

.circleci/workflows.yml
Xylakant: Same here: Using slightly more resources to make things run faster.

jyn514 (approved):

.circleci/workflows.yml
```diff
@@ -1042,28 +1099,35 @@ workflows:
...
             - x86_64-linux-build
-      - x86_64-linux-test-library-std:
+      - x86_64-linux-generic-test-container:
+          name: x86_64-linux-test-library-std
+          job: test:library-std
+          resource-class: ferrocene/k8s-amd64-medium
           test-variant: "2021"
           requires:
             - x86_64-linux-build
+      # - x86_64-linux-test-library-std:
```
Hoverbear: Why leave this?

Hoverbear (approved):

PR #2315, opened 2026-04-16
tidy: Add back `find_base_commit`

jyn514:
This makes tidy avoid checking untracked files locally. There is no change when running in CI, except a warning printed to stderr.


Urhengulas (approved):
bors merge

PR #1904, opened 2025-11-07
pull-upstream: various improvements

jyn514:
See individual commit messages for more info.

To test this, you can run `ferrocene/tools/pull-upstream/pull.sh HEAD main`. Be very sure to ONLY do this if you have already committed all your local changes, or they will be discarded. Before this PR, you would have a bunch of unstaged changes in `src/etc/completions`, and it wouldn't be possible to test on MacOS, only Linux.

Example conflicted hunk after this PR:
```
<<<<<<< ferrocene/main
// ferrocene-annotations: um_rustc_C_target_feature
||||||| dd82fd2034e5
=======
//[hexagon-v60]~? WARN unstable feature specified for `-Ctarget-feature`: `v60`
//[hexagon-v68]~? WARN unstable feature specified for `-Ctarget-feature`: `v68`
//[hexagon-hvxv66]~? WARN unstable feature specified for `-Ctarget-feature`: `hvxv66`
>>>>>>> rust-lang/rust/HEAD--generated-by-pull-upstream
```

Before it would be something like `<<< HEAD` and `>>> pull-upstream-DO-NOT-USE`.

PR #2316, opened 2026-04-16
pull-upstream: more improvements

jyn514:
Primarily, this fixes a regression from https://github.com/ferrocene/ferrocene/pull/1904 (see the second commit). There are also a few other misc improvements, like making the script faster and quieter to run locally, and blessing `symbol-report` automatically.

The symbol-report commit is interesting. There's a tradeoff here because the compiler takes so long to build: Either we can have people on rotation build it locally, like we do today, or we can build it in CI, like after this PR. The downside of building in CI is that the pull-upstream job will take a fair bit longer to run (maybe 20-30 minutes, instead of 10 like today). I think this is a worthwhile tradeoff.


ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -263,12 +261,12 @@ then
...
         if git status --porcelain=v1 | grep "^UU ${lock}$" >/dev/null; then
             echo "pull-upstream: automatically resolving conflict for ${lock}..."
-            git show "${current_commit}:${lock}" > "${lock}"
+            git restore --worktree --source="${current_commit}" -- "${lock}"
 
             # Invoking any Cargo command touching the lockfile will cause the
             # lockfile to be updated. "cargo metadata" is one of the fastest ones.
             # The bootstrap flag is needed as the workspace uses unstable features.
-            if ! cargo metadata --format-version=1 "--manifest-path=${prefix}Cargo.toml" >/dev/null; then
+            if ! cargo metadata --no-deps --format-version=1 "--manifest-path=${prefix}Cargo.toml" >/dev/null; then
```
tshepang: will this not leave external deps in lockfile stale

tshepang: just checked that it will, so new question is if this is something we should care about... if it's something that can happen here

jyn514: whoa, that's a fantastic catch. I didn't expect that. I'll revert this part of the change.

tshepang (approved):

tshepang:
bors merge

PR #2340, opened 2026-04-23
make the windows/darwin job execution conditional

Xylakant:
for a lot of experimentation, we don't really want to run these jobs and they cost a lot of time and a lot of money. Make then configurable.

bors cannot pass the required parameters, so you'll need to start a job via the circleci UI for now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,18 @@
 version: 2.1
 setup: true
 
+parameters:
+  # use this parameter to disable building the windows host. This is useful to cut down build times and
+  # cost in cases where you want to experiment with linux builds and are not interested in windows results
+  full-build-x86_64-pc-windows-msvc-host:
+    type: boolean
+    default: true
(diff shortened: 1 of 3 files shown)


Xylakant:
How to test this: In the circle UI, use "trigger pipeline" and set the `full-build-x86_64-pc-windows-msvc-host` and `full-build-aarch64-darwin-host` to false. This will start a run where execution of these jobs is skipped.

Currently, builds that don't execute windows/darwin get marked as failed in "finish build" as the artifacts produced are missing. I'd leave that for now since it also prevents misinterpreting the build result.

jyn514 (approved):
I assume https://app.circleci.com/pipelines/github/ferrocene/ferrocene/17723/workflows/655cf5cd-840e-4905-953f-bd0d9e17250f is an example of such a run?

This seems good to me, but it would be nice if the `validate-packages-toml.py` understood this parameter and printed a better error message instead of "expected files missing from artifacts".

Hoverbear (approved):
Changes since last review look good

PR #2343, opened 2026-04-24
Refactor std test ci jobs

Xylakant:
This refactors the build jobs around the standard library.

We used to split the std build into its own build step with a different executor because the circleci docker executor is limited when it comes to ipv6. After moving these jobs to self-hosted, this limitation no longer exists and we can refactor the library tests into cleaner sets, one for the full set including std and one for no-std targets

This also reduces the number of jobs we need to run and cuts down on the overhead of starting a new build step.

This changeset is based on the changes of #2312, but split out for easier review.

### DIFF
--- a/ferrocene/ci/split-tasks.py
+++ b/ferrocene/ci/split-tasks.py
@@ -135,14 +135,8 @@ JOBS_DEFINITION: JobsDefinition = {
         ),
 
         # Library tests are the second slowest part of a test run, so we run
-        # them in a separate job to reduce the CI wall clock time. Note that
-        # stdlib tests are run in a separate job, as those require IPv6 and
-        # thus can't be executed in containers due to CircleCI limitations.
-        "library": ["library/core", "library/alloc", "library/test"],
-
-        # The standard library tests require IPv6, which is not available in
-        # containers. Run them separately in a VM.
-        "library-std": ["library/std"],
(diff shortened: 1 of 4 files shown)


ferrocene/ci/split-tasks.py
```diff
@@ -137,6 +137,9 @@ def __lt__(self, other):
         # Library tests are the second slowest part of a test run, so we run
         # them in a separate job to reduce the CI wall clock time.
         "library": ["library/core", "library/alloc", "library/test", "library/std"],
+
+        # This is the library without std, for all targets that have no std.
+        "library-nostd": ["library/core", "library/alloc", "library/test"],
```
jyn514, line 142:
```suggestion
        # This is the library without std, for all targets that have no std.
        # NOTE: This only makes sense on facade targets. Other targets wouldn't be able to compile libtest in the first place.
        "library-nostd": ["library/core", "library/alloc", "library/test"],
```

jyn514 (approved):
LGTM, but please address the reviews on https://github.com/ferrocene/ferrocene/pull/2312 and https://github.com/ferrocene/ferrocene/pull/2340 first.

Hoverbear (approved):

Hoverbear (approved):

Hoverbear:
bors merge

PR #2352, opened 2026-04-28
gate the expensive builds on x86_64 linux tests

Xylakant:
this will increase the total wallclock time in the success case, but lead to a faster fail

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1224,6 +1224,7 @@ workflows:
           test-variant: "2021"
           resource-class: xlarge # 8-core
           requires:
+            - misc-checks
             - x86_64-linux-build
             - x86_64-linux-test
             - x86_64-linux-test-library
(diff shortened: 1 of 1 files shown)


Xylakant:
bors try

Xylakant:
bors try

jyn514:
> this will increase the total wallclock time in the success case, but lead to a faster fail

could we instead set the CircleCI runners to fail-fast? So if the linux builds fail, they automatically cancel the Windows/Darwin jobs? Then we're wasting less time, but without increasing the total wallclock time.

Xylakant:
> > this will increase the total wallclock time in the success case, but lead to a faster fail
> 
> could we instead set the CircleCI runners to fail-fast? So if the linux builds fail, they automatically cancel the Windows/Darwin jobs? Then we're wasting less time, but without increasing the total wallclock time.

We intentionally did not, to enable people to collect failures in other jobs that run in parallel. it would also cancel an expensive llvm-build for win/macos if something unrelated fails in parallel.

PR #2354, opened 2026-04-28
simplify QNX emulator scripts

japaric:
This PR moves both the aarch64 and x86_64 emulators away from using pipe files to drive a serial port console. 

In the case of the x86_64 VM a SSH connection is used for remote command execution as this the easiest to set up using the existing `mkqnximage` workflow. also, instead of sharing a folder from the host to the guest using a virtual drive the remote-test-server binary is now copied over via SSH.

In the case of the aarch64 VM, as we were already patching the IFS build script we tweak further the startup script to run the remote-test-server on boot; this avoids having to remotely run commands after boot. Hence no need for SSH or a serial port console in this case.

Other changes include: removing unnecessary variables; making the IP addresses of both emulators the same; and refactoring duplicated logic into a separate shared file.

This PR is best reviewed on a commit by commit basis

### DIFF
--- a/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
+++ b/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
@@ -16,31 +16,6 @@ panic() {
     exit 1
 }
 
-on_vm() {
-    echo >"${emulatordir}"/pipe.in
-    echo "${@}" >"${emulatordir}"/pipe.in
-    echo 'echo "===$?==="' >"${emulatordir}"/pipe.in
-
-    while read -r line; do
-        if [[ "$line" = "==="*"==="* ]]; then
-            ret=$(echo "$line" | cut -d'=' -f4)
-
-            if [ "$ret" -eq 0  ]; then
-                break
-            else
-                panic failed to run "${@}" on VM. exit code: "$ret"
-            fi
-        fi
-
-        echo "QEMU: $line"
-    done <"${emulatordir}"/pipe.out
-}
-
-on_vm_nowait() {
-    echo >"${emulatordir}"/pipe.in
-    echo "${@}" >"${emulatordir}"/pipe.in
-}
-
 start_vm() {
     echo
     echo "===> setting up QEMU bridge network"
@@ -63,8 +38,6 @@ start_vm() {
     if [ -f "${emulatordir}"/qemu.pid ]; then
         panic a previous instance of the emulator may already be running
     fi
-    rm -f "${emulatordir}"/pipe.*
-    mkfifo "${emulatordir}"/pipe.in "${emulatordir}"/pipe.out
 
     # NOTE(-net nic): the (real) ZCU102 has 4 NICs; only the 4th one can be used in QEMU
     # the unused NICs need to be listed in the command line invocation
@@ -81,16 +54,7 @@ start_vm() {
         -no-reboot \
         -nographic \
         -pidfile "${emulatordir}"/qemu.pid \
-        -serial pipe:"${emulatordir}"/pipe &
-
-    while read -r line; do
-        # last boot message
-        if [[ "$line" = "Boot complete"* ]]; then
-            break
-        fi
-
-        echo "QEMU: ${line}"
-    done <"${emulatordir}"/pipe.out
+        -serial mon:stdio
 }
 
 cmd_prepare() {
@@ -113,9 +77,8 @@ cmd_prepare() {
 
     echo
     echo "===> building remote-test-server"
-    stage="${REMOTE_TEST_SERVER_STAGE-1}"
-    ./x build src/tools/remote-test-server --target "${nto_target}" --stage "${stage}"
-    cp build/host/"stage${stage}-tools"/"${nto_target}"/release/remote-test-server "${emulatordir}"/src/install/aarch64le/sbin
+    ./x build src/tools/remote-test-server --target "${nto_target}"
+    cp build/host/"stage2-tools"/"${nto_target}"/release/remote-test-server "${emulatordir}"/src/install/aarch64le/sbin
 
     echo
     echo "===> building IFS"
@@ -151,8 +114,8 @@ cmd_prepare() {
 /bin/sh=sh\
 /bin/echo=echo' "${buildscript}"
 
-    # signal that boot is complete
-    sed -i '326i echo "Boot complete"' "${buildscript}"
+    # run remote-test-server instead of console
+    sed -i '326i RUST_TEST_THREADS=1 remote-test-server -v --bind 0.0.0.0:12345 --sequential' "${buildscript}"
 
     # use virtual SD card as TMPDIR
     sed -i '268i export TMPDIR=/mnt/tmp' "${buildscript}"
@@ -183,14 +146,6 @@ cmd_run() {
     fi
 
     start_vm
-
-    echo
-    echo "===> starting remote-test-server"
-    on_vm_nowait RUST_TEST_THREADS=1 remote-test-server -v --bind 0.0.0.0:12345 --sequential
-
-    while read -r line; do
-        echo "QEMU: $line"
-    done <"${emulatordir}"/pipe.out
 }
 
 if [[ $# -eq 1 ]] && [[ "$1" = "prepare" ]]; then
--- a/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
+++ b/ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
@@ -5,21 +5,12 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-# this file resides in `/ferrocene/ci/scripts/(here)`
-basedir="$(realpath $(dirname $0)/../../..)"
-qnxdir="$(realpath $(dirname $(which qcc))/../../../../..)"
-emulatordir=/tmp/emulator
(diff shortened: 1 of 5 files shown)


japaric:
bors try

japaric:
bors try

japaric:
bors try

ferrocene/ci/scripts/emulated-x86_64-qnx-test-runner.sh
```diff
@@ -120,6 +95,16 @@ lib/libpci.so.2.3|' "${ifsbuild}"
     cp build/host/"stage${stage}-tools"/"${nto_target}"/release/remote-test-server "${emulatordir}"/shared/
 }
 
+on_vm() {
+    # disable user prompts around host key checking
+    # suppress warnings about adding $vm_ipv4_addr to list of known hosts
+    ssh \
+        -i "${emulatordir}"/$ssh_id \
+        -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null \
```
jyn514:
```suggestion
        -o StrictHostKeyChecking=accept-new \
```
but I see you removed this altogether in a later commit, so probably not worth it.

ferrocene/ci/scripts/emulated-x86_64-qnx-test-runner.sh
```diff
@@ -89,10 +88,8 @@ lib/libpci.so.2.3|' "${ifsbuild}"
 
     echo
     echo "===> building and copying remote-test-server into rootfs"
-    stage="${REMOTE_TEST_SERVER_STAGE-1}"
-    ./x build src/tools/remote-test-server --target "${nto_target}" --stage "${stage}"
-    mkdir -p "${emulatordir}"/shared/
-    cp build/host/"stage${stage}-tools"/"${nto_target}"/release/remote-test-server "${emulatordir}"/shared/
+    ./x build src/tools/remote-test-server --target "${nto_target}"
```
jyn514: Please let's not use the defaults here. I believe this was using `--stage=1` before, could you add that back explicitly?

ferrocene/ci/scripts/emulated-x86_64-qnx-test-runner.sh
```diff
@@ -105,8 +102,17 @@ on_vm() {
         root@$vm_ipv4_addr "${@}"
 }
 
+copy_to_vm() {
+    # same flags as `on_vm`
```
jyn514: ~~on_vm doesn't appear to have any comments on what flags it accepts? I think `$1` is the filepath on the host and `$2` is the path on the guest, can you add that as a comment?~~

you removed this in a later commit.

ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
```diff
@@ -113,9 +77,8 @@ cmd_prepare() {
 
     echo
     echo "===> building remote-test-server"
-    stage="${REMOTE_TEST_SERVER_STAGE-1}"
-    ./x build src/tools/remote-test-server --target "${nto_target}" --stage "${stage}"
-    cp build/host/"stage${stage}-tools"/"${nto_target}"/release/remote-test-server "${emulatordir}"/src/install/aarch64le/sbin
+    ./x build src/tools/remote-test-server --target "${nto_target}"
```
jyn514: Let's please add back `--stage=1`.

ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
```diff
@@ -151,8 +114,8 @@ cmd_prepare() {
 /bin/sh=sh\
 /bin/echo=echo' "${buildscript}"
 
-    # signal that boot is complete
-    sed -i '326i echo "Boot complete"' "${buildscript}"
+    # run remote-test-server instead of console
+    sed -i '326i RUST_TEST_THREADS=1 remote-test-server -v --bind 0.0.0.0:12345 --sequential' "${buildscript}"
```
jyn514, line 116: This feels very fragile but if it worked before we don't have to fix it now. I don't love hard-coding line numbers though :/

jyn514 (commented):
overall this seems much more reliable 👍

ferrocene/ci/scripts/emulated-x86_64-qnx-test-runner.sh
japaric: I'll set it to `--stage 2` which is what CI was using

this was configurable for the benefit of local development as `--stage 1` *used* to build remote-test-server against stage1 libstd which is faster than building against stage2 libstd which CI was doing

but the behavior now seems to be to completely ignore the `--stage` flag; always build remote-test-server against stage1 libstd and yet place the artifact in the stage2-tools directory. the latest change broke the local use of the REMOTE_TEST_SERVER_STAGE env var.

given that `--stage 1` has no speed advantage for local development vs `--stage 2` I'll set it to what we were using in CI

ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
japaric: we checksum the input file, ".build file", so an error will be raised if that changes in any way, which may only happen if we update the BSP package version. if the package version does change then the whole build script needs to be reviewed anyways as the BSP is meant to be used on real hardware, not an emulator, so various changes will probably be needed (e.g. removing services that rely on peripherals on emulated by QEMU).

for this particular change, the startup script contents is part of ("inlined in") the .build source file so a source change is needed. on x86_64, the startup script was a separate source file that was copied into the IFS so in that case we were able to append the changes to the end of the startup script without relying on precise line number locations

ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
japaric: done. used `--stage=2` as per the rationale in my previous/other reply

ferrocene/ci/scripts/emulated-aarch64-qnx-test-runner.sh
jyn514: this helps a lot, thank you! could you add that as a comment in the source?

ferrocene/ci/scripts/emulated-x86_64-qnx-test-runner.sh
jyn514: I'm very confused. The difference is which *compiler* remote-test-server is built with, not which libstd it's built against.

Can you show a transcript of it being ignored? I see it doing the right thing locally:
```
$ BOOTSTRAP_TRACING=info ./x b remote-test-server --stage=1
13:12:40.738  INFO  > tool::ToolBuild {build_compiler: Compiler { stage: 0, host: aarch64-apple-darwin, forced_compiler: false }, target: aarch64-apple-darwin, tool: "remote-test-server", path: "src/tools/remote-test-server", mode: ToolTarget, source_type: InTree, extra_features: [], allow_features: "", cargo_args: [], artifact_kind: Binary} (builder/mod.rs:1661)
Building stage1 remote-test-server (stage0 -> stage1, aarch64-apple-darwin)

$ BOOTSTRAP_TRACING=info ./x b remote-test-server --stage=2
13:12:34.935  INFO  > tool::ToolBuild {build_compiler: Compiler { stage: 1, host: aarch64-apple-darwin, forced_compiler: false }, target: aarch64-apple-darwin, tool: "remote-test-server", path: "src/tools/remote-test-server", mode: ToolTarget, source_type: InTree, extra_features: [], allow_features: "", cargo_args: [], artifact_kind: Binary} (builder/mod.rs:1661)
Building stage2 remote-test-server (stage1 -> stage2, aarch64-apple-darwin)
```

japaric: > The difference is which compiler remote-test-server is built with

with `--stage 2`, only stage1 compiler and libraries are built. 

using `--stage 1` instead of `--stage 2` builds the same artifacts and the build logs are the same (see below). running `./x build --stage 2 (..)` right after `./x clean; ./x build --stage 1 (..)` is a no-op and does not build anything extra. 
``` console
$ ./x clean
$ ./x build src/tools/remote-test-server --stage 2 --target x86_64-pc-nto-qnx710
Building bootstrap
(..)
Building stage1 compiler artifacts (stage0 -> stage1, x86_64-unknown-linux-gnu)
(..)
Building stage1 lld-wrapper (stage0 -> stage1, x86_64-unknown-linux-gnu)
(..)
Building stage1 library artifacts (stage1 -> stage1, x86_64-unknown-linux-gnu)
(..)
Building stage1 library artifacts (stage1:x86_64-unknown-linux-gnu -> stage1:x86_64-pc-nto-qnx710)
(..)
Building stage2 remote-test-server (stage1:x86_64-unknown-linux-gnu -> stage2:x86_64-pc-nto-qnx710)
(..)
```

jyn514: Right I see, I think you're right that this is a cross-compilation thing. Bootstrap can't build a standard library for the target at stage 1 so it's forced to build for the host and bump the stage to 2.

Ok, can you leave a comment saying that?

jyn514 (approved):
bors merge

japaric:
bors retry
network failure

Hoverbear:
bors retry

jyn514:
bors retry

PR #2336, opened 2026-04-22
Fix several bugs in `ferrocene::prevalidated`

jyn514:
Thanks to @Urhengulas for catching these.

1. The THIR pass didn't check associated items at all. This is fixed in the first commit and tested in `tests/ui/ferrocene/lint-prevalidated/thir-impl-item.rs`. The second commit adds the missing items to libcore.
2. The post-mono pass didn't check dynamic casts. I had correctly "failed closed" here -- the THIR pass would ICE if it wasn't able to check a cast — but because of 1. there were several casts in libcore that weren't caught, including in libcore. The ICE is fixed in the third commit and the post-mono pass is extended in the fourth. This is tested in `tests/ui/ferrocene/lint-prevalidated/dyn-cast-post-mono.rs`.
3. The post-mono pass can't handle MIR inlining. ~~I tried to get this working (see https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/SourceScopeData.20-.3E.20SourceScope/with/590200295), but wasn't able to figure out how to do it reliably. Instead I just disable mir inlining in the 5th commit.~~ I got this working, and added a test in `tests/ui/ferrocene/lint-prevalidated/mir-inlining.rs`.

PR #2364, opened 2026-05-05
Expand coverage to include recent changes

Hoverbear:
#2336 fixed a number of bugs in the `#[ferrocene::prevalidated]` lint, this handles coverage coming from those changes.

### DIFF
--- a/library/core/src/fmt/mod.rs
+++ b/library/core/src/fmt/mod.rs
@@ -2062,9 +2062,12 @@ impl<'a> Formatter<'a> {
             } else {
                 // Padding widths are capped at `u16`, so reaching this branch means
                 // the formatted output is also shorter than `u16::MAX`.
+                #[ferrocene::annotation("The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug")]
                 let len = match u16::try_from(len) {
                     Ok(len) => len,
-                    Err(_) => unreachable!(),
+                    Err(_) => {
+                        unreachable!()
+                    },
                 };
                 let post_padding = self.padding(width - len, Alignment::Right)?;
                 // SAFETY: Per the precondition.
--- a/library/core/src/fmt/mod.rs
+++ b/library/core/src/fmt/mod.rs
@@ -2065,11 +2065,9 @@ impl<'a> Formatter<'a> {
                 let len = match u16::try_from(len) {
                     Ok(len) => len,
                     #[ferrocene::annotation(
-                        "The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug"
+                        "The `Err(_)` branch is explicitly marked unreachable, causing it to be reached would be a bug"
                     )]
-                    Err(_) => {
-                        unreachable!()
-                    }
+                    Err(_) => unreachable!(),
                 };
                 let post_padding = self.padding(width - len, Alignment::Right)?;
                 // SAFETY: Per the precondition.
--- a/library/core/src/fmt/mod.rs
+++ b/library/core/src/fmt/mod.rs
@@ -2064,10 +2064,12 @@ impl<'a> Formatter<'a> {
                 // the formatted output is also shorter than `u16::MAX`.
                 let len = match u16::try_from(len) {
                     Ok(len) => len,
-                    #[ferrocene::annotation("The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug")]
+                    #[ferrocene::annotation(
+                        "The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug"
+                    )]
                     Err(_) => {
                         unreachable!()
-                    },
+                    }
                 };
                 let post_padding = self.padding(width - len, Alignment::Right)?;
                 // SAFETY: Per the precondition.
--- a/library/core/src/fmt/mod.rs
+++ b/library/core/src/fmt/mod.rs
@@ -2062,9 +2062,9 @@ impl<'a> Formatter<'a> {
             } else {
                 // Padding widths are capped at `u16`, so reaching this branch means
                 // the formatted output is also shorter than `u16::MAX`.
-                #[ferrocene::annotation("The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug")]
                 let len = match u16::try_from(len) {
                     Ok(len) => len,
+                    #[ferrocene::annotation("The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug")]
                     Err(_) => {
                         unreachable!()
                     },
--- a/library/core/src/intrinsics/mod.rs
+++ b/library/core/src/intrinsics/mod.rs
@@ -3164,6 +3164,12 @@ pub const fn minimum_number_nsz_f16(x: f16, y: f16) -> f16 {
 #[rustc_intrinsic_const_stable_indirect]
 #[rustc_intrinsic]
 #[ferrocene::prevalidated]
+#[ferrocene::annotation(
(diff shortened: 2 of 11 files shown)


library/coretests/tests/ferrocene/slice.rs
```diff
@@ -305,3 +305,22 @@ fn slice_split_first_chunk() {
...
+    let x = b"absentmindednessabsentmindednessabsentmindednessabsentmindedness";
+    let y = b"ABSENTMINDEDNESSABSENTMINDEDNESSABSENTMINDEDNESSABSENTMINDEDNES0";
+    assert!(!x.eq_ignore_ascii_case(y));
+
+    // Two equal sized (16 chunked) values that are equal.
+    let x = b"absentmindednessabsentmindednessabsentmindednessabsentmindedness";
+    let y = b"ABSENTMINDEDNESSABSENTMINDEDNESSABSENTMINDEDNESSABSENTMINDEDNESS";
+    assert!(x.eq_ignore_ascii_case(y));
+
+    // Two inequal sized (16 chunked) values with an inequality at the end.
```
Hoverbear:
```suggestion
    // Two equal sized (16 chunked) values with an inequality at the end.
```

library/core/src/fmt/mod.rs
```diff
@@ -2064,7 +2064,12 @@ impl<'a> Formatter<'a> {
                 // the formatted output is also shorter than `u16::MAX`.
                 let len = match u16::try_from(len) {
                     Ok(len) => len,
-                    Err(_) => unreachable!(),
+                    #[ferrocene::annotation(
+                        "The `Err(_)` branch is explicitly marked unreachable, causing it be reached would be a bug"
```
skade:
```suggestion
                        "The `Err(_)` branch is explicitly marked unreachable, causing it to be reached would be a bug"
```

library/core/src/fmt/mod.rs
```diff
@@ -2064,7 +2064,12 @@ impl<'a> Formatter<'a> {
...
+                    )]
+                    Err(_) => {
```
skade: This expands the original to 3 lines - is this desired?

library/core/src/slice/ascii.rs
```diff
@@ -162,6 +162,14 @@ impl [u8] {
                 if !eq_ignore_ascii_inner(a_rem, b_rem) {
                     return false;
                 }
+            } else {
+                #[ferrocene::annotation("
+                    This branch of is not coverable from public API. \
```
skade: > This branch of is not coverable from public API.

I don't understand this sentence. "This branch of `if`"?

skade (changes requested):
Left relevant review comments.

library/core/src/intrinsics/mod.rs
```diff
@@ -3093,6 +3093,12 @@ pub const fn minimum_number_nsz_f16(x: f16, y: f16) -> f16 {
 #[rustc_intrinsic_const_stable_indirect]
 #[rustc_intrinsic]
 #[ferrocene::prevalidated]
+#[ferrocene::annotation(
+    "This function is replaced by an LLVM instrinsic in LLVM 22 or later (which Ferrocene uses). \
+    See `codegen_intrinsic_call` in `compiler/rustc_codegen_llvm/src/intrinsic.rs. \
+    A manual test was done with this functions forcibly included and it was fully covered,
+    see `test_minimum_number_nsz_f32` in `library/coretests/tests/ferrocene/intrinsics.rs`."
+)]
 pub const fn minimum_number_nsz_f32(x: f32, y: f32) -> f32 {
```
jyn514, line 3173: it's fine for this release, but we could get coverage info for this automatically by outlining it to `instrinsics/fallback.rs`, which never get replaced by LLVM intrinsics.

library/coretests/tests/ferrocene/fmt.rs
```diff
@@ -488,3 +488,74 @@ fn from_fn_debug_fmt() {
...
+fn test_fmt_parse_int_error_not_a_power_of_two() {
+    use core::num::IntErrorKind;
+    pub struct FakeParseIntError {
+        #[allow(dead_code)] // This is used for transmuting.
+        pub kind: IntErrorKind,
+    }
+
+    let val = FakeParseIntError { kind: IntErrorKind::NotAPowerOfTwo };
+    // We need to do a truly cursed thing here since there are no situations in public API where
+    // we can conjure ourselves a `ParseIntError` with a `NotAPowerOfTwo` in it.
```
jyn514: if this is really the case then shouldn't we instead mark that code as unreachable?

library/coretests/tests/ferrocene/str.rs
```diff
@@ -195,6 +195,11 @@ fn test_utf8_chunks_next() {
     let mut chunks = replacement_character.utf8_chunks();
     assert_eq!("�", chunks.next().unwrap().valid());
     assert!(chunks.next().is_none());
+
+    let two_width_chars = "§ɮʃ ͒ʦ";
```
jyn514, line 199: my favorite emoji.

library/coretests/tests/ferrocene/fmt.rs
Hoverbear: There are ways to create `IntErrorKind::PowerOfTwo`, but no ways to create a `ParseIntError` holding a `IntErrorKind::PowerOfTwo` I can find.

Are you suggesting replacing https://github.com/ferrocene/ferrocene/blob/ecfd149e34f7eff7daed153cd058d75ad4177cfd/library/core/src/num/error.rs#L158 with an unreachable? I think that's fair since we should catch test failures if that ever changes.

Do you consider that option better than my conjuration? :sweat_smile:

library/coretests/tests/ferrocene/str.rs
Hoverbear: Nominating ` ͒ʦ` as the new (*wireless*) typescript extension.

library/core/src/intrinsics/mod.rs
Hoverbear: Hrm, so this would mean actually relocating these functions to intrinsics::fallback::, re-exporting them from the old location, and adding functionality behind #[unstable(feature = "ferrocene_test", issue = "none")] to call it for coverage...

Feels a bit invasive but does track reasonably. We'd definitely see a conflict if it ever got disturbed.

library/coretests/tests/ferrocene/fmt.rs
jyn514: exactly, yes, mark specifically the Display impl with unreachable.

i think it's weird for upstream to reuse the type between different APIs like this, but that's another problem ...

library/core/src/intrinsics/mod.rs
jyn514: We discussed internally that this is more invasive a change than we want to make.

library/coretests/tests/ferrocene/fmt.rs
Hoverbear: I'm rather unsure about making a load bearing change to the core library here.

jyn514: i was imagining something like this: https://github.com/ferrocene/ferrocene/blob/68d8a866fa85c568b93dba875844eee5d4186e3f/library/core/src/num/imp/flt2dec/strategy/dragon.rs#L274-L275

jyn514 (approved):
thanks for working on this, and sorry i forgot about it for so long.

bors merge

jyn514:
bors r-

want to merge everything into 1.97 before we queue things up for main

PR #2362, opened 2026-05-04
Small symbol-report improvements

jyn514:
See individual commit messages for details. The second one (Make debug-assertions required) is the main motivation, it kept tripping me up during backports.

PR #2366, opened 2026-05-07
Replace all mentions of `x.py run update-certified-core-symbols`

ada4a:
This was probably missed during https://github.com/ferrocene/ferrocene/pull/2362, cc @jyn514

Its use in `pull.sh` gave me the following error:
```
ERROR: no `run` rules matched [update-certified-core-symbols]
HELP: run `x.py run --help --verbose` to show a list of available paths
NOTE: if you are adding a new Step to bootstrap itself, make sure you register it with `describe!`
```
so I'm assuming it should be removed everywhere


Hoverbear (approved):
bors merge

PR #2367, opened 2026-05-07
Only commit mergiraf's resolutions if there were any

ada4a:
Tested with https://github.com/ferrocene/ferrocene/pull/2368


tshepang (approved):
bors merge

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -347,11 +347,14 @@ then
             # file unchanged. Therefore, everything that was changed is good to add.
             git add .
 
-            git commit -F- <<EOF
+            # Only commit if mergiraf has resolved anything
+            if git status --porcelain=v1 | grep "^ M " >/dev/null; then
```
ada4a, line 351: Note: the reason I didn't use `commit_if_modified` here is that it requires the commit message to be passed as the second argument, while here we have a multiline message and so (need to?) pass it via stdin.

Maybe it'd make sense to add the above as a comment

jyn514: bash has multi-line strings. but this is ok too, it's not a big deal.

PR #2372, opened 2026-05-12
raise an error if linker does not support -fuse-ld= flag

japaric:
the upstream behavior is to re-invoke the linker without the -fuse-ld= flag, without notifying the user


japaric:
bors try

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -926,6 +926,12 @@ fn link_natively(
             && out.contains("-fuse-ld=lld")
             && cmd.get_args().iter().any(|e| e.to_string_lossy() == "-fuse-ld=lld")
         {
+            // Ferrocene-only: if the linker driver does not support the -fuse-ld flag, we
+            // treat that as a fatal error
+            if true {
+                sess.dcx().emit_fatal(errors::NoFuseLdSupport);
+            }
```
jyn514:
```suggestion
            sess.dcx().emit_fatal(errors::NoFuseLdSupport);
```

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
```diff
@@ -0,0 +1,77 @@
...
+        .crate_type("bin")
+        .input("main.rs")
+        .print("link-args")
+        .out_dir(&out_dir)
+        .run();
+
+    let stdout = output.stdout_utf8();
+    if !stdout.contains("-fuse-ld=") {
+        // this target does not use `-fuse-ld` flag so the fallback path cannot be tested
+        return;
```
jyn514: can we add a couple targets where we know we want this test to run? I think aarch64 and x86_64 are the ones where we currently want to use LLD? although that might depend on config.toml ...

in general it's really hard to keep track of which tests are run when they do their own custom logic like this. I'd prefer to make this a hard panic, then add `//@ only-x86_64-unknown-linux-gnu` at the top of the file.

compiler/rustc_codegen_ssa/src/back/link.rs
```diff
@@ -926,6 +926,12 @@ fn link_natively(
             && out.contains("-fuse-ld=lld")
             && cmd.get_args().iter().any(|e| e.to_string_lossy() == "-fuse-ld=lld")
         {
+            // Ferrocene-only: if the linker driver does not support the -fuse-ld flag, we
+            // treat that as a fatal error
+            if true {
+                sess.dcx().emit_fatal(errors::NoFuseLdSupport);
```
jyn514: nit: since this is always a static string you can just do `dcx().fatal("does not support -fuse-ld")`.

(do you know about https://doc.rust-lang.org/nightly/nightly-rustc/ ?)

jyn514 (dismissed):
really nice work, well done :) couple small comments but looks fine otherwise.

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    let mut output = bare_rustc()
+        .target(target())
+        .crate_type("bin")
+        .input("main.rs")
+        .override_path(&new_path)
+        .out_dir(&out_dir)
+        .run_fail();
+
+    assert!(output.stderr_utf8().contains(NEEDLE));
```
jyn514: i think you might be able to use https://doc.rust-lang.org/nightly/nightly-rustc/run_make_support/diff/struct.Diff.html here so it's closer to a UI test.

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
```diff
@@ -0,0 +1,77 @@
+//! Check that if Ferrocene's rustc is used with a linker driver that does not support the
```
jyn514, line 4: since this is depending on defaults and requires that we test exactly `cc` as the linker, can you mark it as `//@ no-cross-compile`?

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
japaric: > in general it's really hard to keep track of which tests are run when they do their own custom logic like this. I'd prefer to make this a hard panic, then add //@ only-x86_64-unknown-linux-gnu at the top of the file.

yeah, I agree. I would however instead have an ignore list so that newly added targets panic, e.g. if `-fuse-ld` is missing. that way new targets get reviewed, instead of having to remember to check if new targets should be added to the list.

compiler/rustc_codegen_ssa/src/back/link.rs
japaric: the `if true` needs a comment but the rationale is that the `emit_fatal` is a divergent function so it raises an "unreachable_code" warning. one option is to remove the code that follows but then that raises the chance of future merge conflicts

compiler/rustc_codegen_ssa/src/back/link.rs
japaric: nice. that's better. I didn't think of looking for a differently named method and `emit_fatal("foo")` did not work

> (do you know about https://doc.rust-lang.org/nightly/nightly-rustc/ ?)

yes, at some point but forgot it existed :sweat_smile:

compiler/rustc_codegen_ssa/src/back/link.rs
jyn514: ah, gotcha. can you add `#[expect(unreachable_code)]` then, to make it more clear what's happening?

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
japaric: > since this is depending on defaults 

yes, that's always the case

> requires that we test exactly cc as the linker,

that only happens to be the case now given the targets with test but a cross-compile target like QNX could be tested like this if `qcc` supported `-fuse-ld`. what I'm trying to say is that we may want to test future cross-compile targets so as per this comment ...

> https://github.com/ferrocene/ferrocene/pull/2372#discussion_r3248037041

I would go with an explicit ignore list that list all ignored targets. which I guess could be refactor a bit if we had something like `@ ignore-directly-calls-lld`

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
jyn514: i think gathering that list is going to be annoyingly hard ... maybe you can parse it out of .circleci/ though.

compiler/rustc_codegen_ssa/src/back/link.rs
jyn514: no worries. yeah rustc's API is uhhhh questionable at best lol. a lot of this is left over from the translation initiative that never went anywhere but also never got reverted ...

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
jyn514: 👍 makes sense. `needs-rust-lld` might be what you're looking for, there's docs in https://rustc-dev-guide.rust-lang.org/tests/directives.html#controlling-when-tests-are-run and a complete reference in `src/tools/compiletest/src/directives/directive_names.rs`.

compiler/rustc_codegen_ssa/src/back/link.rs
japaric: I tried `#[expect]` and it either has to be placed on each and every of the stamements that become unreachable which is noisy and merge conflict prone; or it has to be placed on the `if` statement that surrounds the `fatal` call. Wrapping all the following statement in an expression block and placing a single `#[expect]` on the block also works but that also modifies several lines by way of indentation.

I'll go with the `#[expect]` on `if` option.

tests/run-make/ferrocene/no-fuse-ld-fallback/rmake.rs
japaric: despite the use of ["target"](https://github.com/ferrocene/ferrocene/blob/6e591586e963df72fb77f8e0db05f021ea8fe221/src/tools/compiletest/src/directives/needs.rs#L429) in the `needs-rust-lld` logic, as far as I could test that directive pretty much implies `ignore-cross-compile` because the `rust-lld` binary is always placed in `$SYSROOT/lib/rustlib/$HOST_TUPLE/bin` so even if the target is `aarch64-unknown-ferrocene.facade`, which does use `rust-lld` as the linker, the `needs-rust-lld` directive will ignore that target.

japaric: I added a new directive to only test certain targets based on their default linker-flavor as that's what used to decided whether to pass `-fuse-ld` to the linker. that way if the target's default linker flavor changes (e.g. Linux switched from GCC+LD to GCC+LLD) the test should start covering it without updating the list.

jyn514 (approved):
bosr merge

jyn514:
bors merge

tshepang:
bors retry

PR #2373, opened 2026-05-13
misc cleanups to FlipLink tests

jyn514:
- Checkout the submodule if it hasn't been already
- Give better errors if we know the tests will fail.
- flip-link hardcodes exactly thumbv7em-none-eabi, so we have to do the same in `is_default_step`.


Hoverbear (approved):
Thank you!


bors merge

PR #2381, opened 2026-05-26
Remove `./x test certified-core-symbols` and some small related refactors

Urhengulas:
Now `./x test ferrocene/doc/symbol-report.csv [,--bless]` is the one and only command to check and update the core symbol report.

### DIFF
--- a/src/bootstrap/src/ferrocene/run.rs
+++ b/src/bootstrap/src/ferrocene/run.rs
@@ -132,10 +132,6 @@ impl Step for CertifiedCoreSymbols {
         run.never()
     }
 
-    fn make_run(run: RunConfig<'_>) {
-        run.builder.ensure(CertifiedCoreSymbols::new(run.builder, run.target));
-    }
-
     fn run(self, builder: &Builder<'_>) -> Self::Output {
         if !builder.config.std_debug_assertions {
             panic!("generating the core symbol report requires `rust.debug-assertions-std=true`");
--- a/src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
+++ b/src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
@@ -20,7 +20,7 @@ impl Step for CertifiedCoreSymbols {
     const IS_HOST: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
-        run.path(TRACKED_FILE).alias("certified-core-symbols")
+        run.path(TRACKED_FILE)
     }
 
     fn is_default_step(builder: &Builder<'_>) -> bool {
@@ -44,8 +44,8 @@ impl Step for CertifiedCoreSymbols {
         // load the expected list of qualified functions
         let expected_path = Path::new(TRACKED_FILE);
         let mut expected: Vec<String> = Default::default();
-        let reader = builder.read(expected_path);
-        for qualified_name in reader.lines() {
+        let expected_content = builder.read(expected_path);
+        for qualified_name in expected_content.lines() {
             expected.push(qualified_name.to_string());
         }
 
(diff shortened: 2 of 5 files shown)


src/bootstrap/src/ferrocene/run.rs
```diff
@@ -132,10 +132,6 @@ impl Step for CertifiedCoreSymbols {
         run.never()
     }
 
-    fn make_run(run: RunConfig<'_>) {
-        run.builder.ensure(CertifiedCoreSymbols::new(run.builder, run.target));
-    }
```
jyn514, line 137: 👍 we should merge this regardless of anything else, it's dead code.

src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
```diff
@@ -44,8 +44,8 @@ impl Step for CertifiedCoreSymbols {
         // load the expected list of qualified functions
         let expected_path = Path::new(TRACKED_FILE);
         let mut expected: Vec<String> = Default::default();
-        let reader = builder.read(expected_path);
-        for qualified_name in reader.lines() {
+        let expected_content = builder.read(expected_path);
+        for qualified_name in expected_content.lines() {
```
jyn514, line 48: 👍 seems like a good change

src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
```diff
@@ -20,7 +20,7 @@ impl Step for CertifiedCoreSymbols {
     const IS_HOST: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
-        run.path(TRACKED_FILE).alias("certified-core-symbols")
+        run.path(TRACKED_FILE)
```
jyn514, line 23: i have mixed feelings about this one; i think `certified-core-symbols` was reasonably clear. i don't have strong opinions one way or another though, i'll approve so you can decide.

jyn514 (approved):

src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
tshepang: "certified-core-symbols" is also more clear than "symbol-report.csv"... was thinking we should rename the file to something close, like "../certified-core-symbols", but it felt out of scope for this PR

Urhengulas: > "certified-core-symbols" is also more clear than "symbol-report.csv"... was thinking we should rename the file to something close, like "../certified-core-symbols", but it felt out of scope for this PR

I was thinking the same actually! But I am a bit nervous about doing this since this has big potential for conflicts with other PRs.

Urhengulas: > i have mixed feelings about this one; i think `certified-core-symbols` was reasonably clear. i don't have strong opinions one way or another though, i'll approve so you can decide.

I honestly don't care which one we take. I just would prefer there to be only one and not two.

Urhengulas: I prefer to only keep the filename and discuss renaming it to something like `ferrocene/doc/certified-core-symbols.csv`.

Urhengulas:
bors merge r=jyn514

src/bootstrap/src/ferrocene/test/certified_core_symbols.rs
Urhengulas: The main advantage is that we don't read the file a second time, as it was done before in line 80.

Hoverbear:
https://github.com/ferrocene/ferrocene/pull/2389

Urhengulas:
bors retry

PR #2387, opened 2026-06-01
Resolve a caching issue with `symbol-report`'s builds.

Hoverbear:
If a user ran the following:

```
x.py clean
ferrocene/etc/upstream-pull-basic-checks.sh
```

They'd see something like

```
Generated report at
/home/ana/git/ferrocene/ferrocene/build/x86_64-unknown-linux-gnu/stage1-std/x86_64-unknown-linux-gnu/dist/symbol-report.json

thread 'main' (125911) panicked at
src/bootstrap/src/ferrocene/run/update_certified_core_symbols.rs:37:77:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind:
NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a
backtrace
Build completed unsuccessfully in 0:00:05
```

This was due to a caching issue. I spoke with @jyn514 and we think this might be the least worst solution for that problem.

### DIFF
--- a/src/bootstrap/src/ferrocene/run.rs
+++ b/src/bootstrap/src/ferrocene/run.rs
@@ -145,8 +145,14 @@ impl Step for CertifiedCoreSymbols {
         let symbol_report = builder.ensure(SymbolReport { target_compiler: build_compiler });
 
         // c.f. check::std
-        let mut cargo =
-            Cargo::new(builder, build_compiler, Mode::Std, SourceType::InTree, target, Kind::Check);
+        let mut cargo = Cargo::new(
+            builder,
+            build_compiler,
+            Mode::Std,
+            SourceType::InTree,
+            target,
+            Kind::SymbolReport,
+        );
         let crates = vec!["core".to_owned()]; // currently, only core is certified
         std_cargo(builder, target, &mut cargo, &crates);
         cargo.env("RUSTC_REAL", symbol_report);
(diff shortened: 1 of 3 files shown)


src/bootstrap/src/ferrocene/run.rs
```diff
@@ -145,8 +145,14 @@ impl Step for CertifiedCoreSymbols {
...
         // c.f. check::std
-        let mut cargo =
-            Cargo::new(builder, build_compiler, Mode::Std, SourceType::InTree, target, Kind::Check);
+        let mut cargo = Cargo::new(
+            builder,
+            build_compiler,
+            Mode::Std,
+            SourceType::InTree,
+            target,
+            Kind::SymbolReport,
```
jyn514, line 154: to be more specific about what the bug was: if we
1. ran `x check core`
2. ran `x run symbol-report`
3. ran `x test certified-core-symbols`

then Cargo would build from scratch in 1, *not* rebuild in 2, and 3 would error because it couldn't find the report.

Ana and i went on a long goose chase: to the best of my understanding, this bug happens because Cargo caches whether rustc changes by *path*, and bootstrap sets `RUSTC=target/debug/rustc` (i.e. its own rustc shim), and then chooses after that whether to dispatch to rustc or to symbol-report. So Cargo thought that `rustc` and `symbol-report` were the same binary and didn't know to rebuild.

The fix here is to manually append to Cargo's `-C metadata` hash, which forces it to rebuild even if it doesn't think it needs to.

jyn514 (approved):

Hoverbear:
Holding off on merge due to https://github.com/ferrocene/ferrocene/pull/2389

Urhengulas:
bors merge

PR #2388, opened 2026-06-02
Split platform manuals

plaindocs:
This PR splits out the following docs from safety manual, user manual and qualification docs:

- QNX7
- QNX8
- RHIVOS

Into a single manual per target that contains the relevant content from the 3 documents. I've made the minimal changes required for those to make sense.

Still to do:

- [x] update tests so CI passes 
- [x] update index.html to reference the new targets?
- [x] RHIVOS qualified targets?
- [x] update doc list

Review:

- [ ] be great if someone familiar with the test outcomes could make sure that all looks ok
- [ ] be great if someone could eyeball the new docs with fresh eyes and sure it all seems ok

### DIFF
--- a/ferrocene/doc/index/index.html
+++ b/ferrocene/doc/index/index.html
@@ -57,15 +57,15 @@
                 <div class="docs">
                     <a href="targets/rhivos-manual/index.html">
                         <h3>RHIVOS Manual</h3>
-                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                        <p>User manual for Red Hat® In-Vehicle Operating System target.</p>
                     </a>
                     <a href="targets/qnx7-manual/index.html">
                         <h3>QNX7 Manual</h3>
-                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                        <p>User manual for QNX® SDP 7 target.</p>
                     </a>
                     <a href="targets/qnx8-manual/index.html">
                         <h3>QNX8 Manual</h3>
-                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                        <p>User manual for QNX® SDP 8 target.</p>
                     </a>
                 </div>
             </section>
--- a/ferrocene/doc/index/index.html
+++ b/ferrocene/doc/index/index.html
@@ -51,6 +51,25 @@
                     </a>
                 </div>
             </section>
+
+            <section>
+                <h2>Target specific documentation</h2>
+                <div class="docs">
+                    <a href="targets/rhivos-manual/index.html">
+                        <h3>RHIVOS Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                    </a>
+                    <a href="targets/qnx7-manual/index.html">
+                        <h3>QNX7 Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                    </a>
+                    <a href="targets/qnx8-manual/index.html">
+                        <h3>QNX8 Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                    </a>
+                </div>
+            </section>
+
             <section id="ferrocene-qualification-material">
                 <h2>Ferrocene Qualification Material</h2>
                 <div class="docs">
--- a/ferrocene/doc/qnx8-manual/src/rustc/aarch64-unknown-qnx.rst
+++ b/ferrocene/doc/qnx8-manual/src/rustc/aarch64-unknown-qnx.rst
@@ -1,6 +1,6 @@
 .. SPDX-License-Identifier: MIT OR Apache-2.0
    SPDX-FileCopyrightText: The Ferrocene Developers
 
-.. render-outcomes-template:: templates/tests.jinja2
+.. render-outcomes-template:: ../../qualification-report/src/templates/tests.jinja2
    :host: x86_64-unknown-linux-gnu
    :target: aarch64-unknown-qnx
--- a/src/bootstrap/src/core/builder/cli_paths/snapshots/x_doc.snap
+++ b/src/bootstrap/src/core/builder/cli_paths/snapshots/x_doc.snap
@@ -74,6 +74,15 @@ expression: doc
 [Doc] bootstrap::ferrocene::doc::ReleaseNotes
     targets: [aarch64-unknown-linux-gnu]
     - Set({doc::ferrocene/doc/release-notes})
+[Doc] bootstrap::ferrocene::doc::QNX7Manual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/qnx7-manual})
+[Doc] bootstrap::ferrocene::doc::QNX8Manual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/qnx8-manual})
+[Doc] bootstrap::ferrocene::doc::RHIVOSManual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/rhivos-manual})
 [Doc] bootstrap::ferrocene::doc::DocumentList
     targets: [aarch64-unknown-linux-gnu]
     - Set({doc::ferrocene/doc/document-list})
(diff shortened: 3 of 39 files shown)


ferrocene/doc/qnx7-manual/exts/ferrocene_test_outcomes/outcomes.py
jyn514, line 1: please refactor all these python files out to a helper library rather than duplicating them.

jyn514 (changes requested):

ferrocene/doc/qnx7-manual/src/templates/tests.jinja2
jyn514, line 1: can we reuse this too rather than duplicating it? same question for conf.py.

i believe if we have a unified sphinx project that would also work, and we could tell sphinx "only build this subset of the docs". but i haven't verified that.

jyn514 (changes requested):

ferrocene/doc/qnx7-manual/src/templates/tests.jinja2
plaindocs: I looked briefly at that, and I suspect it's hard coded in the ext script. We could just symlink it? :-p 

All the manuals have individual `conf.py`, and need them when being built individually. I'll happily revisit that when I look at the unified sphinx build.

jyn514: ok, duplicating conf.py isn't a big deal. i'd like to reuse tests.jinja2 if we can though, it's rather complicated.

plaindocs: Thanks for encouraging me to do the right thing here :-D

jyn514:
remaining changes look ok to me but there's a bunch of rebase conflicts.

jyn514 (commented):
@plaindocs you need to run `cargo insta review --manifest-path src/bootstrap/Cargo.toml`; see https://ops-handbook.ferrocene.dev/pr-rotation/upstream-pull-basic-checks.html. 

can you confirm that `x doc ferrocene/doc --exclude ferrocene/doc/rhivos-manual` works and does what you expect?

about the index: since Sphinx allows python extensions, you should be able to add conditional logic here. i haven't looked at the code in detail but you can search for `.py` files in `ferrocene/doc` and that should give you an idea. let me know if you need more guidance or get stuck.

src/bootstrap/src/core/builder/mod.rs
```diff
@@ -995,6 +995,10 @@ impl<'a> Builder<'a> {
                 crate::ferrocene::doc::Specification,
                 crate::ferrocene::doc::UserManual,
                 crate::ferrocene::doc::ReleaseNotes,
+                //Targets
```
Urhengulas, line 998: What does this comment mean?

plaindocs: This is just the 'category' of documents, like the other groups in this file.

plaindocs:
@jyn514 makes sense. Will check `--exclude` next. 

The main doc index is https://github.com/ferrocene/ferrocene/tree/main/ferrocene/doc/index which is just html, so doesn't support sphinx conditionals.

plaindocs:
@jyn514 I don't think `--exclude` works:

```sh
➜  ./x doc ferrocene/doc --exclude ferrocene/doc/rhivos-manual

...

Building Ferrocene documentation index (aarch64-apple-darwin)
Skipping Set({doc::ferrocene/doc/rhivos-manual}) because it is excluded
Build completed successfully in 0:01:14

ferrocene on  split-targets via  v26.0.0 via 🐍 v3.14.5 via 🦀 v1.96.1 took 1m14s
➜ ls build/aarch64-apple-darwin/doc/targets/rhivos-manual
_sources        document-id.txt objects.inv     rustc           search.html     signature
_static         index.html      overview.html   scope.html      searchindex.js  targets
```

My understanding is I need to use the 'source' doc in exclude? 

I'm also not sure if building with `--exclude` would delete it if it was there (I lowkey expect not), but I tested for that.

jyn514:
@plaindocs it's being pull in recursively by DocumentList:

afbeelding

> The main doc index is https://github.com/ferrocene/ferrocene/tree/main/ferrocene/doc/index which is just html, so doesn't support sphinx conditionals.

ok. i would suggest doing something like this:
1. Turn the HTML file into a .jinja file. There's docs here: https://jinja.palletsprojects.com/en/stable/templates/
2. Extend `doc::Index::run` to render Jinja, either in Rust or in a python wrapper. If you go the python route you can use ferrocene/doc/qualification-report/exts/ferrocene_test_outcomes/render_template.py  as a reference.
3. Extend `doc::Index` to carry more data in the struct along with the `target` — i think we probably want a list of all targets it links to, since i imagine we'll want to have more conditional targets in the future.
4. Add a new `build.exclude-targets = []` list to bootstrap, read from bootstrap.toml. Extend `doc::Index::make_run` to use it to determine the targets that are rendered.

src/bootstrap/src/core/builder/cli_paths/snapshots/x_doc.snap
```diff
@@ -74,6 +74,15 @@ expression: doc
...
     - Set({doc::ferrocene/doc/release-notes})
+[Doc] bootstrap::ferrocene::doc::QNX7Manual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/qnx7-manual})
+[Doc] bootstrap::ferrocene::doc::QNX8Manual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/qnx8-manual})
+[Doc] bootstrap::ferrocene::doc::RHIVOSManual
+    targets: [aarch64-unknown-linux-gnu]
+    - Set({doc::ferrocene/doc/rhivos-manual})
```
jyn514, line 85: noting explicitly that the new target docs are built by default, so they should automatically end up in the release tarball.

ferrocene/doc/qnx8-manual/src/rustc/aarch64-unknown-qnx.rst
```diff
@@ -1,6 +1,6 @@
 .. SPDX-License-Identifier: MIT OR Apache-2.0
    SPDX-FileCopyrightText: The Ferrocene Developers
 
-.. render-outcomes-template:: templates/tests.jinja2
+.. render-outcomes-template:: ../../qualification-report/src/templates/tests.jinja2
    :host: x86_64-unknown-linux-gnu
```
tshepang, line 5: is this correct... thought the host was aarch64

jyn514 (approved):
this looks ok for now. i've tested that `x doc ferrocene/doc/index` and `x ferrocene/doc/qnx8` both work. we're planning to build the QNX8 docs separately as a one-off for 26.08, and worry about the long-term fix later.

the one thing i'm not sure about is whether test outcomes are going to render properly. i'm a little hesitant to merge this until we know, although i guess we can always just merge to main without backporting to 26.08. yeah ok, let's do that.

bors merge

rendered: Image

ferrocene/doc/qnx8-manual/src/rustc/aarch64-unknown-qnx.rst
plaindocs: I believe that's correct, according to https://public-docs.ferrocene.dev/main/qualification/report/rustc/index.html

tshepang:
bors cancel

squashing...

ferrocene/doc/index/index.html
```diff
@@ -51,6 +51,25 @@ <h3>Release Notes</h3>
...
+                        <h3>RHIVOS Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                    </a>
+                    <a href="targets/qnx7-manual/index.html">
+                        <h3>QNX7 Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
+                    </a>
+                    <a href="targets/qnx8-manual/index.html">
+                        <h3>QNX8 Manual</h3>
+                        <p>An easy-to-read, how-to manual for operating Ferrocene in a qualified environment.</p>
```
tshepang: a nit to avoid repetitive-ness
```suggestion
                        <h3>RHIVOS Manual</h3>
                        <p>User manual for Red Hat® In-Vehicle Operating System target.</p>
                    </a>
                    <a href="targets/qnx7-manual/index.html">
                        <h3>QNX7 Manual</h3>
                        <p>User manual for QNX® SDP 7 target.</p>
                    </a>
                    <a href="targets/qnx8-manual/index.html">
                        <h3>QNX8 Manual</h3>
                        <p>User manual for QNX® SDP 8 target.</p>
```

Hoverbear (approved):
bors merge

PR #2395, opened 2026-06-10
avoid panic when checking symbol-report on not-x86

tshepang:
We hardcode x86_64-unknown-linux-gnu as target when we run symbol report (`x test ferrocene/doc/symbol-report.csv`), and that resulted in a panic due to calling a function that checks for a linker of that target, a linker that will likely not exist outside of that target.


tshepang:
made part of #2402

PR #2410, opened 2026-06-19
Document the `rustc --check-cfg` option

Urhengulas:


jyn514 (approved):
this broadly seems ok to me. i'm not sure why we have a our own user manually fully separate from upstream's ... but that's a pre-existing issue.

bors merge

jyn514:
ah, it wants you to add a test for the traceability matrix.

Urhengulas:
Noting that I would like to have this in 26.08.0 because I am using this option in a certified project

jyn514 (approved):
bors merge

PR #2411, opened 2026-06-22
Bundle compiler docs into dist tarball

fs-rachel:
For historical reasons, upstream generates the compiler docs in their own directory (`build/<target>/compiler-doc`), separate from the other documentation (in `build/<target>/doc`). Our dist bundle was being built from the latter, so didn't include the compiler docs.

Fix this by adding the option to remap directories when generating our dist bundle, and use that to save the compiler docs under `<prefix>/compiler`.

### DIFF
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -12,6 +12,7 @@ use std::rc::Rc;
 use serde_json::json;
 
 use crate::builder::{Builder, RunConfig, ShouldRun, Step};
+use crate::core::build_steps::doc::Rustc;
 use crate::core::build_steps::run::GenerateCopyright;
 use crate::core::config::TargetSelection;
 use crate::ferrocene::code_coverage::CoverageOutcomesDir;
@@ -50,6 +51,9 @@ impl Step for Docs {
         builder.run_default_doc_steps();
         let doc_out = builder.doc_out(self.target);
 
+        builder.ensure(Rustc::for_stage(builder, builder.top_stage, self.target));
+        let compiler_doc_out = builder.compiler_doc_out(self.target);
+
         // Build the documentation for certified crates
         //
         // NOTE: must be called before .add_directory, since it places the
@@ -62,7 +66,8 @@ impl Step for Docs {
         }
 
         let mut subsetter = Subsetter::new(builder, "ferrocene-docs", "share/doc/ferrocene/html");
-        subsetter.add_directory(&doc_out, &doc_out);
+        subsetter.add_directory(&doc_out, &Path::new(""), &doc_out);
+        subsetter.add_directory(&compiler_doc_out, &Path::new("rustc-docs"), &compiler_doc_out);
         subsetter.add_copyright_files(&builder);
 
         subsetter.into_tarballs().map(|tarball| tarball.generate()).collect()
@@ -160,10 +165,10 @@ impl Step for SourceTarball {
 
         // Copy raw source files
         for item in DIRS {
-            subsetter.add_directory(&builder.src, &builder.src.join(item));
+            subsetter.add_directory(&builder.src, &Path::new(""), &builder.src.join(item));
         }
         for item in FILES {
-            subsetter.add_file(&builder.src, &builder.src.join(item));
+            subsetter.add_file(&builder.src, &Path::new(""), &builder.src.join(item));
         }
         subsetter.add_copyright_files(&builder);
 
@@ -260,7 +265,10 @@ impl<'a> Subsetter<'a> {
         }
     }
 
-    fn add_directory(&mut self, root: &Path, path: &Path) {
+    // Add files from under `path`, remapping a specified directory in the filesystem to another
+    // in the tarball along the way.
+    // Note: `path` must be under `root_in_filesystem`.
+    fn add_directory(&mut self, root_in_filesystem: &Path, root_in_tarball: &Path, path: &Path) {
         let old_subset = self.directory_subset.clone();
 
         let subset_file = path.join("ferrocene-subset");
@@ -273,9 +281,9 @@ impl<'a> Subsetter<'a> {
         for entry in std::fs::read_dir(path).unwrap() {
             let path = entry.as_ref().unwrap().path();
             if path.is_file() {
-                self.add_file(root, &path);
+                self.add_file(root_in_filesystem, root_in_tarball, &path);
             } else if path.is_dir() {
-                self.add_directory(root, &path);
+                self.add_directory(root_in_filesystem, root_in_tarball, &path);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/bootstrap/src/lib.rs
```diff
@@ -1058,7 +1058,9 @@ impl Build {
 
     /// Output directory for all documentation for a target
     fn compiler_doc_out(&self, target: TargetSelection) -> PathBuf {
-        self.out.join(target).join("compiler-doc")
+        // Ferrocene modification: Unify compiler docs with everything else, rather than
+        // them being in a separate compiler-doc/ directory
+        self.out.join(target).join("doc").join("compiler")
```
jyn514: hm, i'm not sure i follow how this helps — can you connect the dots for me? is there some Step that's copying everything in `doc/` blindly? i would expect this to instead be keyed off `Step::is_default` or something like that.

partially related: i believe the reason upstream keeps this separate is because otherwise it causes issues for rustup: https://github.com/rust-lang/rust/issues/75833
oh wait wtf that was fixed this year??? wild, it had been broken for like 8 years. this might be fine then.

fs-rachel: I haven't managed to find the code which actually uploads to https://public-docs.ferrocene.dev/ , but as far as I can tell there's a 1:1 mapping of `doc/... -> https://public-docs.ferrocene.dev/main/...`. Which means the compiler docs are not currently uploaded, because they aren't under `doc/`

fs-rachel: (I checked and neither `https://public-docs.ferrocene.dev/compiler-doc/` nor `https://public-docs.ferrocene.dev/main/compiler-doc/` exists, for example)

jyn514: take a look around `impl Step for Docs` in `src/bootstrap/src/ferrocene/dist.rs`

fs-rachel: Aha, yes, the docs tarball just copies what's in `doc/` at the moment:

```
        let doc_out = builder.doc_out(self.target);
        // ...
        let mut subsetter = Subsetter::new(builder, "ferrocene-docs", "share/doc/ferrocene/html");
        subsetter.add_directory(&doc_out, &doc_out);
        subsetter.add_copyright_files(&builder);
```

Currently `add_directory` doesn't support full path remapping, only stripping a common initial prefix. But I could add that if it would be preferable over moving the compiler docs output

jyn514: yeah, let's prefer to change our code rather than upstream's code. if we change upstream we have to deal with the merge conflicts going forward.

src/bootstrap/src/ferrocene/dist.rs
```diff
@@ -62,7 +63,8 @@ impl Step for Docs {
         }
 
         let mut subsetter = Subsetter::new(builder, "ferrocene-docs", "share/doc/ferrocene/html");
-        subsetter.add_directory(&doc_out, &doc_out);
+        subsetter.add_directory(&doc_out, &Path::new(""), &doc_out);
+        subsetter.add_directory(&compiler_doc_out, &Path::new("compiler"), &compiler_doc_out);
```
jyn514: why do we need both these flags? can we use `add_directory(compiler_doc_out.join("compiler"))` instead?

fs-rachel: `compiler_doc_out` is the path on the filesystem where the compiler docs were generated, whereas "compiler" is the path inside the tarball where we want to put them, so they can't be combined

There is a bit of redundancy in that the first and third arguments will always be the same for *external* callers, but they can differ when `add_directory` recursively calls itself. But I wouldn't be averse to splitting that up so that the external semantics are more clearly "take this directory from the filesystem and put it here in the tarball"

fs-rachel: just to clear up a potential bit of confusion, the new version of this PR returns to upstream's split where the compiler docs are in `compiler-doc` *next to* `doc`, not inside it

jyn514: this is so unfortunate. ok, if that's what it takes then let's do it.

jyn514 (approved):
bors merge

jyn514:
it's weird that `--set build.compiler-docs` also builds `compiletest`, but at most it does a little extra work so i'm not going to worry about it.

bors merge

jyn514:
some wild flakes this time around.

bors retry

Windows:
```
Downloading package from source 'https://community.chocolatey.org/api/v2/'
ninja not installed. An error occurred during installation:
 Value cannot be null.
Parameter name: packageStream
ninja package files install failed with exit code 1. Performing other installation steps.
```
`ferrocene/ci/scripts/persist-between-jobs.sh upload ./build` (this one i knew about):
```
Too long with no output (exceeded 10m0s): context deadline exceeded
```
generate-installer fails fetching KMS (this one i knew about):
```
Dist rust-std-x86_64-pc-nto-qnx710-038a84ba2
Error: failed to generate installer

Caused by:
    0: failed to retrieve the public key from AWS KMS
    1: dispatch failure
    2: other
    3: credentials provider timed out after 5 seconds
Command `/home/ci/project/build/x86_64-unknown-linux-gnu/stage1-tools-bin/generate-tarball generate --image-dir /home/ci/project/build/tmp/tarball/rust-std/x86_64-pc-nto-qnx710/image --component-name=rust-std-x86_64-pc-nto-qnx710 --rel-manifest-dir=rustlib --legacy-manifest-dirs=rustlib,cargo --product-name=Rust '--success-message=rust-std installed.' --package-name=rust-std-x86_64-pc-nto-qnx710-038a84ba2 --non-installed-overlay /home/ci/project/build/tmp/tarball/rust-std/x86_64-pc-nto-qnx710/overlay --output-dir /home/ci/project/build/dist --ferrocene-component rust-std-x86_64-pc-nto-qnx710 --ferrocene-commit-sha 038a84ba2cc35fe80da9e0f3a72be797e83b28bc --ferrocene-signing-kms-key-arn arn:aws:kms:us-east-1:886866542769:key/cfbd0673-04d8-4368-b09f-56998ede9b96 --ferrocene-managed-prefix lib/rustlib/x86_64-pc-nto-qnx710/lib --work-dir /home/ci/project/build/tmp/tarball/rust-std/x86_64-pc-nto-qnx710 --compression-formats xz --compression-profile balanced --override-file-mtime 1782304092` failed with exit code 1
Created at: src/bootstrap/src/core/build_steps/tool.rs:1630:23
Executed at: src/bootstrap/src/utils/tarball.rs:471:13
```

fs-rachel:
Looks like the build flaked on `ferrocene/ci/scripts/persist-between-jobs.sh upload ./build` again. Let's see if I have magic powers:

bors retry

jyn514:
Felix is making a PR soon to reduce the number of flakes, might make sense to just delay this until then. it's not urgent.

jyn514:
https://github.com/ferrocene/ferrocene/pull/2420#issuecomment-4830612670 is merged.

bors merge

fs-rachel:
Today's lesson: the "default" steps are not in fact a static set, but depend on the builder config. So we need to `ensure` the compiler docs get built, we can't rely on them actually being on by default.

Also while testing this locally, I discovered that upstream's `./x dist rustc-docs` does basically the same thing we're doing here (except in that case the underlying compiler docs *are* magically part of the default set!). And that remaps the compiler docs as if they were written to `doc/rustc-docs`. So for consistency, I'll change this patch to use the same name.

fs-rachel:
bors retry

PR #2413, opened 2026-06-23
Self-Hosted VM Runners

Xylakant:
This adds support for running many of our linux-vm based tasks on self-hosted runners

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1393,7 +1393,7 @@ workflows:
           name: x86_64-qnx-test-compiletest
           job: qnx:compiletest-no-only-hosts
           test-variant: "2021"
-          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 8-core
+          resource-class: ferrocene/vm-ubuntu-20-amd64-large # 8-core
           requires:
             - x86_64-linux-build
             - build-docker--ubuntu-24--x86_64
@@ -1412,7 +1412,7 @@ workflows:
           name: x86_64-qnx8-test-compiletest
           job: qnx:compiletest-no-only-hosts
           test-variant: "2021"
-          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 8-core
+          resource-class: ferrocene/vm-ubuntu-20-amd64-large # 8-core
           requires:
             - x86_64-linux-build
             - build-docker--ubuntu-24--x86_64
@@ -1531,7 +1531,7 @@ workflows:
           name: aarch64-qnx-test-compiletest
           job: qnx:compiletest-no-only-hosts
           test-variant: "2021"
-          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 8-core
+          resource-class: ferrocene/vm-ubuntu-20-amd64-large # 8-core
           requires:
             - x86_64-linux-build
             - build-docker--ubuntu-24--x86_64
@@ -1550,7 +1550,7 @@ workflows:
           name: aarch64-qnx8-test-compiletest
           job: qnx:compiletest-no-only-hosts
           test-variant: "2021"
-          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 8-core
+          resource-class: ferrocene/vm-ubuntu-20-amd64-large # 8-core
           requires:
             - x86_64-linux-build
             - build-docker--ubuntu-24--x86_64
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1295,6 +1295,7 @@ workflows:
           resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 4-core
           requires:
             - x86_64-linux-build
+            - aarch64-none-test-library-compiletest
 
       - facade-generic-test:
           name: aarch64r82-none-test-library-compiletest
@@ -1393,6 +1394,7 @@ workflows:
           resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 4-core
           requires:
             - x86_64-linux-build
+            - thumbv7em-eabihf-test-library-compiletest
 
       - facade-generic-test:
           name: thumbv7em-eabi-test-target-cpu-library-compiletest
@@ -1403,6 +1405,7 @@ workflows:
           resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 4-core
           requires:
             - x86_64-linux-build
+            - thumbv7em-eabi-test-library-compiletest
 
       - aarch64-qnx-generic-test-vm:
           name: aarch64-qnx-test-library-compiletest
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1207,14 +1207,10 @@ workflows:
             - x86_64-linux-test-library
             - x86_64-linux-compiletest
             - x86_64-linux-library-coverage
-            - x86_64-qnx-test-library
-            - x86_64-qnx-test-compiletest
-            - x86_64-qnx8-test-library
-            - x86_64-qnx8-test-compiletest
-            - aarch64-qnx-test-library
-            - aarch64-qnx-test-compiletest
-            - aarch64-qnx8-test-library
-            - aarch64-qnx8-test-compiletest
+            - x86_64-qnx-test-library-compiletest
+            - x86_64-qnx8-test-library-compiletest
+            - aarch64-qnx-test-library-compiletest
+            - aarch64-qnx8-test-library-compiletest
             - aarch64-none-test-library-compiletest
             - aarch64-none-test-target-cpu-library-compiletest
             - aarch64r82-none-test-library-compiletest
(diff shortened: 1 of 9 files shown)


.circleci/workflows.yml
```diff
@@ -1840,17 +1709,17 @@ commands:
...
-          command: |
-            sudo mkdir -p /etc/docker
-            cat \<<EOF | sudo tee /etc/docker/daemon.json
-            {
-              "ipv6": true,
-              "fixed-cidr-v6": "2001:db8:1::/64"
-            }
-            EOF
-            sudo service docker restart
+      # - run:
```
Hoverbear: You can delete this, I think.

Hoverbear (commented):
This looks overall good and the direction we've talked about.

Xylakant:
bors try

Xylakant:
bors try

Hoverbear:
(Review note: We need to check the test outcomes number)

.circleci/workflows.yml
```diff
@@ -354,7 +360,7 @@ jobs:
   x86_64-qnx-generic-test-vm:
     executor: linux-vm
     parameters:
-      job:
+      tasks:
```
jyn514, line 363: c.f. https://support.circleci.com/hc/en-us/articles/21249182109211-Understanding-Jobs-and-Tasks-in-CircleCI

.circleci/workflows.yml
```diff
@@ -831,20 +843,21 @@ jobs:
...
+          override_installed: true
       - run:
           name: Restore files from the x86_64-linux-build job
           command: ferrocene/ci/scripts/persist-between-jobs.sh restore x86_64-linux-build
       - run:
           name: Pull the Docker image used for CI
           command: |
-            aws ecr get-login-password --region us-east-1 \
-              | docker login -u AWS --password-stdin << pipeline.parameters.docker-repository-url--ci-docker-images >>
-            docker pull << pipeline.parameters.docker-repository-url--ci-docker-images >>:<< pipeline.parameters.docker-image-tag--x86_64--ubuntu-20 >>
```
jyn514, line 847: is the login now done in aws-cli/install?

.circleci/workflows.yml
```diff
@@ -1178,32 +1205,24 @@ workflows:
...
+            - aarch64-qnx8-test-library-compiletest
+            - aarch64-none-test-library-compiletest
+            - aarch64-none-test-target-cpu-library-compiletest
+            - aarch64r82-none-test-library-compiletest
+            - aarch64v8r-none-test-library-compiletest
+            - armv7r-eabihf-test-library-compiletest
+            - thumbv7em-eabihf-test-library-compiletest
+            - thumbv7em-eabihf-test-target-cpu-library-compiletest
+            - thumbv7em-eabi-test-library-compiletest
+            - thumbv7em-eabi-test-target-cpu-library-compiletest
```
jyn514, line 1220: what's the reason for putting -library-compiletest in all these names? can we just call them `-test`?

.circleci/workflows.yml
```diff
@@ -1840,25 +1729,12 @@ commands:
...
-          name: Enable IPv6 for Docker
-          command: |
-            sudo mkdir -p /etc/docker
-            cat \<<EOF | sudo tee /etc/docker/daemon.json
-            {
-              "ipv6": true,
-              "fixed-cidr-v6": "2001:db8:1::/64"
-            }
-            EOF
-            sudo service docker restart
```
jyn514, line 1853: i see this disabled, but never re-enabled. in https://github.com/ferrocene/ferrocene/pull/2413/changes/c07ae2df6f73114dd563a601ebf6f3b8e48959cd you say "ipv6 for docker is now baked into the image" — where can i see that change?

ferrocene/ci/scripts/emulated-x86_64-qnx8-test-runner.sh
```diff
@@ -26,8 +26,9 @@ start_vm() {
     # set memory to 11G as `tests/ui/codegen/huge-stacks.rs` requires 5GB+ and has two
     # variants which can run in parallel
     qemu-system-x86_64 \
+        -enable-kvm \
+        -cpu host \
         -smp 2 \
-        -cpu max \
```
jyn514, line 30: why this change? did something change on the infra side?

.circleci/workflows.yml
```diff
@@ -1384,7 +1384,7 @@ workflows:
           job: test:library
           test-variant: "2021"
           extra-args: --tests # We do not run cross compiled doctests for this job on this platform, there are segfaults
-          resource-class: ferrocene/vm-ubuntu-amd64-medium # 4-core
+          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 4-core
```
jyn514, line 1387: it's very scary to me that changing a version number changes the *hosting provider*. is there any way we could make that more clear in the name? `ferrocene/heztner-ubuntu-20` or something like that?

.circleci/workflows.yml
```diff
@@ -1221,7 +1221,9 @@ workflows:
             - aarch64v8r-none-test-library-compiletest
             - armv7r-eabihf-test-library-compiletest
             - thumbv7em-eabihf-test-library-compiletest
+            - thumbv7em-eabihf-test-target-cpu-library-compiletest
             - thumbv7em-eabi-test-library-compiletest
+            - thumbv7em-eabi-test-target-cpu-library-compiletest
```
jyn514, line 1226: do we have an automated check that this `requires:` list is kept up to date? how hard would it be to write one?

jyn514 (commented):
the changes here seem broadly good but i found them quite hard to understand. here's my understanding of what's changed:

- we're installing a consistent version of AWS CLI at runtime, rather than depending on the base image.
- IPV6 is now configured in the docker base image, rather than at runtime.
- we've enabled KVM passthrough for almost all the runners.
- we've switched to self-hosted runners for most jobs, excluding the job that actually builds the docker image.
- we've switched from CircleCI "jobs" to "tasks", which allows us to run more of the test suite in parallel.
- we've consolidated some jobs into one runner, which uses less CPU time in exchange for high wall clock time. this is ok because the wall clock time is already dominated by other jobs.
- we're now using our own Docker cache rather than a public one.

does that all sound right? anything i missed?

.circleci/workflows.yml
Xylakant: Ah, no - that's no circleci's jobs and tasks. We decided to re-use the same name. It's the distinction of whether the things get feed trough split-tasks.oy or have been feed through.

.circleci/workflows.yml
Xylakant: It's magically there. (the VM is authenticated by the cluster)

.circleci/workflows.yml
Xylakant: either name works for me - I picked this name because it's now the combination of (library, compiletest) and we have other tests.

.circleci/workflows.yml
Xylakant: it's baked into the VM image, the machine has IPv6 for docker enabled by default, unlike the runners from circleci.

.circleci/workflows.yml
jyn514: ok. can you add a comment saying so?

ferrocene/ci/scripts/emulated-x86_64-qnx8-test-runner.sh
Xylakant: yes, circleci's runners don't support KVM accel. Ours do.

.circleci/workflows.yml
jyn514: that is extremely confusing. could you document this somewhere please?

.circleci/workflows.yml
Xylakant: we can, but this is a remainder of my testing. It's not the change in the version number that triggers the change - it's just that I defined the resource class I used for testing with a name that was unfortunately close to the final name. I'd rather delete the testing resource class.

I would avoid putting the name of the provider in there - the resource class is meant as "this runner provides this capability" and where it runs is largely intended to be opaque. That allows us to switch the hosting provider without affecting CI. From the perspective of the CI system, that change should be transparent.

.circleci/workflows.yml
Xylakant: We don't and I can't tell. It's not something I can answer for you.

.circleci/workflows.yml
Xylakant: I honestly haven't thought up the system. That's about all I know, reverse engineered from the workflow file. I'd accept better names, but the naming here isn't really my thing - it's all baked into split-tasks.py.

Xylakant:
> the changes here seem broadly good but i found them quite hard to understand. here's my understanding of what's changed:
> 
>     * we're installing a consistent version of AWS CLI at runtime, rather than depending on the base image.

this is part of the changeset in https://github.com/ferrocene/ferrocene/pull/2420 on which I based this PR (both already benefit from the change and to exercise it a little more). Annoyingly GH still displays the change, despite the fact that the other change is merge.

> 
>     * IPV6 is now configured in the docker base image, rather than at runtime.

It's not the docker base image, it's the VM that is by default configured to support IPv6 for docker

> 
>     * we've enabled KVM passthrough for almost all the runners.

For all that can benefit from it. The reason we didn't before is that the circleci hosted runners do not support kvm.

> 
>     * we've switched to self-hosted runners for most jobs, excluding the job that actually builds the docker image.

yes.

> 
>     * we've switched from CircleCI "jobs" to "tasks", which allows us to run more of the test suite in parallel.

It's not CircleCI jobs/task - unfortunately `split-tasks.py` uses the same lingo. Before, each CircleCI job would accept a "job" parameter which was fed through `split-tasks.py` to get the actual test tasks to execute. Now, we feed a list of tasks directly (or, specifically, the invocations of split-tasks.py) which allows us to combine multiple `split-tasks.py` jobs. We can now also feed a combination of both explicitly defined tasks and `split-tasks.py` invocations. This works because on the CircleCI side, this is raw string interpolation and the command is later executed, and expansion takes place at that time.

> 
>     * we've consolidated some jobs into one runner, which uses less CPU time in exchange for high wall clock time. this is ok because the wall clock time is already dominated by other jobs.

Yes

> 
>     * we're now using our own Docker cache rather than a public one.

More specifically: Instead of pulling straight from our (private) ECR registry, we have a pull-through cache that sits in the cluster. This reduces the total bandwidth used on the ECR repo.

> 
> 
> does that all sound right? anything i missed?

No, that's pretty much all.

jyn514:
@Xylakant would you mind adding most of those answers as comments so they're easier for me to find next time I have to touch this code? If you're not sure how to phrase it, or you don't know the system well enough to comment, let me know and I'm happy to suggest wording.

Xylakant:
I'll keep this around until I have split out the other parts of this, but do not intend to merge this.

Xylakant:
all changes here have been extracted to other PRs.

PR #2416, opened 2026-06-24
Deduplicate 'core' documentation

fs-rachel:
Previously we generated two versions of the documentation for core: 1) as part of the standard library documentation, without certification annotations 2) for core alone, with certification annotations

These were both linked from the front page of our docs, leading to potential for confusion.

Address this by adding annotations to the upstream standard library docs generator, removing the custom core-only generator, and pointing both front-page links to the same place.


fs-rachel:
Note: The failed checks are due to changes in the sequence of build steps in the dist_* tests. These are also affected by https://github.com/ferrocene/ferrocene/pull/2411 , so I'll rebase this PR and update those tests after that one is fixed

ferrocene/doc/index/index.html
```diff
@@ -79,9 +79,9 @@ <h2>Core library Certification Material</h2>
                         <h3>Core library certification</h3>
                         <p>Phases and activities that describe the certification of a subset of the core library with the ASIL B level in accordance with ISO 26262 standard and the class SIL 2 in accordance with the IEC 61508 standard.</p>
                     </a>
-                    <a href="certification/api-docs/core/index.html">
+                    <a href="core/index.html">
                         <h3>Certified core library API docs</h3>
-                        <p>The API documentation of certified subset of the core library.</p>
+                        <p>Documentation for Rust's core library, with annotations to indicate which functions are certified</p>
```
jyn514: no longer just `core`, right?
```suggestion
                        <p>Documentation for Rust's standard library, with annotations to indicate which functions are certified</p>
```

src/bootstrap/src/ferrocene/doc/mod.rs
```diff
@@ -583,9 +581,6 @@ macro_rules! sphinx_books {
                     });
                 )*
 
-                // Generate the API docs for the certified crates
-                builder.ensure(CertifiedApiDocs {target: self.target});
```
jyn514, line 587: should this be replaced with `builder.ensure(doc::Std { target })`?

jyn514 (commented):
seems good, like you mentioned i think CI will fail until the upstream-pull is merged.

ferrocene/doc/index/index.html
fs-rachel: Am I right in thinking it's only core *for now*, we just want to expand it later?

There's a more proper list of crates lower down the page, so if we want to list more than just core it's probably best to just take this link out entirely. That or move the "list of crates" page which would normally be generated as `index.html` to somewhere where it doesn't get splatted by our custom front page, and link to that.

src/bootstrap/src/ferrocene/doc/mod.rs
fs-rachel: That's handled by `builder.run_default_doc_steps();` a few lines above this

fs-rachel: I could add that anyway if you want to be extra-certain it'll build the docs, shouldn't hurt anything

jyn514: yeah, it's fine to sprinkle `ensure()` around liberally.

ferrocene/doc/index/index.html
jyn514: it's only core that's *certified* for now, but since we've combined the documentation, it needs to be a full replacement for the thing we're deleting. so we need to build all the standard library docs.

jyn514: "standard library" is close enough here, that's the generic word for "everything in `./library` that's public".

fs-rachel: What I'm trying to say is that, if we make that change, we should also change where the link points to. Maybe it should just point further down the same page, to the "Library APIs" section?

jyn514: right, i see, we're in the "certification materials" section. making it point to the "Library APIs" section seems reasonable 👍

src/bootstrap/src/ferrocene/doc/mod.rs
fs-rachel: Done

ferrocene/doc/index/index.html
fs-rachel: I've had a go at that rewrite + adding a little more clarification to the library APIs section: https://github.com/ferrocene/ferrocene/pull/2416/changes/90e2a6efa1000f27faace8f8b7bf2508dca550b9

I'm not sure if this flows as well as before though, considering the section this text is in is currently core-specific. Feels like it needs some more wide-ranging rewrites to make sense once we expand beyond just core. Please can you have a look at the generated front page and let me know what you think?

(I've deliberately put this in its own commit so you can compare the output before/after that commit)

jyn514: i think this looks fine. if you want it to be even nicer you could make it a deep link through to /core/index.html.

rendered:
image

jyn514 (approved):
lgtm. i think you mentioned we're waiting on https://github.com/ferrocene/ferrocene/pull/2411, i'll review that next.

plaindocs:
> Address this by adding annotations to the upstream standard library docs generator, removing the custom core-only generator, and pointing both front-page links to the same place.

sounds great

fs-rachel:
Now that #2411 has merged, I've rebased this and squashed all the changes into one commit

jyn514 (approved):
this is awesome, thanks again Rachel :)

bors merge

tshepang:
flaky ci...
bors retry

fs-rachel:
oops, wrong force-push command :\

fs-rachel:
bors retry

jyn514:
bors merge

fs-rachel:
bors retry

jyn514:
(failures are from QNX license server issues)

PR #2418, opened 2026-06-25
QNX target renaming

jonathanpallant:
Testing out QNX target renaming. There's an equivalent of this PR that should go upstream first - this is just to test some things.

- `aarch64-unknown-nto-qnx800` becomes simply `aarch64-unknown-qnx`
- `x86_64-pc-nto-qnx800` becomes simply `x86_64-pc-qnx`
- references to QNX OS and QNX Neutrino RTOS are replaced with QNX SDP (Software Development Platform) for uniformity
- various `nto_qnx` modules are renamed to `qnx_sdp` to match the above

The new target names are more consistent with those used by the QNX SDP 8.0 toolchain, and reflect a level on ongoing API stability similar to that for macOS and Linux (which simply have minimum supported versions record for their targets). The new target names also match what the copy of `clang` shipped in the QNX Desktop Image (part of QNX SDP 8.0) self reports as (that is, `clang` running self-hosted *on* QNX).


japaric:
bors try

ferrocene/doc/user-manual/src/targets/aarch64-unknown-qnx.rst
```diff
@@ -1,12 +1,12 @@
...
 
-.. _aarch64-unknown-nto-qnx800:
+.. _aarch64-unknown-qnx:
 
-:target:`aarch64-unknown-nto-qnx800`
+:target:`aarch64-unknown-qnx`
 ====================================
 
-The ``aarch64-unknown-nto-qnx800`` Ferrocene target provides support for QNX on
+The ``aarch64-unknown-qnx`` Ferrocene target provides support for QNX on
```
jonathanpallant: Should we note that this is only for QNX SDP 8.0 and higher? And that the supported baseline version may change in future releases?

jonathanpallant: And that we only test on 8.0.5?

ferrocene/doc/user-manual/src/targets/x86_64-pc-qnx.rst
```diff
@@ -1,12 +1,12 @@
...
 
-.. _x86_64-pc-nto-qnx800:
+.. _x86_64-pc-qnx:
 
-:target:`x86_64-pc-nto-qnx800`
+:target:`x86_64-pc-qnx`
 ==============================
 
-The ``x86_64-pc-nto-qnx800`` Ferrocene target provides support for QNX on
+The ``x86_64-pc-qnx`` Ferrocene target provides support for QNX on
```
jonathanpallant: Same comments as for the aarch64-unknown-qnx target page.

ferrocene/doc/user-manual/src/targets/aarch64-unknown-qnx.rst
jyn514: i would take guidance from our other targets here. here's an example: https://public-docs.ferrocene.dev/main/user-manual/targets/aarch64-rhivos2-linux-gnu.html

> The aarch64-rhivos2-linux-gnu Ferrocene target provides support for Red Hat In-Vehicle Operating System 2 (RHIVOS 2) on aarch64 using glibc 2.31 or higher.
> While this target is technically a full linux and capable of hosting a compiler, the target is only qualified if cross-compiled from RHEL 10 on aarch64 in the version specified by the RHIVOS 2 documentation. This requirement stems from the RHIVOS 2 assumptions of use. The host compiler is Armv8-A Linux (glibc).

so yes, seems good to document minimum versions and the hosts we test on.

jonathanpallant: OK, I've noted "QNX SDP 8.0" in most places, and left the notes lower down which point out *precisely* which version of QNX SDP 8.0 we qualify against.

ferrocene/doc/internal-procedures/src/partners/qnx.rst
```diff
@@ -361,6 +361,7 @@ Create a deployment containing Linux and Windows toolchains:
             -installPackage com.qnx.qnx800.host.linux.x86_64=$QNX_LINUX_HOST_VERSION \
             -installPackage com.qnx.qnx800.target.qemuvirt/0.2.1.00087T202507241124L \
             -installPackage com.qnx.qnx800.target.driver.virtio.devc/0.1.1.00011T202411230100L \
+            -installPackage com.qnx.qnx800.patchset805/8.0.5.00390T202606231321L \
```
jonathanpallant, line 364: If we install this patch set, do the notes about which version we qualified need changing on the target pages as well?

ping @japaric

jonathanpallant:
bors try

jyn514:
i believe we've decided this should go upstream first before we merge it here. i'm going to unsubscribe from notifications, ping me if you need anything.

jonathanpallant:
Yes.

Upstream PR: https://github.com/rust-lang/rust/pull/158449

jyn514:
closing in favor of https://github.com/ferrocene/ferrocene/pull/2439

PR #2421, opened 2026-06-26
Update Python dependencies of Ferrocene docs

fs-rachel:
The previous pinned versions were incompatible with Python 3.14+ due to a transitive dependency on pyo3-ffi v0.23.3, which only supports up to Python 3.13.

Fix this by updating the dependencies, and bump the pinned Python version to 3.14.

Also add a rule telling uv to not install packages newer than 7 days old, as a small mitigation against upstream compromises


ferrocene/doc/pyproject.toml
jyn514, line 7: i suspect that some of the upgraded dependencies will no longer run with python3.10. might be worth trying.

jyn514 (commented):
this seems reasonable. the new `exclude-newer` thing is *really* nice, i didn't know uv supported that.

i'm not sure how to review the mass upgrade to the dependencies. could you maybe post a small command i can run on the old lockfile that replicates the change? not anything complicated, just `uv update` or whatever uv calls it.

fs-rachel:
> i'm not sure how to review the mass upgrade to the dependencies. could you maybe post a small command i can run on the old lockfile that replicates the change? not anything complicated, just `uv update` or whatever uv calls it.

The command is `uv lock --upgrade`. Note that *only* updates the lock file, it doesn't actually use the updated dependencies until you also run `uv sync`

ferrocene/doc/pyproject.toml
fs-rachel: It looks like uv took care of that automatically, the lockfile now includes two versions of sphinx and selects one or the other depending on the Python version.

Double-checked by running

```
uv python pin 3.10
uv run ./x dist ferrocene-docs
```

and that completed with no issues

Xylakant:
Note that this does update python for all the things where we use python since setup-uv is used all over the place.

fs-rachel:
> Note that this does update python for all the things where we use python since setup-uv is used all over the place.

In that case, which option would be preferable?

1) Submit this as-is, see if the CI passes cleanly with Python 3.14

2) Only update the other dependencies, so that Python 3.14 is supported but not the default

fs-rachel:
or 3) not make this change, but instead add a section to the handbook explaining that we need Python 3.12 specifically and how to set that up

jyn514:
if this *works* i don't see any downside to upgrading our deps across the repo. it's good to do that once in a while, we don't want to be stuck on old unmaintained versions, not least because we use these for downloads and we might need to do a security update.

> Submit this as-is, see if the CI passes cleanly with Python 3.14

this is my preference.

jyn514 (approved):
bors merge

i tested this by reverting uv.lock to the old state, then running `cd ferrocene/doc && uv lock --upgrade` and diffing it against Rachel's version. i had to first upgrade uv from 0.8 to 0.11 so it would respect `exclude-newer`. i didn't see any differences in the generated versions, only in the format of the lockfile itself (the `marker = "python_full_version"` stuff went away, which i assume is because i had a different version of uv installed than Rachel.

fs-rachel:
> i had to first upgrade uv from 0.8 to 0.11 so it would respect `exclude-newer`.

Good to know! I was debating whether to update the version of uv in the CI script, but didn't on the basis that it's relatively new compared to the other dependencies. Shouldn't matter for CI, unless it complains about the newer lockfile format, because CI won't be updating the lockfile itself.

> i didn't see any differences in the generated versions, only in the format of the lockfile itself (the `marker = "python_full_version"` stuff went away, which i assume is because i had a different version of uv installed than Rachel.

Possibly? I have version 0.11.21. It looks like those are there to guide the package resolver toward consistent results, because without them it could pick inconsistent package versions run-to-run: https://docs.astral.sh/uv/reference/internals/resolver/#forking

jyn514:
i have `uv 0.11.25 (Homebrew 2026-06-26 aarch64-apple-darwin)`. i think it's probably not worth investigating. good to know what the option does though :)

PR #2422, opened 2026-06-26
update QNX scripts to work with QEMU built from source

japaric:
`qemu-system-$ARCH -netdev bridge,(..)` uses the qemu-bridge-helper script that was installed with `qemu-system-$ARCH`. However, to set up the bridge network in the QNX scripts, we are using the qemu-bridge-helper of the system installed QEMU.

When using a QEMU built from source, we have seen the VM fail to initialize due to a different version of qemu-bridge-helper being used to set up the network bridge. To ensure the right qemu-bridge-helper is using in the QNX script, we search for it using the absolute path of the `qemu-system-$ARCH` as the reference, instead of assuming it'll be in `/usr` as that only works when the system installed QEMU is used.

### DIFF
--- a/ferrocene/ci/scripts/qnx-common.sh
+++ b/ferrocene/ci/scripts/qnx-common.sh
@@ -47,18 +47,24 @@ locate_qemu_bridge_helper() {
 }
 
 locate_bridge_conf() {
-    helperdir="$(dirname "$1")"
-    # default search location when built from source
-    conf1="$helperdir/../etc/qemu/bridge.conf"
-    # location in Debian/Ubuntu package
-    conf2="$helperdir/../../../etc/qemu/bridge.conf"
-    if [ -f "$conf1" ]; then
-        echo "$(readlink -f "$conf1")"
-    elif [ -f "$conf2" ]; then
-        echo "$(readlink -f "$conf2")"
-    else
-        panic "bridge.conf not found; update this search logic (helper=$1)"
-    fi
+    qemu_bridge_helper="$1"
+
+    # The location where `qemu-bridge-helper` expects `bride.conf` to be is
+    # different on a custom QEMU build vs on the Debian packaged QEMU. Also, the
+    # file is not contained in the Debian package not created by
+    # `make install` on a custom build; instead the QNX SDP's `net.sh` script
+    # will create it. As `bridge.conf` does not exist we cannot search for it
+    # using `find`. The most reliable way to get its expected location seems to
+    # be to extract the location directly from the helper binary itself
+    #
+    # Examples of bridge-qemu-helper and bridge.conf locations
+    # - Ubuntu/Debian package:
+    #   - `/usr/lib/qemu/qemu-bridge-helper`;
+    #   - `/etc/qemu/bridge.conf`
+    # - custom build:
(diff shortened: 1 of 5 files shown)


jyn514 (approved):
seems reasonable.

bors r+

jyn514:
seems like a real error.
```
^@^@QNX_SHARED_LICENSE_FILE=/home/ci/project/qnx710-472/license/licenses
QNX_HOST=/home/ci/project/qnx710-472/host/linux/x86_64
QNX_TARGET=/home/ci/project/qnx710-472/target/qnx7
MAKEFLAGS=-I/home/ci/project/qnx710-472/target/qnx7/usr/include

===> setting up QEMU bridge network
error: bridge.conf not found; update this search logic (helper=/usr/lib/qemu/qemu-bridge-helper)
```

japaric:
> seems like a real error.

yeah, bridge.conf does not exist in the Docker image as it is created in the container by the QNX SDP `net.sh` script so we cannot check for its presence to confirm that it is in one of the expected locations. I have changed the logic to extract the expected location from the bridge helper binary instead

ferrocene/ci/scripts/qnx-common.sh
```diff
@@ -47,18 +47,24 @@ locate_qemu_bridge_helper() {
...
+    # be to extract the location directly from the helper binary itself
+    #
+    # Examples of bridge-qemu-helper and bridge.conf locations
+    # - Ubuntu/Debian package:
+    #   - `/usr/lib/qemu/qemu-bridge-helper`;
+    #   - `/etc/qemu/bridge.conf`
+    # - custom build:
+    #   - `$PREFIX/libexec/qemu-bridge-helper`;
+    #   - `$PREFIX/etc/qemu/bridge.conf`
+    strings "$qemu_bridge_helper" | grep '^/.*/bridge.conf$'
```
jyn514, line 67: well this is a little terrifying 😆 is there any way we can run `net.sh` ahead of time so we can see where it actually puts the bridge?

japaric: `net.sh` *takes* the full path to `bridge.conf` as an argument and it will create/extend it there, where you tell it to put it even if it's wrong. you have to pass the full `bridge.conf` path that QEMU expects to `net.sh` or `qemu-bridge-helper` will fail later on. so, we cannot get the expected `bridge.conf` location from `net.sh`, only from QEMU itself, or rather, from `qemu-bridge-helper`.

the other option I can think of is running a dummy QEMU invocation first that tries to use the bridge network and helper, which will fail because `bridge.conf` does not exist, and then parse the error message. the expected `bridge.conf` location will be contained there

the expected `bridge.conf` appears to be in the `--sysconfdir` location you pass to QEMU's `configure` script but `qemu-system-* --version` does not print this (or any) configuration setting , like say `gcc -v` does.

`qemu-bridge-helper` itself has a very primitive command line parsing. it does not accept `--help` or `--version` flags; it only prints an "usage" one liner if the arguments are wrong.

jyn514: this is so sad. i guess this is ok; you're the one using this script most often, if you're willing to maintain it i'm ok with the change.

jyn514 (approved):
bors merge

PR #2425, opened 2026-06-29
replace the ubuntu 20 executors with a parametrized executor, delete unused

Xylakant:
we can pass parameter to executors and since all of the ubuntu executors only differ in the image used, we can replace them all with a single executor

note: the ubuntu-24 docker executors were stale and unused


Hoverbear (approved):

jyn514:
bors merge

PR #2428, opened 2026-06-30
Fix symbol report in CI

fs-rachel:
The pull-upstream script attempts to update the symbol report, but this was always failing with a message:

    assertion `left == right` failed: x.py should be run with `--stage 2` on CI, but was run with `--stage 1`

This is a bit misleading: we really do want to use stage 1, and that *is* permitted as long as it's requested explicitly. It's only implicit use of stage 1 which is disallowed, because that's usually wrong.

Improve the error message to better reflect the actual error condition, and add an explicit `--stage 1` to our CI script to bypass the error.


jyn514 (approved):
Rachel is planning to send that first commit upstream.

bors merge

fs-rachel:
Error message fix sent upstream as https://github.com/rust-lang/rust/pull/158614

jyn514:
bors retry

(yielding to https://github.com/ferrocene/ferrocene/pull/2429)

jyn514:
ugh that pr wasn't actually approved

PR #2434, opened 2026-07-01
feat: tag built ci docker images with stable aliases on main

dotcarmen:
we want to build qemu using the ferrocene ci images, but right now there's no stable identifier that can be used to pull these images from our pull-through cache on harbor

this PR adds a new `misc` workflow for ci actions that aren't part of the main testing workflow.

when a commit hits `main`, pushes an aliased tag (currently the same, without the hash suffix) to the ECR registry for the docker image that was built from the current tag.

image

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1029,11 +1029,10 @@ jobs:
           name: Empty step to make sure the job always has steps
           command: echo
 
-  alias-docker-x86_64:
-    description: Specialized version of build-docker-x86_64 for running on main and tagging the image as such
+  alias-docker:
+    description: Specialized version of build-docker-* for running on main and tagging the image as such
     docker:
       - image: cimg/base:current
-    resource_class: large # 4-core
     parameters:
       image:
         description: The image that needs to be built
@@ -1044,6 +1043,10 @@ jobs:
       alias:
         description: The tag of the alias to push
         type: string
+      resource_class:
+        description: Resource class of the runner
+        type: string
+    resource_class: << parameters.resource_class >>
     steps:
       - aws-cli/install:
           version: << pipeline.parameters.awscli-version >>
@@ -1096,37 +1099,6 @@ jobs:
       - run:
           name: Empty step to make sure the job always has steps
           command: echo
-
-  alias-docker-aarch64:
-    description: Specialized version of build-docker-aarch64 for running on main and tagging the image as such
-    docker:
-      - image: cimg/base:current
-    resource_class: arm.medium # 2-core
-    parameters:
-      image:
-        description: The image that needs to be built
-        type: string
-      tag:
-        description: The tag of the image that needs to be built
-        type: string
-      alias:
-        description: The tag of the alias to push
-        type: string
-    steps:
-      - aws-cli/install:
-          version: << pipeline.parameters.awscli-version >>
-          override_installed: true
-      - aws-oidc-auth
-      - ferrocene-git-shallow-clone:
-          depth: 1
-      - setup_remote_docker:
-          version: default
-      - run:
-          name: Maybe build x86_64-<< parameters.image >> and upload
-          environment:
-            ECR_REPOSITORY: ci-docker-images
-            ECR_REGION: us-east-1
-          command: ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh '<< parameters.image >>' '<< parameters.tag >>' '<< parameters.alias >>'
 
   # The image gets built for every job because it needs to contain build results per job
   build-docker-rhivos2-test-server-aarch64:
@@ -1685,20 +1657,23 @@ workflows:
   tag-images:
     jobs:
       # tag built docker images with the `:main` tag
-      - alias-docker-x86_64:
+      - alias-docker:
           filters: pipeline.git.branch == "main"
+          resource_class: large # 4-core
           name: alias-docker-x86_64-ubuntu-20
           image: ubuntu-20
           tag: << pipeline.parameters.docker-image-tag--x86_64--ubuntu-20 >>
           alias: x86_64-ubuntu-20
-      - alias-docker-x86_64:
+      - alias-docker:
           filters: pipeline.git.branch == "main"
+          resource_class: large # 4-core
           name: alias-docker-x86_64-ubuntu-24
           image: ubuntu-24
           tag: << pipeline.parameters.docker-image-tag--x86_64--ubuntu-24 >>
           alias: x86_64-ubuntu-24
-      - alias-docker-aarch64:
+      - alias-docker:
           filters: pipeline.git.branch == "main"
+          resource_class: arm.medium # 2-core
           name: alias-docker-aarch64-ubuntu-20
           image: ubuntu-20
           tag: << pipeline.parameters.docker-image-tag--aarch64--ubuntu-20 >>
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1058,30 +1058,7 @@ jobs:
           environment:
             ECR_REPOSITORY: ci-docker-images
             ECR_REGION: us-east-1
-          command: |
-            account="$(aws sts get-caller-identity --region "${ECR_REGION}" | jq -r .Account)"
-            registry="${account}.dkr.ecr.${ECR_REGION}.amazonaws.com"
-
-            aws ecr get-login-password --region "${ECR_REGION}" \
-              | docker login --username AWS --password-stdin "${registry}"
-
-            manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag='<< parameters.tag >>' --output text --query 'images[].imageManifest')"
-            if [[ "$?" -ne '0' ]]; then
-              echo "image << parameters.tag >> doesn't exist, building..."
-              IMAGE_NAME="<< parameters.image >>" \
-                IMAGE_TAG="<< parameters.tag >>" \
-                ferrocene/ci/scripts/build-and-push-docker-image.sh
-              manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag='<< parameters.tag >>' --output text --query 'images[].imageManifest')"
-            else
-              alias_manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag='<< parameters.alias >>' --output text --query 'images[].imageManifest')"
-              src_digest="$(echo "$manifest" | jq -r .config.digest)"
-              alias_digest="$(echo "$alias_manifest" | jq -r .config.digest)"
-              if [[ "$src_digest" == "$alias_digest" ]]; then
-                echo "alias << parameters.alias >> already points to << parameters.tag >>"
-                exit 0
-              fi
-            fi
-            aws ecr put-image --repository-name "$ECR_REPOSITORY" --image-tag '<< parameters.alias >>' --image-manifest "$manifest"
+          command: ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh '<< parameters.image >>' '<< parameters.tag >>' '<< parameters.alias >>'
 
   build-docker-aarch64:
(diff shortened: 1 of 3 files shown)


.circleci/workflows.yml
```diff
@@ -1761,6 +1819,29 @@ workflows:
     jobs:
       - skip
 
+  # workflow for miscellaneous tasks that don't affect the status of a commit
+  misc:
+    jobs:
```
jyn514, line 1660: can you rename this to "tag-images" or something like that? "misc" could be literally anything.

.circleci/workflows.yml
```diff
@@ -1029,6 +985,60 @@ jobs:
...
+            else
+              alias_manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag='<< parameters.alias >>' --output text --query 'images[].imageManifest')"
+              src_digest="$(echo "$manifest" | jq -r .config.digest)"
+              alias_digest="$(echo "$alias_manifest" | jq -r .config.digest)"
+              if [[ "$src_digest" == "$alias_digest" ]]; then
+                echo "alias << parameters.alias >> already points to << parameters.tag >>"
+                exit 0
+              fi
+            fi
+            aws ecr put-image --repository-name "$ECR_REPOSITORY" --image-tag '<< parameters.alias >>' --image-manifest "$manifest"
```
jyn514: this is impossible for me to test and quite hard for me to review. could you at least make it a separate script please? it could go in `ferrocene/ci/scripts`. once you've done that, please run shellcheck on it.

jyn514 (changes requested):

jyn514:
> when a commit hits main, pushes an aliased tag (currently the same, without the hash suffix) to the ECR registry for the docker image that was built from the current tag.

why are we making this change? why is it useful?

dotcarmen:
> we want to build qemu using the ferrocene ci images, but right now there's no stable identifier that can be used to pull these images from our pull-through cache on harbor

updated PR description:

> we want to build qemu using the ferrocene ci images, but right now there's no stable identifier that can be used to pull these images from our pull-through cache on harbor

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,39 @@
...
+
+manifest=""
+if manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag="$SRC" --output text --query 'images[].imageManifest')"; then
+    echo "image $SRC doesn't exist, building..."
+    IMAGE_NAME="$IMAGE" \
+    IMAGE_TAG="$SRC" \
+    ferrocene/ci/scripts/build-and-push-docker-image.sh
+    manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag="$SRC" --output text --query 'images[].imageManifest')"
+else
+    alias_manifest="$(aws ecr batch-get-image --repository-name "$ECR_REPOSITORY" --image-ids imageTag="$ALIAS" --output text --query 'images[].imageManifest')"
```
jyn514: nit: for my sanity, could you make this a `get_manifest_digest` function rather than duplicating it across lines 23 and 30? nbd, just easier for me to read

jyn514 (commented):
seems broadly fine. i think you're still testing this, let me know when you want a re-review.

dotcarmen:
here's the working pipeline https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18583

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,47 @@
...
+function get_manifest_digest() {
+    jq -r .config.digest
+}
+
+aws ecr get-login-password --region "${ECR_REGION}" \
+    | docker login --username AWS --password-stdin "${registry}"
+
+manifest=""
+if manifest="$(get_manifest "$SRC")"; then
+    echo "image $SRC doesn't exist, building..."
```
jyn514: i'm not sure i follow this logic ... if you *can* get the image manifest, rebuild it? shouldn't that be reversed?

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,47 @@
...
+    IMAGE_NAME="$IMAGE" \
+    IMAGE_TAG="$SRC" \
+    ferrocene/ci/scripts/build-and-push-docker-image.sh
+    manifest="$(get_manifest "$SRC")"
+else
+    alias_manifest="$(get_manifest "$ALIAS")"
+    src_digest="$(echo "$manifest" | get_manifest_digest)"
```
jyn514: won't `manifest` be unset here, since `get_manifest` failed?

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
jyn514:
```suggestion
#!/usr/bin/env bash
```

jyn514 (commented):
@dotcarmen i see that new job rebuilding the image from scratch, is that intentional? there's no way to just say "tag this existing image with a new tag"? this might be related to the new review comment i've just left, unsure.

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
dotcarmen: yes it should be... sorry

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
dotcarmen: like you said above, the condition in the `if` should be negated - so `manifest` _should_ be set here

dotcarmen: fixed https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18687

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
dotcarmen: fixed https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18687

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function get_manifest_digest() {
+    jq -r .config.digest
+}
+
+aws ecr get-login-password --region "${ECR_REGION}" \
+    | docker login --username AWS --password-stdin "${registry}"
+
+manifest=""
```
jyn514: nit: this is unnecessary, it'll get set on the line below either way.

jyn514 (commented):
code looks good :) do you have an example of a working pipeline after the logic fix?

dotcarmen:
> code looks good :) do you have an example of a working pipeline after the logic fix?

https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18687

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
dotcarmen: this is to ensure the variable is set at the right scope... i'll test without this line, but bash scopes are weird, which is why i included it :)

jyn514: bash doesn't have scopes, period, unless you use `local`. you don't use `local` here so everything is in global dynamic scope.

jyn514 (approved):
> alias aarch64-ubuntu-20 already points to aarch64-ubuntu-20-961cc067a91caa0b7c801750319c535814ae185aac65d4b28fd61cce92933322

beautiful

bors merge

dotcarmen:
ok it _actually_ works this time, i verified!
- [when the alias already points to the src](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18714/workflows/54de1072-3745-473d-bcb7-2092339d6e13/jobs/174087)
- [when the image must be rebuilt](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18712/workflows/a54f829b-7ef0-423e-8697-d766f19b2513/jobs/174081)
- [when the image already exists, but the alias isn't pointing to it](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18717/workflows/afef7a87-1016-451b-9d29-a4c04223ddc3/jobs/174090)

note these are technically with different commits, but i fixed the bug from (https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18712/workflows/a54f829b-7ef0-423e-8697-d766f19b2513/jobs/174082), cleaned up comments, and otherwise the 'DO NOT MERGE' commit had a file i used to force-change the hash so i could test scenarios 2 and 3

.circleci/workflows.yml
```diff
@@ -1029,6 +1029,37 @@ jobs:
           name: Empty step to make sure the job always has steps
           command: echo
 
+  alias-docker-x86_64:
```
Xylakant: This can be unified with aarch64 version - it does essentially the same and can be parameterized.

ferrocene/ci/scripts/alias-docker-image-tag.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
Xylakant, line 1: The result of calculate-parameters already tells you whether the image exits (respectively needs to be rebuilt). If this runs as a workflow that depends on the setup step, we should rely on the result of that step

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
```diff
@@ -0,0 +1,54 @@
...
+src_digest="$(get_manifest_digest "$src_manifest")"
+if [[ -n "$src_digest" ]]; then
+    echo "image $SRC exists"
+    alias_manifest="$(get_manifest "$ALIAS")"
+    alias_digest="$(get_manifest_digest "$alias_manifest")"
+    if [[ "$src_digest" == "$alias_digest" ]]; then
+        echo "alias $ALIAS already points to $SRC"
+        exit 0
+    fi
+else
```
Xylakant: I think that if the image does not exist, we should just have a separate step, just reusing the one that normally builds the image.

Xylakant (changes requested):
Sorry to be late to the party. I also believe we should revisit the name of the tags. The current tag names are based on the versions of the underlying operating system and not on the function these images have. That's something we should look at and discuss internally.

.circleci/workflows.yml
dotcarmen: ah, i previously tried to parameterize the `resource_class`, but i think i was doing it wrong. fixed

ferrocene/ci/scripts/alias-docker-image-tag.sh
dotcarmen: fixed by depending on the existing build steps

ferrocene/ci/scripts/maybe-build-push-and-alias-docker-images.sh
dotcarmen: fixed by depending on the existing build steps

dotcarmen:
addressed felix's feedback

tests:
- [with rebuild](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18759/workflows/45c76d43-ca84-406e-83ff-3edb2f1e1c27) (ubuntu-24--x86_64 is the only one affected)
  - aliases already match for the other images
- after ^, [this run](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/18762/workflows/bfd9fa25-e382-4a61-b913-fae927b306c1/jobs/174418) restored the image tag

jyn514:
i am planning to let Felix drive the rest of the review; ping me if you want me to take another look, but otherwise i trust the two of you to do a good job here :)

PR #2435, opened 2026-07-01
Inline ignore-tests.toml into individual tests

fs-rachel:


src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -2172,6 +2172,18 @@ NOTE: if you're sure you want to do this, please open an issue as to why. In the
...
+        const DEBUGINFO_DISABLED_TARGETS: [&'static str; 5] = [
+            "aarch64r82-unknown-ferrocene.facade",
+            "aarch64v8r-unknown-ferrocene.facade",
+            "aarch64-unknown-ferrocene.facade",
+            "thumbv7em-ferrocene.facade-eabi",
+            "thumbv7em-ferrocene.facade-eabihf",
+        ];
+        if self.suite == "debuginfo" && DEBUGINFO_DISABLED_TARGETS.contains(&&*self.target.triple) {
+            return;
+        }
```
jyn514: i think we might have a helper for "is a facade target", don't remember off the top of my head.

src/tools/compiletest/src/directives/directive_names.rs
```diff
@@ -147,6 +159,9 @@ pub(crate) const KNOWN_DIRECTIVE_NAMES: &[&str] = &[
     "ignore-x86",
     "ignore-x86_64",
     "ignore-x86_64-apple-darwin",
+    // Ferrocene additions
+    "ignore-x86_64-pc-nto-qnx710",
+    "ignore-x86_64-pc-nto-qnx800",
```
jyn514: let's visually distinguish these so we can tell which are additions and which aren't.
```suggestion

    // Ferrocene additions
    "ignore-x86_64-pc-nto-qnx710",
    "ignore-x86_64-pc-nto-qnx800",

```
(and similarly for the other new comments)

tests/run-make/c-link-to-rust-va-list-fn/rmake.rs
```diff
@@ -7,6 +7,24 @@
...
+//@ ignore-aarch64-unknown-nto-qnx710
+//@ ignore-x86_64-pc-nto-qnx710
+//@ ignore-aarch64-unknown-nto-qnx800
+//@ ignore-x86_64-pc-nto-qnx800
+
+// Ferrocene addition: libcheckrust includes symbols which cannot be used without `-fPIC` on
+// these platforms
+//@ ignore-aarch64r82-unknown-ferrocene.facade
+//@ ignore-aarch64v8r-unknown-ferrocene.facade
+//@ ignore-aarch64-unknown-ferrocene.facade
```
jyn514: can we add new `ignore-facade` and `ignore-qnx` directives so we don't have to duplicate all these so much?

tests/run-make/pointer-auth-link-with-c/rmake.rs
```diff
@@ -11,6 +11,11 @@
 //@ ignore-cross-compile
 // Reason: the compiled binary is executed
 
+// Ferrocene addition: This tests the interoperabilty an unstable feature of the rust compiler
+// (enabling PAC) with newer C compilers (‘-mbranch-protection=’).
```
jyn514: are these targets using too old a version of gcc or something like that?

tests/ui/alloc-error/alloc-error-backtrace.rs
```diff
@@ -23,6 +23,14 @@
 //@ ignore-backends: gcc
 //@ ignore-msvc see #62897 and `backtrace-debuginfo.rs` test
 
+// Ferrocene addition: These targets cannot unwind
+//@ ignore-aarch64-unknown-ferrocene.facade
+//@ ignore-aarch64r82-unknown-ferrocene.facade
+//@ ignore-aarch64v8r-unknown-ferrocene.facade
+//@ ignore-armv7r-ferrocene.facade-eabihf
+//@ ignore-thumbv7em-ferrocene.facade-eabi
+//@ ignore-thumbv7em-ferrocene.facade-eabihf
```
jyn514: this looks super broken (not your fault, pre-existing). backtrace printing is *not* the same as unwinding and should work even with panic=abort. if backtraces are broken on facade targets it's not a *giant* deal, since those are only used for tests, but it's still a little concerning. i think instead we should probably mark all the facade targets such that `needs-subprocess` skips them? you can't spawn subprocesses on embedded targets anyway.

tests/ui/hygiene/panic-location.rs
```diff
@@ -1,3 +1,6 @@
+// Ferrocene addition: QEMU user space emulation outputs an extra message when an abort happens
```
jyn514, line 1: we have a whole bunch of these. can we normalize away the extra QEMU message rather than ignoring all these tests? see `fn normalize_output` in src/tools/compiletest/src/runtest.rs.

tests/ui/sanitizer/leak.rs
```diff
@@ -6,6 +6,10 @@
 //@ run-fail
 //@ error-pattern: LeakSanitizer: detected memory leaks
 
+// Ferrocene addition: Sanitizers are not a qualified compiler feature. Support on RHIVOS2 needs
+// investigation.
+//@ ignore-aarch64-rhivos2-linux-gnu
```
jyn514, line 11: i think you've added this for everything in the `sanitizer` directory, right? can you configure this in one place in bootstrap for the rhivos target, rather than duplicating it into every test?

src/bootstrap/src/core/build_steps/test.rs
fs-rachel: From a quick search, there doesn't seem to be a widely-used helper, there's just lots of `target.contains("facade")` and `target.contains("ferrocene.facade")` in various places.

I can do that, but just worth noting that that would disable these tests on more targets than the current code does

tests/ui/hygiene/panic-location.rs
fs-rachel: I was thinking of that as a followup, but can do it in this patch set if you'd prefer

tests/ui/sanitizer/leak.rs
fs-rachel: Interestingly there are a bunch of other tests in `ui/sanitizer/` which weren't excluded for this platform. Not sure if they were already excluded due to other ignore rules or what

tests/run-make/c-link-to-rust-va-list-fn/rmake.rs
fs-rachel: I dug into how the `ignore-*` rules work, `prepare_conditions()` seems to be the important bit. And it looks like the rules we want already exist, as `ignore-ferrocene.facade` and `ignore-nto`. So I can use those.

fs-rachel: Only potential issue with doing that is that we'd lose specific attribution of "this fails on platform X for reason Y"

tests/run-make/pointer-auth-link-with-c/rmake.rs
fs-rachel: I don't know, let's try re-enabling it and see how the CI goes

tests/run-make/c-link-to-rust-va-list-fn/rmake.rs
fs-rachel: Note to self: if we do replace the single-target ignores for the QNX builds with `//@ ignore-nto`, I should double it up with `//@ ignore-qnx` as well to match https://github.com/ferrocene/ferrocene/pull/2418/changes/bbf0f29c7c51f4c0740d715b3a4525190ff87ffb

src/bootstrap/src/core/build_steps/test.rs
fs-rachel: Done

src/tools/compiletest/src/directives/directive_names.rs
fs-rachel: Done, the format check didn't like the blank lines so I've switched to adding `// Ferrocene addition` on the same line as each new directive

tests/ui/alloc-error/alloc-error-backtrace.rs
fs-rachel: Done

src/tools/compiletest/src/directives/directive_names.rs
jyn514: that's fine, `#[rustfmt::skip]` would also have been fine.

tests/run-make/c-link-to-rust-va-list-fn/rmake.rs
jyn514: great, let's use the existing `ignore-ferrocene.facade` and `ignore-nto` directives. i'll leave a note on Jorge's PR so he knows he'll need to update these when he rebases.

tests/ui/hygiene/panic-location.rs
jyn514: 👍 follow-up seems good

tests/ui/sanitizer/leak.rs
jyn514: absolutely wild. ok, let's worry about that in the follow-up too.

jyn514 (approved):
this all seems reasonable. happy to send it to bors; also happy to let you keep iterating. i trust your judgement here :)

also recall that you have the ability to trigger CircleCI jobs without going through bors, which avoids running macOS / Windows runners. i think that's probably a good next step here.

tests/run-make/c-link-to-rust-va-list-fn/rmake.rs
fs-rachel: Done

tests/run-make/pointer-auth-link-with-c/rmake.rs
fs-rachel: Answer: yes, the compiler on our CI system is outdated. It complains about not understanding `-mbranch-protection=gcs`, which was added to Clang (at least) in 2023: https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20231211/1261499.html

fs-rachel:
Updates:
1) Rebase to pick up disabling tests on qnx8
2) Fix a line number change I missed
3) Re-disable `tests/run-make/pointer-auth-link-with-c` and `tests/ui/threads-sendsync/mpsc_stress`. I speculatively re-enabled these to see what would happen, but they still break on our CI setup due to outdated gcc and qemu respectively

jyn514:
bors merge

fs-rachel:
I'm still testing to make sure nothing else breaks, probably best not to merge yet

bors cancel

fs-rachel:
@jyn514 Everything passes now, so this is ready for review/merge

jyn514:
bors merge

PR #2436, opened 2026-07-01
Test: Remove src/gcc from tree

fs-rachel:
Checking out gcc takes far longer than any other submodule. See if we can get away with not including it in CI runs


fs-rachel:
Closing in favour of https://github.com/ferrocene/ferrocene/pull/2448

PR #2438, opened 2026-07-02
feat: use squid-cache for a bunch of ci downloads

dotcarmen:
blocked on https://github.com/ferrocene/rancher-infrastructure/pull/32 which itself is still blocked on a couple of PRs. once those land, this should work

this PR opts-in a bunch of commands to use the CI squid cache. 🤞🏻 there are fewer network-related failures and retries

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -135,12 +135,12 @@ jobs:
         ./x.py test $(ferrocene/ci/split-tasks.py test:misc-checks)
     steps:
       - aws-oidc-auth
+      - use-cache-proxy
       - ferrocene-checkout:
           llvm-subset: true
       - run:
           name: Perform licensing checks
           command: |
-            source ferrocene/ci/scripts/use-squid-cache.sh
             uvx reuse@5.1.1 --include-submodules lint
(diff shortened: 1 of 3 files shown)


dotcarmen:
i just went through and found a bunch of commands that look like they _should_ be cached. i'll do a more thorough investigation next week

jyn514:
i note https://github.com/ferrocene/rancher-infrastructure/pull/32 is now merged. not sure the state of this.

.circleci/workflows.yml
```diff
@@ -140,6 +140,7 @@ jobs:
       - run:
           name: Perform licensing checks
           command: |
+            source ferrocene/ci/scripts/use-squid-cache.sh
```
Xylakant: I think the better way to solve this is to just set the environment variables instead of using a script.

Xylakant (changes requested):
I would generally set the env variables instead of using a script.

.circleci/workflows.yml
dotcarmen: set the environment variables globally? setting via a script makes sure there's no typos :)

PR #2441, opened 2026-07-02
Document the forward porting process

Hoverbear:
We have been cultivating a new practice of better tracking our forward ports. This attempts to document that.


jyn514 (approved):
thank u i appreciate u

do you think it makes sense to add automation for this the same way we have automation for backports?

bors merge

PR #2442, opened 2026-07-03
minor improvements to tests for unvalidated lint

tshepang:


tests/ui/ferrocene/lint-prevalidated/main.rs
```diff
@@ -1,4 +1,3 @@
-#![crate_type = "bin"]
```
jyn514, line 1: this is a no-op because rustc defaults to `--crate-type=bin`, but IMO it makes it hard to understand what this is actually testing. the point is that `fn main` is considered an entrypoint for binary crates but not libraries.

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
```diff
@@ -6,7 +6,6 @@
 // https://github.com/rust-lang/rfcs/pull/3808#discussion_r2616622459,
 // and https://github.com/rust-lang/rust/issues/54726#issuecomment-431931811 for a possible
 // way to relax the limitation.
-// #![ferrocene::prevalidated]
```
jyn514: this was intentional. if you remove the commented attribute, please also update the comment above.
```suggestion
// FIXME: `#![ferrocene::prevalidated]` doesn't work at crate level, "custom inner attributes are unstable"
// and it has to be that way, see
// https://github.com/rust-lang/rfcs/pull/3808#discussion_r2616622459,
// and https://github.com/rust-lang/rust/issues/54726#issuecomment-431931811 for a possible
// way to relax the limitation.
```

tests/ui/ferrocene/lint-prevalidated/trait-ops.rs
jyn514, line 1: recording that i reviewed this file and all the trait ops are still present, just reordered. for next time, it would be helpful if you could split the reorder and additional tests, this was quite hard to review.

jyn514 (approved):
two nits, otherwise lgtm.

tests/ui/ferrocene/lint-prevalidated/main.rs
tshepang: maybe it surprised me because it's normally not mentioned in compiletests, and I see it's not there in the other bin tests in its parent directory... should we keep this specific one with a comment

jyn514: i think it's ok to remove the `#![]` attribute if we also add a comment like "test that `fn main` is considered validated in binary crates", does that work?

tshepang: 183d2bdd785b84da108734d1116f225060ca719c

tests/ui/ferrocene/lint-prevalidated/trait-ops.rs
tshepang: https://github.com/ferrocene/ferrocene/pull/2442/commits/f6027f5c6d50f59b7c6bbef44047ffb688e4b9b6

jyn514 (approved):
bors merge

PR #2444, opened 2026-07-03
Combine many of our testjob pairs to a single job

Xylakant:
We had a pair of testjobs for every target which made the individual jobs take less wall-clock time, at the expense of transferring more data and more CPU time.

Other downsides are that a higher number of jobs leads to increased risks of individual job failure and higher peak load once we move to a fully self-hosted CI.

This is a slice of the changes of #2413, split out for easier review.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -878,7 +878,7 @@ jobs:
         type: string
       target:
         type: string
-      job:
+      tasks:
         type: string
       resource-class:
         type: string
@@ -887,7 +887,7 @@ jobs:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
       FERROCENE_TARGETS: << parameters.target >>
       SCRIPT: |
-        sudo update-binfmts --import && sudo update-binfmts --enable << parameters.qemu-arch >> && ./x.py --stage 2 test $(ferrocene/ci/split-tasks.py << parameters.job >>) --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
+        sudo update-binfmts --import && sudo update-binfmts --enable << parameters.qemu-arch >> && ./x.py --stage 2 test << parameters.tasks >> --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
     steps:
       - aws-oidc-auth
       - ferrocene-checkout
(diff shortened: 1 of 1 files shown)


.circleci/workflows.yml
```diff
@@ -1334,80 +1329,42 @@ workflows:
...
-      - facade-generic-test:
-          name: aarch64v8r-none-test-library
+          name: aarch64v8r-none-test-library-compiletest
           target: aarch64v8r-unknown-ferrocene.facade
           test-variant: 2021-neoverse-v1
           qemu-arch: qemu-aarch64
-          tasks: $(ferrocene/ci/split-tasks.py test:library-nostd)
-          resource-class: large # 4-core
+          tasks: $(ferrocene/ci/split-tasks.py qnx:compiletest-no-only-hosts) $(ferrocene/ci/split-tasks.py test:library-nostd)
+          resource-class: ferrocene/vm-ubuntu-20-amd64-medium # 4-core
```
jyn514: am i reading correctly that previously this was split over a self-hosted and CircleCI-hosted runner? and that after this change, both of them are on the same self-hosted runner?

jyn514 (approved):
LGTM. i have one question — i think the code is fine regardless of the answer, i'm just not sure whether the behavior here matches the "split out changes which don’t touch what runners we’re testing on" rationale we discussed in chat.

jyn514:
CI also seems to be failing on the "add AWS CLI" step, maybe you need to add a few more of the commits from the other PR?

.circleci/workflows.yml
Xylakant: Indeed this should not have been pulled in by this patchset. That's also the cause of the failure downstream.

Xylakant:
> CI also seems to be failing on the "add AWS CLI" step, maybe you need to add a few more of the commits from the other PR?

Actually rather: There were changes in the PR that weren't supposed to be here and thus cause the fail. Sorry about that.

Hoverbear (approved):

Xylakant:
Screenshot 2026-07-03 at 21 13 40

Comparison of the test results of the latest merge to main to the last testrun. This PR is not exactly on the same base as the merge, which explains the slight differences.

Xylakant:
bors merge

PR #2439, opened 2026-07-02
QNX target renaming

japaric:
**NOTE** depends and includes PR #2454 

This is a rebased version of #2418 now that https://github.com/rust-lang/rust/pull/158449 has been merged. the first six commits in this PR are a cherry picks the commits in https://github.com/rust-lang/rust/pull/158449. the rest of commits are a cherry pick of #2418. note that commit [fix rustdoc-html tests](https://github.com/ferrocene/ferrocene/pull/2418/commits/14836809d26d30002502fd9e0533ebab633c50d7) has been dropped because it became unnecessary with upstream commit https://github.com/rust-lang/rust/pull/158449/changes/4eec7f5d40dacb0813a95fa1c761c077f11a9911

---

- `aarch64-unknown-nto-qnx800` becomes simply `aarch64-unknown-qnx`
- `x86_64-pc-nto-qnx800` becomes simply `x86_64-pc-qnx`
- references to QNX OS and QNX Neutrino RTOS are replaced with QNX SDP (Software Development Platform) for uniformity
- various `nto_qnx` modules are renamed to `qnx_sdp` to match the above

The new target names are more consistent with those used by the QNX SDP 8.0 toolchain, and reflect a level on ongoing API stability similar to that for macOS and Linux (which simply have minimum supported versions record for their targets). The new target names also match what the copy of `clang` shipped in the QNX Desktop Image (part of QNX SDP 8.0) self reports as (that is, `clang` running self-hosted *on* QNX).

PR #2430, opened 2026-06-30
Feature/orbify things

Xylakant:
This PR moves shared parts of the CI workflow definitions into a circleci orb at https://github.com/ferrocene/circleci-orb-ferrocene-common

While this means that we're suffering from the added friction of having a two-stage development process for changes to these parts, it significantly improves the development process of the shared parts because the orb development process is better organized. Orb are split into multiple files that allow for better separation and visibility, has tooling to lint both shell comands and yaml files and is in general easier to work with. It also allows reusing parts across multiple configuration files.

Most of the extracted parts have essentially been stable for an extended period of time, so I would expect that we don't see much development on the extracted parts.

This PR contains changes to the remote-testserver startup which are not immediately relevant to this change, but will be in future PRs.

PR #2446, opened 2026-07-05
use docker executors for embedded targets

Xylakant:
This uses the docker executor for facade tests, employing the same strategy that we use for the rhivos tests.

This PR contains the changes from https://github.com/ferrocene/ferrocene/pull/2484 rewritten to match the new facade test setup.

This change updates QEMU to a recent version as part of the setup work.

Switch the `x86_64-linux-generic-test-container` container to a platform-independent `linux-generic-test-container`. CircleCI does not support nested containers, so we cannot preserve the existing `x86_64-linux-generic-test-container`.

This also switches `linux-generic-test-container` from a `ferrocene/k8s-amd64-large` to `medium` runner, since our jobs hardly ever used the resources of the larger runner.

`should_panic` doctests for`panic=abort` platforms cannot be run through remote-testserver and have been disabled. This is a current limitation of remote-testserver. Doctests are run opportunistic and are  not qualification-relevant, so the slight reduction in tests is not harmful.


jyn514:
noting that https://github.com/ferrocene/ferrocene/pull/2439 will be merged through https://github.com/ferrocene/ferrocene/pull/2447, not as its own PR.

Xylakant:
> noting that #2439 will be merged through #2447, not as its own PR.

This was indeed a typo, this includes the changes from #2430, not 2439

Xylakant:
bors try

Xylakant:
bors try

Xylakant:
bors try

Xylakant:
bors retry

Xylakant:
bors try

Xylakant:
bors try

dotcarmen:
> This PR incorporates the changes of #2430 which will need to be merged first.

i'm gonna set the target branch to the branch from #2430 to make review easier. once #2430 lands, github will automatically change the target branch to `main`.

.circleci/workflows.yml
```diff
@@ -1216,55 +1170,87 @@ workflows:
...
+            - build-docker--ubuntu-24--x86_64
 
-      - facade-generic-test:
+      - ferrocene/facade_generic_test:
           name: thumbv7em-eabihf-test-library-compiletest
           target: thumbv7em-ferrocene.facade-eabihf
-          test-variant: 2021-cortex-m4
-          qemu-arch: qemu-arm
+          test_variant: 2021-cortex-m4
+          emulated: "true"
```
dotcarmen, line 1173: can you set `type: boolean` in the orb instead of relying on a string? also of note, [this typo](https://github.com/ferrocene/circleci-orb-ferrocene-common/blob/e9f2e7e79161139ef015f74bcf1aa31392f5f67d/src/executors/facade_testserver.yaml#L31) doesn't have an effect, since you pass `<< parameters.emulated >>` to the testserver script and ignore the env variable inside the script

.circleci/workflows.yml
```diff
@@ -158,13 +163,12 @@ jobs:
...
 
   # x86_64-unknown-linux-gnu jobs
-
   x86_64-linux-build:
     executor: *executor-x86_64-ubuntu-20
     resource_class: ferrocene/k8s-amd64-large
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
-      FERROCENE_TARGETS: aarch64-unknown-linux-gnu
+      FERROCENE_TARGETS: aarch64-unknown-none
```
dotcarmen: why?

ferrocene/ci/docker-images/common/setup-awscli.sh
dotcarmen, line 1: noting this change is from #2480

ferrocene/ci/docker-images/emulator/Dockerfile
dotcarmen, line 1: from #2480

ferrocene/ci/docker-images/emulator/run-testserver.sh
```diff
@@ -0,0 +1,39 @@
...
+
+TEST_DEVICE_ADDR=$2
+EMULATED=$3
+EMULATOR_VERSION=$4
+EMULATOR_ARCH=$5
+EMULATOR_CPU=$6
+
+env
+
+until [ -e /shared/testserver/remote-test-server  ]; do
```
dotcarmen, line 17: huh, til bash has `until` loops

library/alloc/src/vec/mod.rs
```diff
@@ -250,7 +250,7 @@ mod spec_extend;
 /// However be careful: if you try to access an index which isn't in the `Vec`,
 /// your software will panic! You cannot do this:
 ///
-/// ```should_panic
+/// ```should_panic,ignore-ferrocene.facade
```
dotcarmen, line 253: is there no way to support `should_panic` in the facade tests with the new qemu setup?

dotcarmen (commented):
mostly lgtm

.circleci/workflows.yml
Xylakant: indeed a mistake, thanks.

.circleci/workflows.yml
```diff
@@ -1062,8 +978,8 @@ workflows:
           awscli_version: << pipeline.parameters.awscli-version >>
 
       - misc-checks:
-          requires:
-            - build-docker--ubuntu-20--x86_64
```
dotcarmen: this was the correct indentation, can you undo this change?

.circleci/workflows.yml
Xylakant: the problem is that at some point this needs to change to be a string, because otherwise this ends up being weird in bash. I'd rather keep this as a string.

library/alloc/src/vec/mod.rs
Xylakant: no at the moment, but possibly in the future. Since doctests are opportunistic anyways, this is ok for now.

Xylakant:
bors try

.circleci/workflows.yml
dotcarmen: why not wrap the boolean like `"<< parameters.emulated >>"`? not gonna die on this hill though

but you should probably fix the typo of the env var in the orb before merging this

fs-rachel:
Looked through this, seems reasonable to me. Once this is merged up to the main branch, I'll update https://github.com/ferrocene/ferrocene/pull/2459 to work with the new way of doing things.

Note to self: Make the coverage jobs depend on the corresponding test job passing, like this patch does for the specific-cpu tests and for the same reasons

dotcarmen:
i'm having some issues downstream with getting the emulated platforms to work eg https://app.circleci.com/pipelines/gh/ferrocene/ferrocene/19409/workflows/5b08608d-9bcf-4423-8925-83986fd45667/jobs/188385

currently looking into this

fs-rachel:
@dotcarmen I saw a similar failure on a different patch as well: https://app.circleci.com/pipelines/gh/ferrocene/ferrocene/19398/workflows/17f584cd-a45a-4e8a-b6e2-bca9e4c56cb1/jobs/188173 for PR https://github.com/ferrocene/ferrocene/pull/2493

dotcarmen:
the problem is:

```
/bin/run-testserver.sh: line 39: /opt/qemu-ferrocene/11.0.1/bin/qemu-arm: No such file or directory
```

dotcarmen:
> > This PR incorporates the changes of #2430 which will need to be merged first.
> 
> i'm gonna set the target branch to the branch from #2430 to make review easier. once #2430 lands, github will automatically change the target branch to `main`.

this did not work 🙃 this branch has to be manually rebased against latest main

.circleci/workflows.yml
Xylakant: I believe I tried this and circleci was acting up and didn't transform that into the string `true`.

.circleci/workflows.yml
```diff
@@ -456,25 +467,6 @@ jobs:
...
-        type: string
-    resource_class: << parameters.resource-class >>
-    environment:
-      FERROCENE_HOST: aarch64-unknown-linux-gnu
-      SCRIPT: |
-        # See ferrocene/ci/split-tasks.py for a list of tasks executed by this.
-        ./x.py --stage 2 test $(ferrocene/ci/split-tasks.py << parameters.job >>) --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
-    steps:
-      - ferrocene-job-test-container:
-          restore-from-job: aarch64-linux-build
```
jyn514, line 476: noting that this is a change from https://github.com/ferrocene/ferrocene/pull/2484/ ("Disable aarch64-linux-library-coverage")

.circleci/workflows.yml
```diff
@@ -127,16 +132,16 @@ jobs:
...
     resource_class: ferrocene/k8s-amd64-large
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
-      FERROCENE_TARGETS: aarch64-unknown-linux-gnu
+      FERROCENE_TARGETS: << pipeline.parameters.targets--x86_64-unknown-linux-gnu--build >>
       SCRIPT: |
         ferrocene/ci/scripts/llvm_cache.py download
-        ./x.py --stage 2 build library src/tools/rustdoc
+        ./x.py --stage 2 build cargo library src/tools/rustdoc symbol-report
+        ./x.py --stage 1 build compiletest coverage-dump src/tools/rustdoc src/tools/run-make-support
```
jyn514, line 144: noting that this is a change from https://github.com/ferrocene/ferrocene/pull/2484/ ("Combine test jobs for different variants of the same target")

.circleci/workflows.yml
```diff
@@ -1088,78 +1006,111 @@ workflows:
       - x86_64-linux-dist-src:
           requires:
             - x86_64-linux-build
-      - x86_64-linux-generic-test-container:
+      - linux-generic-test-container:
           name: x86_64-linux-test
+          executor: *executor-x86_64-runner
           job: test
-          resource-class: ferrocene/k8s-amd64-large
+          resource-class: ferrocene/k8s-amd64-medium
```
jyn514, line 1013: why this change?

.circleci/workflows.yml
```diff
@@ -1088,78 +1006,111 @@ workflows:
...
           test-variant: "2021"
-          requires:
-            - x86_64-linux-build
-      - x86_64-linux-generic-test-container:
+          ferrocene_host: x86_64-unknown-linux-gnu
+          restore_from_job: x86_64-linux-build
+          requires: [ x86_64-linux-build ]
```
jyn514, line 1017: it makes my job as reviewer harder when you combine semantic changes with reformatting.

jyn514, line 1017: nit: maybe instead of making x86_64-linux-generic-test-container a fully generic `linux-generic-test-container`, you could add `linux-generic-test-container` as the base and have `x86_64-linux-generic-test-container` reuse it? i don't know if circleCI supports that, but if so it would avoid having to change all these callsites.

ferrocene/ci/docker-images/force_rebuild
```diff
@@ -2,4 +2,4 @@
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
 # increment this to force a rebuild
-GENERATION 1
\ No newline at end of file
+GENERATION 3
```
jyn514, line 5: should probably be 2?
```suggestion
GENERATION 2
```

ferrocene/ci/scripts/build-and-push-docker-image.sh
```diff
@@ -28,5 +28,5 @@ aws ecr get-login-password --region "${ECR_REGION}" \
 ferrocene/ci/scripts/ensure-ubuntu-20-packages.sh
 
 # Build and push the image
-docker build --progress=plain -t "${image}" -f "ferrocene/ci/docker-images/${IMAGE_NAME}/Dockerfile" .
+docker build --progress=plain -t "${image}" --build-arg BASE_IMAGE=${BASE_IMAGE:-""} -f "ferrocene/ci/docker-images/${IMAGE_NAME}/Dockerfile" .
```
jyn514: are you sure this behaves correctly if no BASE_IMAGE is set in the environment?

ferrocene/ci/split-tasks.py
```diff
@@ -163,6 +163,7 @@ def __lt__(self, other):
                 "coverage-run-rustdoc",
                 "pretty",
                 "rustdoc",
+                "rustdoc-html",
```
jyn514, line 166: noting that this is from https://github.com/ferrocene/ferrocene/pull/2484/ ("Add tests/rustdoc-html to the list of host-only tests which should not be run when testing with target != host").

library/alloc/src/vec/mod.rs
jyn514, line 253: rather than adding dozens of these annotations, can we ignore them in one central place? rustdoc libtest runners have a `--exclude-should-panic` flag, you should be able to set that in bootstrap:
```
$ cargo test --quiet --doc -- --exclude-should-panic -Zunstable-options

running 0 tests
$ cargo test --quiet --doc

running 1 test
src/lib.rs - (line 1) --- FAILED
```

src/bootstrap/src/core/builder/mod.rs
```diff
@@ -878,7 +878,8 @@ impl<'a> Builder<'a> {
                 tool::LlvmBitcodeLinker,
                 tool::RustcPerf,
                 tool::WasmComponentLd,
-                tool::LldWrapper
+                tool::LldWrapper,
+                tool::RunMakeSupport,
```
jyn514, line 882: noting that this is a change from https://github.com/ferrocene/ferrocene/pull/2484/ ("the test jobs need a lot of host tools which were not being built by the {x86_64-linux, aarch64-linux, aarch64-darwin}-build jobs").

jyn514 (dismissed):
i'm a little hesitant to make this change because this all feels quite fragile, but well, binfmt was fragile too. i'm ok with throwing it at CI and seeing what flakes in the next few weeks.

some comments on the implementation below.

jyn514:
oh and i remember that @jonathanpallant was working on a fork of QEMU — did we end up needing that? did you manage to get it working without?

jyn514:
right, i think that's in the executor, not this PR, which is why i didn't see it here ... could you point me to the PR that added it?

.circleci/workflows.yml
Xylakant: As part of the implementation, I did look at resource usage of the various runs. This job never uses the memory of the larger runner and very rarely uses the CPU allocation. In effect, it's barely slower with a medium runner for a much better resource efficiency.

.circleci/workflows.yml
Xylakant: It doesn't. The only option is full duplication.

ferrocene/ci/docker-images/force_rebuild
Xylakant: The number doesn't matter. It just needs to change the dockerfile since we currently don't have any other mechanism to force a rebuild of the docker image.

ferrocene/ci/scripts/build-and-push-docker-image.sh
Xylakant: It will obviously error in the case that the image you're building requires a base image to be set, but it does not error if the env var is unset and the image does not require a base image.

However, I believe this might be leftover from an intermediary step where we'd produce a per-run testserver image and might not be needed any longer.

Xylakant:
> oh and i remember that @jonathanpallant was working on a fork of QEMU — did we end up needing that? did you manage to get it working without?

This uses https://github.com/ferrocene/qemu/tree/ferrocene/release/11.0.1-patched, the relevant changeset is https://github.com/ferrocene/qemu/commit/5b22993d7fafa4131d2a41ef7eca922f82e8d259

The emulator binary is provided by the CI runner.

.circleci/workflows.yml
jyn514: (this was github's fault, there's no reformat here.)

library/alloc/src/vec/mod.rs
jyn514: you asked internally what the right place for this is -- i'm not sure exactly, but my guess would be somewhere around `Builder::cargo` in `src/bootstrap/src/core/builder/cargo.rs`. look for `rustdocflags`.

fs-rachel:
Caught up on the changes since last time. It's difficult to review a patchset this large, especially when the same line is modified multiple times in quick succession by different commits. Some judicious use of `git rebase -i` to squash related groups of changes into a single commit would help a lot.

That said, as far as I can tell at the moment this is all reasonable, and it would be good to get it merged soon

.circleci/workflows.yml
```diff
@@ -86,6 +86,11 @@ parameters:
     type: string
     default: ""
 
+  # The qemu-version to run emulated tests in (read from ferrocene/ci/qemu-version by calcualate-params.py)
+  qemu-version:
+    type: string
+    default: ""
```
dotcarmen, line 92: is the default necessary? this param should always be populated by calculate-params.py right? we probably want the failure to be loud if it doesnt

.circleci/workflows.yml
dotcarmen: are you sure that's not a result of the env var typo? i'd be surprised if any of these tests are being picked up as `EMULATED=true` right now considering the orb is setting `EMULTATED=true`

dotcarmen (changes requested):
still need the typo fixed in order for the emulation to work

ferrocene/ci/docker-images/emulator/run-testserver.sh
```diff
@@ -0,0 +1,42 @@
...
```
Hoverbear, line 17: I mildly worry about this permanently hanging without a timeout, normally we could rely on circle but the echo would prevent that.

Hoverbear (commented):
I think most of my concerns have been raised by others, I have no major blocking concerns.

I do think we might need to update https://public-docs.ferrocene.dev/main/qualification/internal-procedures/working-with-the-ci.html#reproducing-ci-jobs soon to capture things like the orb.

ferrocene/ci/docker-images/emulator/run-testserver.sh
```diff
@@ -0,0 +1,42 @@
...
+  echo "Waiting until testserver binary is found in /shared/testserver/remote-test-server"
+  sleep 1
+done
```
dotcarmen, line 20: related to ana's mention of circleci's timeout not working with the echo

```suggestion
echo "Waiting until testserver binary is found in /shared/testserver/remote-test-server..."
until [ -e /shared/testserver/remote-test-server  ]; do
  sleep 1
done
```

dotcarmen: wow, that's a bad diff preview...

.circleci/workflows.yml
Xylakant: They are, the relevant part is here https://github.com/ferrocene/circleci-orb-ferrocene-common/blob/main/src/executors/facade_testserver.yaml#L50 and not the env var which is mistyped indeed. Env vars do not affect the service container, so the relevant parameters get passed directly to the script.

ferrocene/ci/docker-images/emulator/run-testserver.sh
Xylakant: No, that's deliberate. It should print that line every second so you can tell when it it's still waiting. CircleCIs timeout won't be affected since the print happens in a service container and not in the main runner.

ferrocene/ci/docker-images/emulator/run-testserver.sh
Xylakant: No, the print happens in a service container which doesn't affect the main ci runners lifetime. It gets killed the moment the main container terminates or hangs beyond the configured timeout.

.circleci/workflows.yml
Xylakant: It's necessary - that's why all of our parameters have a default value. I haven't felt like figuring out whether circleci or calculate-parameters is the culprit, but things fail if you don't have a default.

library/alloc/src/vec/mod.rs
Xylakant: I have set this via the RUSTDOCFLAGS env var for all facade test jobs

ferrocene/ci/docker-images/emulator/run-testserver.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+if [[ "" != "$EMULATOR_CPU" ]]; then
+    export QEMU_CPU="${EMULATOR_CPU}" # we need to export the variable to make child processes use it
+fi
+
+# restart the testserver continously if it fails, but not if it exits gracefully
+while true; do
+    ${EMULATOR} \
+    /shared/testserver/remote-test-server -v --bind $TEST_DEVICE_ADDR  && break;
```
dotcarmen, line 41: i'd prefer this, but it can be done as follow-up, no need to block:
```suggestion
    if ${EMULATOR} /shared/testserver/remote-test-server -v --bind $TEST_DEVICE_ADDR; then
        break;
    fi
```

at least it'd be nice to not have the arguments be at the same indentation level as the emulator, to make it obvious that `${EMULATOR}`:
1. may have additional arguments
2. is the wrapper, not an environment variable being set

dotcarmen (approved):
lgtm

ferrocene/ci/docker-images/emulator/run-testserver.sh
```diff
@@ -0,0 +1,42 @@
...
+env
+
+until [ -e /shared/testserver/remote-test-server  ]; do
+  echo "Waiting until testserver binary is found in /shared/testserver/remote-test-server"
+  sleep 1
+done
+echo "Found testserver binary, proceeding"
+
+# Wait some in case copying takes a few seconds
+sleep 5
```
fs-rachel, line 24: I am a bit wary of this, would be better to upload a second file as a "done" marker and wait on that instead. But happy for that to be handled in a separate PR.

fs-rachel (approved):
Also LGTM

Xylakant:
bors merge

PR #2448, opened 2026-07-06
Remove gcc submodule and mark GCC codegen backend as unsupported

fs-rachel:


fs-rachel:
Current status:

* CI passes on an earlier version of this patch, the only difference to the current patch is that it had broad `#[allow(dead_code)]` annotations instead of tightly-scoped `#[expect(dead_code)]` annotations

* The build succeeds on my machine with the current patch

So this is ready for review

src/bootstrap/src/core/builder/mod.rs
```diff
@@ -781,7 +782,8 @@ impl<'a> Builder<'a> {
                 compile::Rustc,
                 compile::Assemble,
                 compile::CraneliftCodegenBackend,
-                compile::GccCodegenBackend,
+                // Ferrocene modification: Dummy out gcc backend as we do not support it
+                //compile::GccCodegenBackend,
```
jyn514: you've disabled this in two places, once here and once in the `Step` itself where you call `exit!`. that means that we'll never actually see the warning you wrote. can you double-check whether that's the case?

if so, the fix is just to uncomment this.

src/bootstrap/src/core/build_steps/clippy.rs
```diff
@@ -350,7 +354,13 @@ impl Step for CodegenGcc {
         builder.ensure(CodegenGcc::new(builder, run.target, config));
     }
 
+    #[allow(unreachable_code)]
+    #[allow(unused_variables)]
     fn run(self, builder: &Builder<'_>) -> Self::Output {
+        // Ferrocene addition
+        eprintln!("error: Ferrocene does not support the gcc codegen backend");
+        exit!(1);
```
jyn514: nit: i *think* you can move this to `codegen_backend_stamp` so you don't have to duplicate this everywhere. i don't feel strongly though, the current code is fine.

jyn514 (commented):
one thing i'd like you to check, one possible cleanup that's non-blocking. overall looks good :)

src/bootstrap/src/core/builder/mod.rs
fs-rachel: My thought was that it provides a second layer of defense against some future change making it possible to use again. But if I move this message to `codegen_backend_stamp` like you mentioned, that should serve the same purpose with less duplication

src/bootstrap/src/core/build_steps/clippy.rs
fs-rachel: Done

src/bootstrap/src/core/builder/mod.rs
jyn514: concretely, i mean that `./x check rustc_codegen_gcc` says "ERROR: no `check` rules matched [rustc_codegen_gcc]" rather than "Ferrocene does not support the gcc codegen backend". i think that's *probably* ok, but in that case we can delete the new build stamp changes.

my preference is still to uncomment these structs so we have fewer divergences from upstream.

fs-rachel: I commented those out to fix the `check_cross_compile` test, and that caused a bunch of knockon consequences. But looking at it again, I've found a better solution which doesn't need all this faff.

fs-rachel:
Fixed the test failures I originally had in a better way, and removed the commits which were superceded by other commits. `./x test bootstrap` passes locally

src/bootstrap/src/core/build_steps/gcc.rs
```diff
@@ -80,7 +80,10 @@ impl Step for Gcc {
     const IS_HOST: bool = true;
 
     fn should_run(run: ShouldRun<'_>) -> ShouldRun<'_> {
-        run.path("src/gcc").alias("gcc")
+        // Ferrocene modification: As we have removed the gcc submodule,
+        // `run.path("src/gcc")` would error out. So drop that part
+        run.alias("gcc")
+        //run.path("src/gcc").alias("gcc")
```
jyn514:
```suggestion
        // `run.path("src/gcc")` would error out. Use `alias` instead.
        run.alias("src/gcc").alias("gcc")
```

src/tools/tidy/src/main.rs
```diff
@@ -108,7 +108,7 @@ fn main() {
         // Checks that only make sense for the compiler.
         check!(error_codes, &root_path, &[&compiler_path, &librustdoc_path]);
         check!(target_policy, &root_path);
-        check!(gcc_submodule, &root_path, &compiler_path);
+        //check!(gcc_submodule, &root_path, &compiler_path);
```
jyn514, line 112:
```suggestion
        // Ferrocene addition
        //check!(gcc_submodule, &root_path, &compiler_path);
```

jyn514 (approved):
one nit, looks good to me with or without the suggested change. you don't need to ask for re-approval if you accept the change.

fs-rachel:
Accepted both suggestions

bors merge

PR #2451, opened 2026-07-07
Sync dependencies between linkchecker and dist tarball

fs-rachel:


src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -200,16 +201,30 @@ You can skip linkcheck with --skip src/tools/linkchecker"
             return;
         }
 
-        // Build all the default documentation.
+        // Build all of the documentation.
         builder.run_default_doc_steps();
+        let doc_out = builder.doc_out(host);
+
+        builder.ensure(Rustc::for_stage(builder, builder.top_stage, host));
+        //let compiler_doc_out = builder.compiler_doc_out(host); // For the future
```
jyn514: what's this "for the future" thing? is this about another PR in flight?

src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -200,16 +201,30 @@ You can skip linkcheck with --skip src/tools/linkchecker"
...
+        }
 
         // Build the linkchecker before calling `msg`, since GHA doesn't support nested groups.
         let linkchecker = builder.tool_cmd(Tool::Linkchecker);
 
         // Run the linkchecker.
         let _guard = builder.msg_test("Linkcheck", bootstrap_host, 1);
         let _time = helpers::timeit(builder);
-        linkchecker.delay_failure().arg(builder.out.join(host).join("doc")).run(builder);
+        linkchecker.delay_failure().arg(doc_out).run(builder);
```
jyn514: why this change? i don't see `doc_out` used earlier.

jyn514 (commented):
the idea makes sense to me and i think this is "just" a bug. i'm confused that we never hit it in CI though ... could you spend a bit of time investigating whether we ever run the linkchecker?

src/bootstrap/src/core/build_steps/test.rs
fs-rachel: This and moving doc_out into a different variable are to prepare for when we add links to the compiler docs. At the moment, they're built separately to the main docs and only get merged during the dist tarball build, so we'll need to mirror that logic across to the linkchecker to keep it happy.

But happy to remove this for now and put it back in when it's needed, if you'd prefer

fs-rachel:
> i'm confused that we never hit it in CI though ... could you spend a bit of time investigating whether we ever run the linkchecker?

The CI does run the linkchecker (it's under `ferrocene/ci/split-tasks.py test:docs`). But just before that, it fetches the coverage outcomes from a previous CI step, so they're already present

fs-rachel:
Also the CI failure here is bizarre: the doctests failed, only on `thumbv7em-ferrocene.facade-eabihf`, and not because of a specific test failing but because

`error: io error when listing tests: Os { code: 14, kind: Uncategorized, message: "Bad address" }`

Not sure what to make of that, I'll rerun after updating the patchset (if needed) and see if it reoccurs

src/bootstrap/src/core/build_steps/test.rs
jyn514: yeah, let's keep the code the same and add a comment? that way this doesn't show as a change in the git history if i need to bisect it later.

jyn514:
> error: io error when listing tests: Os { code: 14, kind: Uncategorized, message: "Bad address" }

this is probably something to do with remote-test-server :/ it's less flaky than it used to be but still unreliable.

this LGTM other than the one comment i left above.

fs-rachel:
@jyn514 How about this version?

Xylakant:
> > error: io error when listing tests: Os { code: 14, kind: Uncategorized, message: "Bad address" }
> 
> this is probably something to do with remote-test-server :/ it's less flaky than it used to be but still unreliable.

This test is not (yet) running via remote-testserver. It does run under emulation, though, so qemu might be a culprit.

jyn514 (approved):
bors merge

PR #2454, opened 2026-07-08
Restore QNX8 jobs

Hoverbear:
Reverses #2440

PR #2457, opened 2026-07-08
QNX 8 docs bump & new schema

Hoverbear:
Should be after https://github.com/ferrocene/ferrocene/pull/2454

We're swapping to a new, better scheme for QNX 8 deployments.

### DIFF
--- a/ferrocene/doc/internal-procedures/src/partners/qnx.rst
+++ b/ferrocene/doc/internal-procedures/src/partners/qnx.rst
@@ -338,52 +338,65 @@ provided by QNX for our CI/CD.
 To create the deployment, first, ensure you are using a myQNX account with a
 **QNX Software Development Platform Subscription - Build Server License** deployed to it.
 
-Create a deployment containing Linux and Windows toolchains:
+Create deployments containing Linux and Windows toolchains:
 
 .. code-block::
 
     LICENSE_KEY="FILL_ME_IN"
     QNX_USER="FILL_ME_IN"
     QNX_PASSWORD="FILL_ME_IN"
-    QNX_VERSION="8.0.0.00141T202311271501L"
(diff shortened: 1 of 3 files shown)


jyn514:
i can't really review this effectively. @japaric would you mind taking a look?

ferrocene/doc/internal-procedures/src/partners/qnx.rst
```diff
@@ -338,58 +338,72 @@ provided by QNX for our CI/CD.
...
+        -destination qnx/qnx-8.0.0-baseline
     qnx/qnxsoftwarecenter/qnxsoftwarecenter_clt \
-            -installBaseline com.qnx.qnx800/$QNX_VERSION \
-            -installPackage com.qnx.qnx800.host.win.x86_64=$QNX_WINDOWS_HOST_VERSION \
-            -installPackage com.qnx.qnx800.host.linux.x86_64=$QNX_LINUX_HOST_VERSION \
-            -installPackage com.qnx.qnx800.target.qemuvirt/0.2.1.00087T202507241124L \
-            -installPackage com.qnx.qnx800.target.driver.virtio.devc/0.1.1.00011T202411230100L \
-            -destination qnx/qnx800-141 \
-            -cleanInstall
+        -installBaseline com.qnx.qnx800/8.0.0.00141T202311271501L \
```
japaric, line 356: if we don't specify a patchset package this command will bring in the latest versions of all dependencies. the oldest patchset I see in the software center is 8.0.2 so outside of listing all packages and their versions manually in this command line invocation I think we cannot repro a 8.0.0 baseline deployment.

but if we don't need it I think we can just document how to repro the 8.0.5 one until we need a newer one

japaric (changes requested):
left a comment about the 8.0.0 deployment. this PR also needs a rebase

PR #2167, opened 2026-02-18
Add jobs to generate coverage reports in all the platforms

pvdrz:

PR #2459, opened 2026-07-09
Generate a coverage report per target platform

fs-rachel:
Alternative approach to https://github.com/ferrocene/ferrocene/pull/2167

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1267,9 +1267,12 @@ workflows:
             - aarch64r82-none-test-library-compiletest
             - aarch64v8r-none-test-library-compiletest
             - armv7r-eabihf-test-library-compiletest
+            - armv7r-eabihf-library-coverage
             - thumbv7em-eabihf-test-library-compiletest
+            - thumbv7em-eabihf-library-coverage
             - thumbv7em-eabihf-test-target-cpu-library-compiletest
             - thumbv7em-eabi-test-library-compiletest
+            - thumbv7em-eabi-library-coverage
             - thumbv7em-eabi-test-target-cpu-library-compiletest
             - aarch64-linux-test
             - aarch64-linux-test-library
@@ -1425,6 +1428,16 @@ workflows:
             - x86_64-linux-build
 
       - facade-generic-test:
+          name: armv7r-eabihf-library-coverage
+          target: armv7r-ferrocene.facade-eabihf
+          test-variant: 2021-cortex-r5f
+          qemu-arch: qemu-arm
+          tasks: --coverage=library $(ferrocene/ci/split-tasks.py test:library-coverage) --tests
+          resource-class: large # 4-core
+          requires:
+            - x86_64-linux-build
+
+      - facade-generic-test:
           name: thumbv7em-eabihf-test-library-compiletest
           target: thumbv7em-ferrocene.facade-eabihf
           test-variant: 2021-cortex-m4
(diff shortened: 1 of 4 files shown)


jyn514:
this looks fine to me, i'll let Johann review since you've been working closely.

.circleci/workflows.yml
```diff
@@ -882,12 +882,20 @@ jobs:
...
       FERROCENE_HOST: x86_64-unknown-linux-gnu
       FERROCENE_TARGETS: << parameters.target >>
       SCRIPT: |
-        sudo update-binfmts --import && sudo update-binfmts --enable << parameters.qemu-arch >> && ./x.py --stage 2 test << parameters.tasks >> --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
+        sudo update-binfmts --import
+        sudo update-binfmts --enable << parameters.qemu-arch >>
+        ./x.py --stage 2 test << parameters.tasks >> --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
+        if << parameters.run-coverage >>; then
+          ./x.py --stage 2 test --coverage=library $(ferrocene/ci/split-tasks.py test:library-coverage) --tests --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
+        fi
```
Urhengulas: Can this be an if-else so it only runs the tests once?

Urhengulas: Ah no, because we don't have separate coverage jobs but just add the coverage to the compiletest job?

ferrocene/ci/split-tasks.py
```diff
@@ -139,6 +139,9 @@ def __lt__(self, other):
         # This is the library without std, for all targets that have no std.
         "library-nostd": ["library/core", "library/alloc", "library/test"],
 
+        # We currently only generate coverage for a subset of the standard library
```
Urhengulas: I think this is a bit misleading. Right now we only generate coverage for the `core` library. We use the `coretsts` and `alloctests` test suites to get the coverage for the `core` library.
```suggestion
        # We currently only generate coverage for the core library
```

.circleci/workflows.yml
fs-rachel: That's what this latest patch does, I'm experimenting with running the coverage in the compiletest job instead of a separate job to see if it works and gives the same results.

ferrocene/ci/split-tasks.py
fs-rachel: Noted, will change that next time I update the PR

fs-rachel:
@Urhengulas This is ready for re-review now

ferrocene/doc/core-certification/src/safety-report/code-coverage.rst
```diff
@@ -4,4 +4,4 @@
 Code coverage report
 ====================
 
-`Code coverage report <../../../coverage/index.html>`_  of the certified core library, showing the statement coverage.
+`Code coverage report <../../../coverage/x86_64-unknown-linux-gnu/index.html>`_  of the certified core library, showing the statement coverage.
```
Urhengulas: Can you link in the other coverage reports as well please? This way it is easier to find and look at them.

This option of showing it looks a little bit horrible in source but rendered it is pretty nice:

```suggestion
The code coverage report showing the statement coverage is available for following platforms:

.. |cov-x86_64| replace:: :target-with-tuple:`x86_64-unknown-linux-gnu`
.. _cov-x86_64: ../../../coverage/x86_64-unknown-linux-gnu/index.html
- |cov-x86_64|_

.. |cov-aarch64| replace:: :target-with-tuple:`aarch64-unknown-linux-gnu`
.. _cov-aarch64: ../../../coverage/aarch64-unknown-linux-gnu/index.html
- |cov-aarch64|_
```

Image

src/bootstrap/src/ferrocene/doc/code_coverage.rs
```diff
@@ -6,6 +6,16 @@ use crate::core::builder::{Builder, RunConfig, ShouldRun, Step};
 use crate::core::config::{FerroceneCoverageOutcomes, TargetSelection};
 use crate::ferrocene::code_coverage::CoverageOutcomesDir;
 
+// List of targets that we generate coverage for in CI
+const COVERAGE_TARGET_TRIPLES: [&'static str; 6] = [
```
Urhengulas:
```suggestion
const COVERAGE_TARGET_TUPLES: [&str; 6] = [
```
I think we generally started calling them target tuples instead of triples, because a few of them are not triples in fact...

Urhengulas (changes requested):

ferrocene/doc/core-certification/src/safety-report/code-coverage.rst
fs-rachel: Ah, I didn't know about that particular bit of magic! Yes, I can do that

src/bootstrap/src/ferrocene/doc/code_coverage.rs
Urhengulas: Also the static is not needed in the type

fs-rachel:
CI passed, and the coverage reports look good. Parking for now.

Urhengulas (approved):

PR #2463, opened 2026-07-10
Add per-platform coverage reports to generated docs

fs-rachel:
The reports themselves are already generated by the coverage jobs, we just need to copy them over when we generate the documentation


fs-rachel:
Accidental duplicate of https://github.com/ferrocene/ferrocene/pull/2459 , closing in favour of that one

PR #2466, opened 2026-07-13
fix a bunch of bugs in release automation

jyn514:
See individual commit messages.

### DIFF
--- a/ferrocene/tools/automations-common/src/automations_common/automated_prs.py
+++ b/ferrocene/tools/automations-common/src/automations_common/automated_prs.py
@@ -116,7 +116,7 @@ class AutomatedPR(abc.ABC):
         self.current_branch = self.cmd_capture(["git", "branch", "--show-current"])
         self.current_hash = self.cmd_capture(["git", "rev-parse", "HEAD"])
 
-        existing_pull = self.__find_open("pulls", self.pr_title(), self.pr_labels())
+        existing_pull = self.__find_open("pulls", self.__resolved_pr_title(), self.pr_labels())
         if existing_pull is not None:
             log("An automated PR is already open, a new one won't be created.")
             log(f"==> {existing_pull['html_url']}")
@@ -159,7 +159,7 @@ class AutomatedPR(abc.ABC):
             response = self.http.post(
                 self.__repo_api("pulls"),
                 json={
-                    "title": self.pr_title(),
+                    "title": self.__resolved_pr_title(),
                     "head": branch_name,
                     "base": self.base_branch(),
                     "body": self.pr_body(branch_name),
@@ -287,6 +287,12 @@ class AutomatedPR(abc.ABC):
     #     Internals     #
     #                   #
     #####################
+
+    def __resolved_pr_title(self):
+        if self.base_branch() == 'main':
+            return self.pr_title()
+        version = self.base_branch().removeprefix("release/")
+        return f"[{version}] {self.pr_title()}"
 
     def __find_open(self, kind, expected_title, expected_labels):
         """
--- a/ferrocene/tools/automations-common/src/automations_common/automated_prs.py
+++ b/ferrocene/tools/automations-common/src/automations_common/automated_prs.py
@@ -107,7 +107,7 @@ class AutomatedPR(abc.ABC):
         Handle the creation of the PR, and open an issue if an error occurs.
         """
         self.dry_run = dry_run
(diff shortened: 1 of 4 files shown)


ferrocene/tools/pull-upstream/create-release-branch.sh
```diff
@@ -18,7 +18,7 @@ IFS=$'\n\t'
 
 UPSTREAM_REPO="rust-lang/rust"
 VERSION_FILE="src/version"
-FERROCENE_ORIGIN="origin"
+: "${FERROCENE_ORIGIN:=origin}"
```
dotcarmen: this syntax isn't great.

```suggestion
FERROCENE_ORIGIN="${FERROCENE_ORIGIN:-origin}"
```

ferrocene/tools/pull-upstream/create-release-branch.sh
```diff
@@ -98,7 +107,7 @@ else
     fi
     git branch "${branch_name}" "${last_commit}"
 
-    git push "${FERROCENE_ORIGIN}" "${branch_name}"
+    git push --no-verify "${FERROCENE_ORIGIN}" "${branch_name}"
```
dotcarmen: as a comment, this is to prevent pre-push hooks from running. seems fine.

dotcarmen (approved):
lgtm. would prefer you accept my suggestion, but not a blocker

jyn514:
bors p=-1

ferrocene/tools/automations-common/src/automations_common/automated_prs.py
```diff
@@ -107,7 +107,7 @@ def create(self, dry_run=False):
         Handle the creation of the PR, and open an issue if an error occurs.
         """
         self.dry_run = dry_run
-        self.origin = ORIGIN
+        self.origin = os.environ.get("FERROCENE_REMOTE", ORIGIN)
```
dotcarmen: i didn't catch this before; `os.environ.get` [isn't documented](https://docs.python.org/3/library/os.html#os.environ). `.get()` seems to be a method on dicts, whereas `os.environ` is only guaranteed to be an `abc.Mapping`, which explicitly [does not have anything in its interface](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping)

please use [`os.getenv`](https://docs.python.org/3/library/os.html#os.getenv) instead

(not a blocker if this line works, but i do prefer to use actually-documented functions... especially in python land where things break so regularly)

ferrocene/tools/backport/one.py
```diff
@@ -149,21 +149,28 @@ def main():
...
-    authors = git_output("log", "--format=%aN <%aE>", f"{base}..{head}").splitlines()
+    pr_target, head = metadata["base"]["sha"], metadata["head"]["sha"]
+
+    git("fetch", "--quiet", f"https://github.com/{repository}", pr_target, head)
+
+    # NOTE: you almost never want the pr_target; you want the "logical diff" between the
+    # merge base and the PR HEAD, which we calculate here.
+    # The difference is that this `merge-base` allows the PR to not be fully rebased
+    # against `main` HEAD.
+    merge_base = git_output("merge-base", pr_target, head).rstrip()
```
dotcarmen, line 160: not familiar enough with these shenanigans; i'll trust this is correct

ferrocene/tools/pull-upstream/create-release-branch.sh
```diff
@@ -31,10 +31,19 @@ fetch_upstream_version() {
...
-find_last_commit_with_version() {
+find_last_commit_with_minor_version() {
     version="$1"
     root="$(git rev-parse --show-toplevel)"
 
+    # See if we are currently on the expected version.
+    # If so, we don't know when the next release branches, so we can't return the *last*
+    # commit with the right minor version.
+    # Instead we just return the current commit.
+    if [[ "$(git show "HEAD:${VERSION_FILE}")" = "${version}."* ]]; then
```
dotcarmen, line 42: huh, til

```shell
bash-3.2$ [[ 'foo.bar' = 'foo'* ]]
bash-3.2$ echo $?
0
bash-3.2$ [[ 'foo.bar' = 'bar'* ]]
bash-3.2$ echo $?
1
```

dotcarmen (approved):
lgtm

jyn514:
bors merge

PR #2468, opened 2026-07-13
qual-report: add facade target mapping to ARMv7r target

japaric:
This is a no-std target so the test outcomes will be associated to a facade target. The outcomes need to be mapped back to the no-std target with the `:bare_metal_test_target:` field


jyn514 (approved):
bors merge

Xylakant:
bors retry

PR #2470, opened 2026-07-13
sphinx/test-outcomes: make test target mandatory

japaric:
see commit message for rationale. also check the qualification report target pages in tidy

### DIFF
--- a/ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
+++ b/ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
@@ -273,7 +273,7 @@ also accepts multiple options:
 
 * ``host`` (required): the target tuple of the host platform
 * ``target`` (required): the target tuple of the compilation target
-* ``library_compiletest_target`` (required): the target tuple of the target
+* ``tested_target_with_std`` (required): the target tuple of the target
   used to run library and compiler tests. For targets that have a standard
   library this is the same tuple as the compilation targets; for targets
   without a standard library it will be a facade target
@@ -290,7 +290,7 @@ also accepts multiple options:
    .. render-outcomes-template:: templates/tests.jinja2
       :host: x86_64-unknown-linux-gnu
       :target: aarch64-unknown-none
-      :library_compiletest_target: aarch64-unknown-ferrocene.facade
+      :tested_target_with_std: aarch64-unknown-ferrocene.facade
       :remote_testing:
 
 Rendering a summary of all test outcome pages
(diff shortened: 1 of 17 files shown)


ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
```diff
@@ -273,9 +273,10 @@ also accepts multiple options:
 
 * ``host`` (required): the target tuple of the host platform
 * ``target`` (required): the target tuple of the compilation target
-* ``bare_metal_test_target`` (optional): the target tuple of the special
-  target used for bare metal testing; it should be omitted if no special target
-  was used
+* ``library_compiletest_target`` (required): the target tuple of the target
+  used to run library and compiler tests. For targets that have a standard
+  library this is the same tuple as the compilation targets; for targets
+  without a standard library it will be a facade target
```
jyn514: i find the name of this option hard to understand -- if you don't like the "bare metal" part, can we just call it "tested_target" to match render_template.py?

jyn514 (commented):
this is lovely, thank you :)

ferrocene/doc/internal-procedures/src/docs/sphinx-extensions.rst
japaric: if I'm modifying the target (`.rst`) pages in the qualification report and I don't know that the sphinx extensions have docs (past me says  :raising_hand_man:) and I see `tested_target` I'll probably just duplicate the `target` field. if the name conveys the idea that the target is used to run the compiler and library tests then that should give me some pause "probably can't run compiler tests on a `-none-` target so I can't duplicate the `aarch64-unknown-none` that I see in `:target:`"

`bare_metal_target` I find even more confusing. if `:target:` is `x86_64-unknown-linux-gnu` then I would probably set `:bare_metal_target:` to `x86_64-unknown-none` or something.

> I don't know that the sphinx extensions have docs

IMO, if you want documentation to be read you need to put in front of people's face at all times, and not tuck it away in some drawer. Hence, I'd like to have self-documenting variable names.

jyn514: can we name it `tested_target_with_std`, then? `library_compiletest_target` reads to me like something that's going to passed to `x test <...>`, not a `--target` value at all.

japaric: doned

Urhengulas (approved):
bors merge

PR #2475, opened 2026-07-14
make the llvm-builder for windows an xlarge machine

Xylakant:
this change chops 50 minutes of the llvm build. While the runner is nearly twice as expensive as the large runner, cutting the time in half makes up for the cost increase


jyn514 (approved):
this seems like a straightforward win. thanks Felix :)

Xylakant:
bors merge

PR #2481, opened 2026-07-15
[1.97] edition lift

jyn514:
first and last commit are hand-written, middle commit is auto-generated by edition-lift.py. i'll be making PRs to main and upstream shortly but let's get the release ready for the assessor first.

the tait-normalize.rs change shouldn't have been necessary, but rustc crashes on it in 2021 edition: https://github.com/rust-lang/rust/issues/159324

PR #2482, opened 2026-07-15
fix `--test-variants=2021` to actually use 2021 edition

jyn514:
previously, we did this weird thing called an "edition lift" where we would bless all the tests as a one-off. the conflicts are relatively minor (mostly; the major ones are landing in https://github.com/rust-lang/rust/pull/159330), so let's just accept the merge conflicts in exchange for not having to add more steps to the release process.

this is a direct forward port of https://github.com/ferrocene/ferrocene/pull/2481.


Hoverbear (commented):
This looks good and I think it makes sense to do it! It has some conflicts but I'd be happy to review once those are resolved.

PR #2484, opened 2026-07-15
CI optimization tests

fs-rachel:
I ran the following experiments to see if we can remove some redundant work from our CI workflow:

1) Disable aarch64-linux-library-coverage: Currently this job is run every CI run, but its results are not used. This can be re-enabled when we merge https://github.com/ferrocene/ferrocene/pull/2459 . Saves about 20 machine-minutes per run.

2) Use a partial checkout of LLVM for facade-generic-test jobs: Saves 25s * 9 jobs = 3.75 machine-minutes per run.

3) Add tests/rustdoc-html to the list of host-only tests which should not be run when testing with target != host: Saves about 80 machine-minutes per run across 13 jobs

4) Disable profiling for non-coverage runs: Did not save an appreciable amount of time

5) Combine test jobs for different variants of the same target: Saves about 3 * 15 minutes = 45 machine-minutes per run. The combined runs are still shorter than the QNX tests, so do not increase the critical path length. However, we decided against this in its current form to avoid potential interference between the two test sets.

6) On looking at (5) more closely, it turns out that the test jobs need a lot of host tools which were not being built by the {x86_64-linux, aarch64-linux, aarch64-darwin}-build jobs. So these were rebuilt per test job, causing a lot of redundant work. Moving these into the build jobs makes those jobs take 5-6 minutes longer each, but is more than compensated by time savings on the corresponding host=target test jobs. It also saves 9 * 12 minutes on the facade-generic-test jobs and 4 * 15 minutes on the qnx test jobs (relative to changes 1+2+3), for a saving of another 2.5 machine-hours or so.

This patch applies changes 1+2+3+6, with a total savings north of 4 machine-hours per CI run. It also shortens the critical path length (which goes through the qnx test jobs) by ~20 minutes, bringing it close to even with the path through the aarch64-linux jobs.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -561,21 +561,21 @@ jobs:
           name: Download dist artifacts and run the self-test tool
           command: ferrocene/ci/scripts/run-self-test.sh
 
-  aarch64-linux-library-coverage:
-    executor:
-      name: docker-custom-image
-      image-tag: << pipeline.parameters.docker-image-tag--aarch64--ubuntu-20 >>
-    resource_class: ferrocene/k8s-arm64-large
-    environment:
-      FERROCENE_HOST: aarch64-unknown-linux-gnu
-      SCRIPT: |
-        ferrocene/ci/scripts/llvm_cache.py download
-        ./x.py test --stage=1 --coverage=library library/core library/alloc --tests
-    steps:
-      - aws-oidc-auth
-      - ferrocene-checkout:
-          llvm-subset: true
-      - ferrocene-ci
+#  aarch64-linux-library-coverage:
+#    executor:
+#      name: docker-custom-image
+#      image-tag: << pipeline.parameters.docker-image-tag--aarch64--ubuntu-20 >>
+#    resource_class: ferrocene/k8s-arm64-large
+#    environment:
+#      FERROCENE_HOST: aarch64-unknown-linux-gnu
+#      SCRIPT: |
+#        ferrocene/ci/scripts/llvm_cache.py download
+#        ./x.py test --stage=1 --coverage=library library/core library/alloc --tests
+#    steps:
+#      - aws-oidc-auth
+#      - ferrocene-checkout:
+#          llvm-subset: true
+#      - ferrocene-ci
 
   aarch64-linux-rhivos2-test-container:
     executor: docker-aarch64-ubuntu-20-rhivos2-test-server
@@ -1273,7 +1273,7 @@ workflows:
             - aarch64-linux-test
             - aarch64-linux-test-library
             - aarch64-linux-compiletest
-            - aarch64-linux-library-coverage
+            #- aarch64-linux-library-coverage
             - aarch64-linux-rhivos2-test-library-compiletest
       - x86_64-linux-traceability-matrix:
           requires: *test-outcomes-dependencies
@@ -1506,9 +1506,9 @@ workflows:
           test-variant: "2021"
           requires:
             - aarch64-linux-build
-      - aarch64-linux-library-coverage:
-          requires:
-            - aarch64-linux-build
+      #- aarch64-linux-library-coverage:
+      #    requires:
+      #      - aarch64-linux-build
       - aarch64-linux-self-test:
           requires:
             - aarch64-linux-dist
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -169,7 +169,8 @@ jobs:
       FERROCENE_TARGETS: aarch64-unknown-linux-gnu
       SCRIPT: |
         ferrocene/ci/scripts/llvm_cache.py download
-        ./x.py --stage 2 build library src/tools/rustdoc
+        ./x.py --stage 2 build cargo library src/tools/rustdoc symbol-report
+        ./x.py --stage 1 build compiletest coverage-dump src/tools/rustdoc src/tools/run-make-support
     steps: [ferrocene-job-build]
 
   x86_64-linux-docs:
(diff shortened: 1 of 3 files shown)


ferrocene/ci/split-tasks.py
```diff
@@ -163,6 +163,7 @@ def __lt__(self, other):
                 "coverage-run-rustdoc",
                 "pretty",
                 "rustdoc",
```
Xylakant, line 165: We'd need to have some sort of argument why this doesn't reduce the test coverage. We're generally interested in running as many tests as reasonably possible.

.circleci/workflows.yml
```diff
@@ -561,21 +561,21 @@ jobs:
...
-      FERROCENE_HOST: aarch64-unknown-linux-gnu
-      SCRIPT: |
-        ferrocene/ci/scripts/llvm_cache.py download
-        ./x.py test --stage=1 --coverage=library library/core library/alloc --tests
-    steps:
-      - aws-oidc-auth
-      - ferrocene-checkout:
-          llvm-subset: true
-      - ferrocene-ci
+#  aarch64-linux-library-coverage:
```
Xylakant, line 565: This is kinda neutral, since this job runs on our own hardware which would be idling if there's no job to run. I have no strong opinion on this.

.circleci/workflows.yml
```diff
@@ -890,7 +890,64 @@ jobs:
         sudo update-binfmts --import && sudo update-binfmts --enable << parameters.qemu-arch >> && ./x.py --stage 2 test << parameters.tasks >> --ferrocene-test-one-crate-per-cargo-call --test-variant << parameters.test-variant >>
     steps:
       - aws-oidc-auth
-      - ferrocene-checkout
+      - ferrocene-checkout:
```
Xylakant, line 894: I like using a subset, the change is also beneficial in terms of reducing network bandwidth.

.circleci/workflows.yml
```diff
@@ -890,7 +890,64 @@ jobs:
...
+              --volume $(pwd):/home/ci/project \
+              --volume /tmp/awsjwt:/tmp/awsjwt \
+              --user $(id -u):$(id -g) \
+              --network host \
+              --privileged \
+              --interactive \
+              --tty \
+              << pipeline.parameters.docker-repository-url--ci-docker-images >>:<< pipeline.parameters.docker-image-tag--x86_64--ubuntu-20 >> \
+
+  facade-generic-test-pair:
```
Xylakant: I like the idea, but not the implementation. What happens if we have three variants, or 4, or 5? Do we have any idea why this is so much faster? Is it re-using compiled artifacts from the previous run? If so, maybe caching the build results for the follow-up job would be an option.

One of our arguments we make in certification is that we run tests in a clean environment with no build residue from earlier runs and this at least bumps against that edge.

Xylakant (changes requested):
Note that build runtimes for the facade tests drop significantly with #2446 which may change the calculus on the changes pairing up the test-variants.

ferrocene/ci/split-tasks.py
fs-rachel: Rustdoc-html, like the other rustdoc-* tests, is marked as `IS_HOST: true` (see src/bootstrap/src/core/build_steps/test.rs). Which means that every single facade test runs `rustdoc-html` on x86_64-linux, regardless of the target platform, and so does not contribute to coverage for the target. Instead that should be done once by x86_64-test-linux

Oddly, on the current main branch, x86_64-linux-test currently skips `tests/rustdoc-html`. I'll have to look into why, but that's the correct place to test it, once.

.circleci/workflows.yml
fs-rachel: The specific structure was just to see if it's viable *at all*; my plan for a final version would be to make facade-generic-test accept a list of test variants to run

The speedup is due to only compiling the test code once per target platform. As far as I can tell, the test variants are handled *at runtime* within the test runner and do not require a full rebuild, so building once per test variant is redundant.

fs-rachel: Also I did a side-by-side comparison of the output for the two test variants, and confirmed that no tests were skipped in the second set. Apart from the rustdoc-html tests, because those run against the host and so don't depend on the test variant.

Xylakant: > The specific structure was just to see if it's viable _at all_; my plan for a final version would be to make facade-generic-test accept a list of test variants to run
> 
> The speedup is due to only compiling the test code once per target platform. As far as I can tell, the test variants are handled _at runtime_ within the test runner and do not require a full rebuild, so building once per test variant is redundant.

This may be true for the test-variants we currently have, but I don't think that's true in the general case. Test-Variants potentially target CPU variants that have additional features (or even lack some features) in which case the generated machine code needs to be different.

ferrocene/ci/split-tasks.py
Xylakant: This makes sense. We just need to remember to write that down :)

.circleci/workflows.yml
fs-rachel: Huh. If that's handled by generating different code, rather than by dispatching to different variants at runtime, then we can't combine jobs for multiple test variants at all. And it isn't useful to combine test runs for different targets, because there's only 1-2 minutes' worth of setup work which could be shared in that case.

ferrocene/ci/split-tasks.py
fs-rachel: Aha! `tests/rustdoc-html` is run by `x86_64-linux-compiletest`, not `x86_64-linux-test`. So we're all good.

Xylakant:
these have been cherry-picked into other PRs and merged. Thank you for the investigation.

PR #2490, opened 2026-07-16
feat: simplify docker build parameters

dotcarmen:
depends on #2430 

the first commit renames the docker images:
- `ubuntu-20-<< specific hash >>-<< arch >>` -> `runner-<< specific hash >>-<< arch >>`
- `ubuntu-24-<< specific hash >>-<< arch >>` -> `emulator-<< specific hash >>-<< arch >>`
- `run-id.<< stable workflow id >>.rhivos-2.0-test-server` -> `rhivos2-emulator--<< stable workflow id >>--<< arch >>`
- deleted unused rhel image

the second commit does a bit more:
- finish the renaming
  - instead of the "specific hash" (which is calculated by only hashing `ferrocene/ci/docker-images/<< image name >>`), uses a more general hash that's shared between all tags in the same workflow (calculated by hashing all of `ferrocene/ci/docker-images` and a few dependencies)
  - also added dashes to better separate the "parts" of the image tag
- eliminates the `docker-image-tag--*` parameters from the workflow
  - new `docker-images-hash` is used as described above
- eliminates the `docker-image-rebuild--*` parameters from the workflow
  - now `llvm-rebuild--*` is reused for this purpose
- `llvm-rebuild--*` parameters are now calculated using the `docker-image-hash`
  - this means that, if the docker images are rebuilt, so is llvm. this is good, since we always want llvm to rebuild in the environment when the environment changes

unfortunately, the "rebuild every 7 days" part has been eliminated as well. this should be restored, somehow


Xylakant (changes requested):
I think we should allow for an x86/aarch64 emulator pair.

ferrocene/ci/scripts/calculate_parameters.py
```diff
@@ -141,27 +137,36 @@ def calculate_docker_image_tag(platform_plus_image: str):
...
+    details = None
     try:
-        image = ecr.describe_images(
-            repositoryName=repo,
-            imageIds=[{"imageTag": calculate_docker_image_tag(f"{platform}--{image}")}],
-        )["imageDetails"][0]
+        details = ecr.describe_images(
+            repositoryName="ci-docker-images",
+            imageIds=[
+                {"imageTag": f"emulator--{images_hash}"},
```
Xylakant: could you add one for emulator-rhivos2? We're currently not rebuilding that regulary since it's essentially 100% stable unless there's an OS update, but if we're touching this, we should add it.

ferrocene/ci/scripts/calculate_parameters.py
```diff
@@ -117,18 +118,13 @@
...
     """
-    platform, image = platform_plus_image.split("--", 1)
-
-    path = os.path.join(DOCKER_IMAGES_PATH, image)
-    if not os.path.exists(os.path.join(path, "Dockerfile")):
-        raise ScriptError(f"unknown Docker image: {image}")
-
     all_files: list[str] = []
-    for root, _, files in os.walk(path):
+    for root, _, files in os.walk(DOCKER_IMAGES_PATH):
```
Xylakant: There is at least the aws-cli.version file that doesn't live in the docker-images folder and should be taken into account. qemu is mounted at runtime, so a change in qemu-version is not relevant.

Xylakant (changes requested):

ferrocene/ci/scripts/calculate_parameters.py
dotcarmen: should i rename `rhivos-2.0-test-server` as well? and `rhel-10.1-test-server` while we're at it?

Xylakant: delete the rhel10, rename the rhivos2.

ferrocene/ci/scripts/calculate_parameters.py
Xylakant: I also found https://github.com/ferrocene/ferrocene/blob/main/ferrocene/ci/mirrors/hashes.txt which has effect on the image and should be included.

ferrocene/ci/scripts/calculate_parameters.py
dotcarmen: i only see aarch64 rhivos being tested. should we also be testing x64?

Xylakant: there's no x86_64 rhivos. It's only aarch64

ferrocene/ci/scripts/calculate_parameters.py
dotcarmen: i included `ferrocene/ci/scripts/ensure-ubuntu-20-packages.sh` as well

.circleci/workflows.yml
```diff
@@ -1340,39 +1319,27 @@ workflows:
...
       - ferrocene/build_docker:
-          name: build-docker-rhivos2-test-server-aarch64
-          image: rhivos-2.0-test-server
-          tag: run-id.<< pipeline.parameters.stable-workflow-id >>.rhivos-2.0-test-server
+          name: build-docker--rhivos2-emulator--aarch64
+          image: rhivos2-emulator
+          tag: rhivos2-emulator--<< pipeline.parameters.stable-workflow-id >>--aarch64
           rebuild: true
-          resource_class: arm.medium
+          resource_class: large.gen2
```
dotcarmen: uh, not sure how this happened

dotcarmen:
i split this PR into 2 commits for easier review

the first commit renames the docker images (except the rhel image, which was deleted)

the second commit eliminates the `docker-image-tag--*` and `docker-image-rebuild--*` parameters from the workflow

.circleci/workflows.yml
```diff
@@ -1340,39 +1319,27 @@ workflows:
...
-          name: build-docker-rhel-test-server-aarch64
-          image: rhel-10.1-test-server
-          tag: run-id.<< pipeline.parameters.stable-workflow-id >>.rhel-10.1-test-server
-          rebuild: true
-          resource_class: arm.medium
-          awscli_version: << pipeline.parameters.awscli-version >>
-          restore_from_job: aarch64-linux-build
-          requires:
-            - aarch64-linux-build
-
```
Xylakant, line 1355: note for other reviewers: We no longer test against rhel, so this has been cleaned up

ferrocene/doc/internal-procedures/src/working-with-the-ci.rst
```diff
@@ -196,12 +196,12 @@ Using the CI Docker images
 Docker images used in CI can be found in ``ferrocene/ci/docker-images``. Each
 image documents how to build it at the top of the ``Dockerfile``.
 
-To build the most common image, ``ubuntu-20``:
+To build the most common image, ``runenr``:
```
Xylakant: typo, must be `runner`

Xylakant (approved):
Currently, all images are named `<arch>-<os>-<hash>`, this changes things so that the arch is at the end of the name. This will lexically sort all images belonging to a function together, before it was for an arch. Neither way is truly better, but it's a notable change. (best for sorting would be starting with the hash, but that makes reading the image names extremely hard)

1 small typo.

dotcarmen:
should also mention, this pr [passed CI](https://app.circleci.com/pipelines/github/ferrocene/ferrocene/19648) (the workflow only failed because i didn't run windows/macos jobs, which shouldn't be affected by this pr)

Xylakant:
> unfortunately, the "rebuild every 7 days" part has been eliminated as well. this should be restored, somehow

This will continue to happen when the images expire in the registry, so we can control it there.

dotcarmen:
bors merge

dotcarmen:
oh, i guess my hashing logic update breaks ci

dotcarmen:
setup step succeeded; this is the only part of the ci functionally affected by the 'cryptographically secure hash' commit

bors merge

dotcarmen:
bors merge

dotcarmen:
> Bors will automatically try to run when all required PR statuses are set.

liar

Xylakant:
bors retry

Xylakant:
bors merge

Xylakant:
bors merge

PR #2493, opened 2026-07-17
Fix channel name displayed by rustc --version

fs-rachel:
Before (displaying the content of `ferrocene/version`):
```
$ ./x install && ../ferrocene-install/bin/rustc --version
rustc 1.99.0-dev (Ferrocene rolling)
```

After (displaying the combined release channel):
```
$ ./x install && ../ferrocene-install/bin/rustc --version
rustc 1.99.0-dev (Ferrocene nightly)
```

### DIFF
--- a/src/bootstrap/src/ferrocene/dist.rs
+++ b/src/bootstrap/src/ferrocene/dist.rs
@@ -497,7 +497,11 @@ impl Step for GenerateBuildMetadata {
             );
         }
 
-        let channel = crate::ferrocene::ferrocene_channel(builder, ferrocene_version);
+        let channel = crate::ferrocene::ferrocene_release_channel(
+            &*builder.config.channel,
+            ferrocene_channel,
+            ferrocene_version,
+        );
 
         // Whenever the contents of this JSON file change, even just adding new fields,
         // make sure to increase the metadata version number and update publish-release
--- a/src/bootstrap/src/ferrocene/mod.rs
+++ b/src/bootstrap/src/ferrocene/mod.rs
@@ -76,8 +76,12 @@ pub(crate) fn download_from_local_filesystem(path: &str, dest: &Path, help_on_er
     }
 }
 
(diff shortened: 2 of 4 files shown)


jonathanpallant:
I previously observed the output to be:

```console
$ criticalup run rustc --version
rustc 1.98.0-nightly (b088ad0bf 2026-06-18) (Ferrocene rolling by Ferrous Systems)
```

The `($HASH $DATE)` I assume only appears when `rustc` is compiled from a git tag. But I'm curious what happened to "by Ferrous Systems"?

Edit: Never mind, that's set [in the bootstrap config file](https://github.com/ferrocene/ferrocene/blob/9189d947de2e0b39b2b888dc5fb4dfe3c711c68a/src/bootstrap/defaults/bootstrap.ferrocene-dist.toml#L54-L59)

fs-rachel:
Aha, yes, I have a minimal bootstrap.toml that doesn't set that field

Urhengulas (changes requested):
Looks good to me. Can you please add some tests?

fs-rachel:
> Looks good to me. Can you please add some tests?

Added a test. As mentioned on Zulip, it's not easy to know where exactly to test: on the one hand, just testing `ferrocene_release_channel()` (the combiner function) seems pointless, but we can't run `rustc --version` without a bunch of other faff. Compromised on testing `Build::rust_version()`, which generates the bulk of the string that `rustc --version` prints.

The test is basically a copy of the existing logic, but with a lot of the annoying parts stripped out so we can at least say "this Rust version string is expected to go with this Ferrocene version string" in one place

Urhengulas (approved):
bors merge

fs-rachel:
Merge conflict with https://github.com/ferrocene/ferrocene/pull/2488

bors r-

fs-rachel:
Rebased + fixed merge conflict with #2488

Urhengulas (approved):
bors merge

fs-rachel:
aarch64 tests hung for reasons which seem unrelated to this patch

bors retry

PR #2496, opened 2026-07-19
Enable local CI cache

Xylakant:
We have a local s3 compatible endpoint with a config automatically injected into the runners if they're local to the cache. Enable using this cache.


Xylakant:
bors try

Xylakant:
bors try

PR #2500, opened 2026-07-21
`ferrocene::prevalidated`: Treat all non-generic annotated functions as mono roots

Urhengulas:
Before only the `main` function and non-generic public functions that are marked with the prevalidated attribute were considered an entry point for the prevalidated analysis. With this change also non-generic private functions are considered entrypoints for the analysis!

### DIFF
--- a/compiler/rustc_monomorphize/src/collector.rs
+++ b/compiler/rustc_monomorphize/src/collector.rs
@@ -1668,32 +1668,26 @@ impl<'v> RootCollector<'_, 'v> {
     }
 
     fn is_root(&self, def_id: LocalDefId) -> bool {
-        // Ferrocene addition: reuse same strategy as Lazy collector
-        // closure so that we don't slow down incremental by calling queries we don't use
-        let is_reachable_lazy = || {
-            self.entry_fn.and_then(|(id, _)| id.as_local()) == Some(def_id)
-                || self.tcx.is_reachable_non_generic(def_id)
-                || {
-                    let flags = self.tcx.codegen_fn_attrs(def_id).flags;
-                    flags.intersects(
-                        CodegenFnAttrFlags::RUSTC_STD_INTERNAL_SYMBOL
-                            | CodegenFnAttrFlags::EXTERNALLY_IMPLEMENTABLE_ITEM,
-                    )
-                }
-        };
-
         !self.tcx.generics_of(def_id).requires_monomorphization(self.tcx)
             && match self.strategy {
                 MonoItemCollectionStrategy::Eager => {
                     !matches!(self.tcx.codegen_fn_attrs(def_id).inline, InlineAttr::Force { .. })
                 }
-                MonoItemCollectionStrategy::Lazy => is_reachable_lazy(),
+                MonoItemCollectionStrategy::Lazy => {
+                    self.entry_fn.and_then(|(id, _)| id.as_local()) == Some(def_id)
+                        || self.tcx.is_reachable_non_generic(def_id)
+                        || {
+                            let flags = self.tcx.codegen_fn_attrs(def_id).flags;
+                            flags.intersects(
+                                CodegenFnAttrFlags::RUSTC_STD_INTERNAL_SYMBOL
+                                    | CodegenFnAttrFlags::EXTERNALLY_IMPLEMENTABLE_ITEM,
+                            )
+                        }
+                }
                 // Ferrocene addition
                 MonoItemCollectionStrategy::Validated => {
-                    let entrypoint = is_reachable_lazy();
                     // Explicit match is intentional, please update this if you add new fields.
-                    entrypoint
-                        && matches!(self.tcx.codegen_fn_attrs(def_id).validated, Some(Validated {}))
+                    matches!(self.tcx.codegen_fn_attrs(def_id).validated, Some(Validated {}))
                 }
             }
     }
--- a/compiler/rustc_monomorphize/src/collector.rs
+++ b/compiler/rustc_monomorphize/src/collector.rs
@@ -1684,7 +1684,8 @@ impl<'v> RootCollector<'_, 'v> {
                             )
                         }
                 }
-                // Ferrocene addition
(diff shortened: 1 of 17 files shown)


compiler/rustc_monomorphize/src/collector.rs
```diff
@@ -1694,14 +1690,17 @@ impl<'v> RootCollector<'_, 'v> {
...
+                    self.is_prevalidated_root(def_id, is_reachable_lazy())
                 }
             }
     }
 
+    // Ferrocene addition
+    fn is_prevalidated_root(&self, def_id: LocalDefId, _is_entrypoint: bool) -> bool {
+        // Explicit match is intentional, please update this if you add new fields.
+        matches!(self.tcx.codegen_fn_attrs(def_id).validated, Some(Validated {}))
+    }
```
jyn514: 👍 this code change looks correct

library/core/src/num/f128.rs
```diff
@@ -1046,6 +1046,8 @@ impl f128 {
     /// assert_eq!((12.5f128).to_bits(), 0x40029000000000000000000000000000);
     /// # }
     /// ```
+    // TODO: why is the caller of this function (impl Debug for f128) not detected?
+    #[ferrocene::prevalidated]
```
jyn514, line 1049: not detected, as in, doesn't warn? because we aren't treating all functions as roots. previously this function was just not validated and you couldn't call it. i think you may have forgotten to regenerate the symbol report after fixing core?

jyn514 (commented):
the compiler changes look correct. i do not believe the *title* of the PR is correct: it should instead be "Treat all **non-generic annotated** functions as mono roots". See `!self.tcx.generics_of(def_id).requires_monomorphization(self.tcx)` on line 1685 ("non-generic") and your new function on 1701 ("annotated").

library/core/src/num/f128.rs
Urhengulas: Not exactly, I will message you the details on Zulip.

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
```diff
@@ -65,11 +65,9 @@ fn reachable_unvalidated() {
     only_instantiated_by_unvalidated(());
 }
 
-#[ferrocene::prevalidated] //~ NOTE marked
-pub fn instantiation_reachable() { //~ NOTE is validated
+#[ferrocene::prevalidated]
+pub fn instantiation_reachable() {
```
Urhengulas, line 69: Why is only `instantiation_unreachable` throwing warnings now and not both `instantiation_unreachable` and `instantiation_reachable`? Is it some internal de-duplication and by coincidence `instantiation_unreachable` has priority?

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
Urhengulas, line 62: Should the `.into` call within `only_instantiated_by_unvalidated` function not throw a `ERROR calls an unvalidated` now as well? Because with the patch we should consider private and not only public functions anymore, no?

compiler/rustc_monomorphize/src/collector.rs
```diff
@@ -1672,32 +1668,26 @@ impl<'v> RootCollector<'_, 'v> {
...
+                        || self.tcx.is_reachable_non_generic(def_id)
+                        || {
+                            let flags = self.tcx.codegen_fn_attrs(def_id).flags;
+                            flags.intersects(
+                                CodegenFnAttrFlags::RUSTC_STD_INTERNAL_SYMBOL
+                                    | CodegenFnAttrFlags::EXTERNALLY_IMPLEMENTABLE_ITEM,
+                            )
+                        }
+                }
                 // Ferrocene addition
```
jyn514:
```suggestion
                // Ferrocene addition: mark all non-generic functions as annotated, even if they're private.
                // This catches errors sooner when `cargo build`-ing a library.
```

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
jyn514: hm, i believe it's some kind of deduplication, but i'm not sure why ... this has `//@ revisions: dedup no-dedup` at the top, this should only be hidden with the `dedup` revision. i'd have to review the code more to know for sure.

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
jyn514: `only_instantiated_by_unvalidated` is generic, so it is not a root. it's called by `reachable_unvalidated`, but that function isn't annotated with the attribute, so it's also not a root.

Urhengulas: That makes sense!

tests/ui/ferrocene/lint-prevalidated/post-mono.rs
Urhengulas: I don't think this is a blocker, more a curiosity question

jyn514 (approved):
bors merge

jyn514:
you need to mark a few `nofloat` functions as prevalidated:
```
error: validated function possibly calls an unvalidated method
    --> /home/ci/project/library/core/src/fmt/rt.rs:98:29
     |
  98 |         argument_new!(T, x, <T as Debug>::fmt)
     |                             ^^^^^^^^^^^^^^^^^
     |
    ::: /home/ci/project/library/core/src/fmt/nofloat.rs:9:20
     |
   9 |                 fn fmt(&self, _fmt: &mut Formatter<'_>) -> Result {
     |                    --- `nofloat::<impl fmt::Debug for f32>::fmt` is unvalidated
```
also for f16, f64, and f128.

jyn514 (approved):
bors merge

Urhengulas:
> failed to retrieve the public key from AWS KMS

bors retry

PR #2501, opened 2026-07-21
Mirror the sources for a variety of our deps

Xylakant:
* Build from mirrored source
* Provides a makefile to semi-automate the process
* The resulting binary packages now carry the commit that they've been built from in the package name

### DIFF
--- a/ferrocene/ci/mirrors/Makefile
+++ b/ferrocene/ci/mirrors/Makefile
@@ -93,4 +93,9 @@ upload-all: upload-musl-cross upload-python upload-gdb upload-cmake
 
 clean-build:
 	rm -rf build/
+
+clean-output:
+	rm -rf output/
+
+clean-all: clean-build clean-output
 #   docker image rm musl-tarballs
--- a/ferrocene/ci/mirrors/Makefile
+++ b/ferrocene/ci/mirrors/Makefile
@@ -5,10 +5,11 @@ musl-cross-target-version=v0.9.11
 python-version=3.12.3
 gdb-version=12.1
 cmake-version=3.21.1
+commit != git rev-parse HEAD | cut -c 1-8
+host-arch != uname -m
 
 prepare:
 	mkdir output
-	commit != git rev-parse HEAD | cut -c 1-8
 
 .ONESHELL:
 musl-cross: musl.dockerfile prepare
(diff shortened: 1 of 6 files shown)


ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+musl-cross-target-version=v0.9.11
```
fs-rachel, line 4: I generally prefer to use `:=` for assignments rather than `=`. It's equivalent here, but can avoid issues in more complex makefiles

Compare https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html and https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+python-version=3.12.3
+gdb-version=12.1
+cmake-version=3.21.1
+commit != git rev-parse HEAD | cut -c 1-8
```
fs-rachel, line 8: Similarly, I'd prefer `commit := $(shell ...)` instead of `commit != ...`. The issue is that `!=` will expand any variable references it sees in the output, which can lead to confusing issues if it's not intended:

https://www.gnu.org/software/make/manual/html_node/Setting.html

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+	cd hashes; sha256sum *tar.xz > ../hashes.txt
+
+clean-build:
+	rm -rf build/
+
+clean-output:
+	rm -rf output/
+
+clean-all: clean-build clean-output
+#   docker image rm musl-tarballs
```
fs-rachel, line 116: I would add a `.PHONY:` rule at the end of the file listing all of the targets which aren't real files (which is all of them here). That way things won't mysteriously break if someone accidentally creates a file with an overlapping name

https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+
+.ONESHELL:
+musl-cross: musl.dockerfile prepare
+	mkdir -p build/musl-cross
+	cd build/musl-cross
+	aws s3 cp s3://ferrocene-ci-artifacts/manual/musl-cross-make/musl-cross-make-$(musl-cross-target-version)-with-source.tar.xz musl-cross-make.tar.xz
+	tar xf musl-cross-make.tar.xz
+	docker build -t musl-tarballs --target tarballs --build-arg TARGETVERSION=$(musl-cross-target-version) -f ../../musl.dockerfile .
+	{ \
+		set -xe ; \
```
fs-rachel, line 22: Make already prints each subcommand one by one (even in `.ONESHELL` mode) as it executes, and errors out immediately if any subcommand fails. So `set -xe` is redundant.

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+host-arch != uname -m
+
+prepare:
+	mkdir -p output
+
+.ONESHELL:
+musl-cross: musl.dockerfile prepare
+	mkdir -p build/musl-cross
+	cd build/musl-cross
+	aws s3 cp s3://ferrocene-ci-artifacts/manual/musl-cross-make/musl-cross-make-$(musl-cross-target-version)-with-source.tar.xz musl-cross-make.tar.xz
```
fs-rachel, line 18: I prefer to use `${var}` for simple variable substitutions and `$(command ...)` when using Make's built-in functions.

Although both forms are equivalent in Makefiles, this makes things easier to understand for people who aren't used to Makefiles, who might come in expecting the shell-style behaviour where `$(...)` runs a command instead of expanding a variable.

ferrocene/ci/mirrors/Makefile
fs-rachel: Also are the braces here necessary at all?

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/musl/musl-cross-make-aarch64-to-x86_64-$(musl-cross-target-version)-$(commit).tar.xz hashes
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/python/python-$(python-version)-aarch64-$(commit).tar.xz hashes/
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/gdb/gdb-$(gdb-version)-aarch64-$(commit).tar.xz hashes/
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/cmake/cmake-$(cmake-version)-aarch64-$(commit).tar.xz hashes/
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/musl/musl-cross-make-x86_64-to-aarch64-$(musl-cross-target-version)-$(commit).tar.xz hashes
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/musl/musl-cross-make-x86_64-to-x86_64-$(musl-cross-target-version)-$(commit).tar.xz hashes
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/python/python-$(python-version)-x86_64-$(commit).tar.xz hashes/
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/gdb/gdb-$(gdb-version)-x86_64-$(commit).tar.xz hashes/
+	aws s3 cp s3://ferrocene-ci-mirrors/manual/cmake/cmake-$(cmake-version)-x86_64-$(commit).tar.xz hashes/
+	cd hashes; sha256sum *tar.xz > ../hashes.txt
```
fs-rachel, line 107: `sha256sum hashes/*.tar.xz > hashes.txt` would be cleaner

ferrocene/ci/mirrors/Makefile
```diff
@@ -0,0 +1,116 @@
...
+# SPDX-FileCopyrightText: The Ferrocene Developers
+
+musl-cross-target-version=v0.9.11
+python-version=3.12.3
+gdb-version=12.1
+cmake-version=3.21.1
+commit != git rev-parse HEAD | cut -c 1-8
+host-arch != uname -m
+
+prepare:
```
fs-rachel, line 11: Bear in mind that the first target in the Makefile is the default if you run `make` with no arguments. So it's good practice for the first rule to be `default: [whatever you want the default target to be]`.

ferrocene/ci/mirrors/Makefile
Xylakant: That includes the directory in the output which is not intended, so that's deliberate. I can add a comment to explain.

fs-rachel: Ah right, that makes sense. A comment would be helpful

PR #2502, opened 2026-07-21
try building deps on al2

Xylakant:


Xylakant:
these have been integrated into other PRs where suitable

PR #2508, opened 2026-07-24
add more tolerance to fetching signing keys from AWS

dotcarmen:
we're seeing several CI failures due to fetching signing keys from AWS failing on a timeout

this PR
- expands the timeout from 3100ms (the default per `aws-config` crate) to 10s
- changes the retry policy from the default 3x to an adaptive 3x that adds client-side throttling between retries

since the errors don't always happen, we may need to tweak these settings again later if CI continues to fail on this task


Hoverbear (approved):
bors merge

PR #2509, opened 2026-07-24
checkout-submodules.sh: fail if any of the git commands fail

dotcarmen:
sometimes checking out a submodule fails. this should cause the job to fail, but instead the `checkout-submodules` step doesn't fail because we don't check the exit status of `git submodule update`

### DIFF
--- a/ferrocene/ci/scripts/checkout-submodules.sh
+++ b/ferrocene/ci/scripts/checkout-submodules.sh
@@ -14,9 +14,17 @@ current_branch=$(git symbolic-ref --short --quiet HEAD || echo "main")
 # Spawn all `git submodule update`s in parallel and then wait for all of them
 # to complete. This intentionally doesn't use the `-j` flag, as that flag
 # apparently still clones the individual submodules in serial.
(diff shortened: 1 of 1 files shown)


ferrocene/ci/scripts/checkout-submodules.sh
```diff
@@ -14,9 +14,17 @@ current_branch=$(git symbolic-ref --short --quiet HEAD || echo "main")
...
     # Don't break if the default remote has a name other than `origin`.
     # See the comment in bootstrap::config::update_submodule.
     git -c "branch.${current_branch}.remote=origin" submodule update --depth 1 "${submodule}" &
+    pids+=($!)
+done
+
+for pid in "${pids[@]}"; do
+    wait "$pid"
 done
-wait
```
jyn514, line 22:
```
$ bash -c 'help wait'
wait: wait [n]
    Wait for the specified process and report its termination status.  If
    N is not given, all currently active child processes are waited for,
    and the return code is zero.
```

terrifying. fantastic catch.

jyn514 (approved):
bors merge

PR #2512, opened 2026-07-24
run libc-test for QNX targets in CI

japaric:
upstream does not run this test suite for QNX targets so we run it here. this PR contains commits that fix libc failures and will be upstreamed.


japaric:
bors try

japaric:
bors try

ferrocene/library/libc/src/unix/nto/neutrino.rs
```diff
@@ -248,6 +269,30 @@ s_no_extra_traits! {
...
+        pub num_cpu: u16,
+        pub system_private: syspage_entry_info,
+        version: [u16; 2], // inline struct
+        pub hwinfo: syspage_entry_info,
+        pub qtime: syspage_entry_info,
+        pub callout: syspage_entry_info,
+        pub typed_strings: syspage_entry_info,
+        pub strings: syspage_entry_info,
+        pub smp: syspage_entry_info,
+        __reserved: Padding<[u8; 160]>, // anonymous union with architecture dependent structs
```
jonathanpallant: should this be Padding<[u64; 20]> to match the header?

ferrocene/library/libc/src/unix/nto/neutrino.rs
```diff
@@ -452,8 +539,19 @@ pub const _NTO_SIDE_CHANNEL: u32 = 1073741824;
...
-pub const _NTO_TIMEOUT_MASK: u32 = (1 << STATE_MAX) - 1;
-pub const _NTO_TIMEOUT_ACTIVE: u32 = 1 << STATE_MAX;
+const STATE_TIMEOUT_MAX: u8 = 24;
+pub const _NTO_TIMEOUT_MASK: u32 = if cfg!(target_os = "nto") {
+    (1 << STATE_MAX) - 1
+} else {
+    (1 << STATE_TIMEOUT_MAX) - 1
+};
+#[cfg(target_os = "nto")]
+pub const _NTO_TIMEOUT_ACTIVE: u32 = if cfg!(target_os = "nto") {
```
jonathanpallant, line 548: `cfg!` inside `cfg(` with same condition?

ferrocene/library/libc/src/unix/nto/neutrino.rs
japaric: yes, that makes sense

ferrocene/library/libc/src/unix/nto/neutrino.rs
japaric: refactor error. nice catch

jonathanpallant (approved):
LGTM from a libc point of view

japaric:
> https://github.com/ferrocene/ferrocene/pull/2512#discussion_r3665728847

I noticed this should be done on the other `cfg(target_os)` version of the struct as well and have amended the commit. also amend a typo in the semver commit message ("one extra API" but the example mentions more than one item).

src/bootstrap/src/core/build_steps/test.rs
```diff
@@ -3597,6 +3597,92 @@ impl Step for Crate {
...
+            eprintln!(
+                "non-QNX targets are currently not tested as that requires using a \
+toolchain with a more recent libc version"
+            );
+            return;
+        }
+
+        // Prepare sysroot
+        // See [field@compile::Std::force_recompile].
+        builder.ensure(Std::new(build_compiler, build_compiler.host).force_recompile(true));
```
jyn514, line 3678: why does this need force_recompile? that should be very rare.

jyn514 (approved):
LGTM, one question

src/bootstrap/src/core/build_steps/test.rs
japaric: this came from the logic in `Crate` step in this file, which is used to run library tests for crates in the libstd facade, and in that step, `force_recompile` is also unconditionally enabled. 

I can't say I'm sure about the semantics of this flag. with it set to `true`,  calling `./x test ferrocene/library/libc` two times in a row does not cause any of the crates in the libstd facade, including our libc, to be recompiled. modifying `ferrocene/library/libc/libc-test` does not make this `ensure` call recompile anything either. modifying `ferrocene/library/libc/src/lib.rs` does cause libc and the libstd crates that depend on it to be recompiled but that's expected behavior because of the dependency relationship.

should we leave it out? it does not seem to affect the normal Cargo behavior AFAICT. Though, I'll add that I have been testing with `rust.download-rustc=false` (because of `profile = ferrocene-dist`) and with `--stage 1`

jyn514: right ... i think this flag only matters in the presence of download-rustc (where the build stage and test stage might be different). we don't support download-rustc in ferrocene anyway so let's just keep it closer to upstream.

jyn514:
bors merge

jyn514:
bors r-

let's let https://github.com/ferrocene/ferrocene/pull/2517 merge first, it's got a bunch of queued changes and also updates the libc subtree.

fs-rachel:
Should be an easy rebase at least, the only conflict is:

```
// In src/bootstrap/src/core/builder/mod.rs
<<<<<<< HEAD
                test::IntrinsicTest,
||||||| cc57d100734
=======
                // Ferrocene addition
                test::FerroceneCrateLibc,
>>>>>>> ferrocene/ja/libc-test-in-ci
```

and we want to keep both additions

japaric:
> Should be an easy rebase at least, the only conflict is:

there was also a semantic conflict as the `Step` trait was renamed but it has been fixed.

japaric:
bors r=jyn514

PR #2513, opened 2026-07-27
Use 3-way conflict markers for merge conflicts in subtree pulls

fs-rachel:
We already do this for pulls from rust-lang/rust, just not from the other repositories we use as subtrees. Three-way conflict markers make it much easier to identify the conflicting changes when fixing things by hand.


jyn514 (approved):
LGTM, feel free to merge if it helps you

fs-rachel:
Seeing as there's already a pending merge job, might as well roll this in

bors merge

PR #2514, opened 2026-07-27
fix: check for Mergiraf resolutions at an appropriate time

ada4a:
Turns out my fix in https://github.com/ferrocene/ferrocene/pull/2367 was plain wrong...

Previously, if Mergiraf resolved some conflicts, we would stage them, and _then_ run the `status | grep` thing. However, the latter would never match anything: `^ M ` matches files modified in the work tree, and would therefore not see our modified files, which would already be in the staging area.

An alternative fix would be to change the regex to `^M  `, as that one matches precisely the staged modified files; but I figured we could save a tiny bit of work by not `add`ing when there's nothing to `add`.

### DIFF
--- a/ferrocene/tools/pull-upstream/pull.sh
+++ b/ferrocene/tools/pull-upstream/pull.sh
@@ -342,13 +342,13 @@ then
                 mergiraf solve --keep-backup=false "$file" || true
             done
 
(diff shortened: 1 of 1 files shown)


jyn514 (approved):
bors merge

ada4a:
Just a bunch of network failures...

Urhengulas:
bors retry

PR #2516, opened 2026-07-27
Use mergiraf to fix merge conflicts in subtree pulls

fs-rachel:
Mergiraf is currently invoked by pulls from upstream rust-lang/rust, but not for pulls for subtrees. Extract the logic into a separate script so it can be called from both scripts


jyn514 (approved):
this looks good to me. eventually it would be nice to share more between pull.py and pull.sh but this is a good first step 👍 

up to you whether you consider this ready to merge, i'm happy with it.

fs-rachel:
Changes from last time:
1) Rebased onto current main branch, and tidied up commit messages
2) Reference the new fix-merge script with a path relative to the repository root, not relative to the currently running script

fs-rachel:
@jyn514 this is ready for real now, please can you re-review?

jyn514 (approved):
bors merge

PR #2518, opened 2026-07-29
feat: simplify llvm-rebuild parameters

dotcarmen:
after #2490, llvm rebuild parameters can also be reduced into a single `llvm-rebuild` parameter

### DIFF
--- a/ferrocene/ci/scripts/utils/src/utils/docker_images.py
+++ b/ferrocene/ci/scripts/utils/src/utils/docker_images.py
@@ -36,11 +36,11 @@ def calculate_hash() -> str:
     for file in sorted(all_files):
         with open(file, "rb") as f:
             filename = file.encode("utf-8")
-            hash.update(f"{len(filename)}".encode())
+            hash.update(f"{len(filename)}|".encode())
             hash.update(filename)
 
             contents = f.read()
-            hash.update(f"{len(contents)}".encode())
+            hash.update(f"{len(contents)}|".encode())
             hash.update(contents)
 
     return hash.hexdigest()
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -15,7 +15,7 @@ version: 2.1
 #
 # * `docker-images-hash`: the hash for docker image tags.
 # * `docker-repository-url--`: calculates the URL of an ECR repository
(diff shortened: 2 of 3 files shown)


dotcarmen:
once #2490 lands, i'll rebase this pr on latest `main`. until then, the only relevant commit is the latest one in this branch

ferrocene/ci/scripts/utils/src/utils/docker_images.py
```diff
@@ -0,0 +1,46 @@
...
+            all_files += [os.path.join(root, file) for file in files]
+
+    # This is done in two steps to guarantee a stable sorting for the files,
+    # otherwise inconsistencies in the filesystem could result in different
+    # hashes even though the two directories are equal.
+    hash = hashlib.sha256()
+    for file in sorted(all_files):
+        with open(file, "rb") as f:
+            filename = file.encode("utf-8")
+            hash.update(f"{len(filename)}".encode())
```
fs-rachel: The tiniest of tiny nitpicks, but: as the length field is itself variable-length, there's still potential for a collision between (say) a 117-byte filename and an 11-byte filename which starts with '7' (imagine trying to decode the hash input back into the original data, that situation would be ambiguous)

There's a couple of ways to deal with that, but the easiest is to throw in a separator which can't be part of the length field, eg:

```suggestion
            hash.update(f"{len(filename)}|".encode())
```

The length of the filename is then fully pinned down, so we don't necessarily need another `|` after the filename itself

ferrocene/ci/scripts/utils/src/utils/docker_images.py
```diff
@@ -0,0 +1,46 @@
...
+            hash.update(filename)
+
+            contents = f.read()
+            hash.update(f"{len(contents)}".encode())
```
fs-rachel: Same deal here:

```suggestion
            hash.update(f"{len(contents)}|".encode())
```

fs-rachel (approved):

ferrocene/ci/scripts/utils/src/utils/docker_images.py
dotcarmen: ah, that should've been in #2490 😅

dotcarmen:
bors merge

PR #2521, opened 2026-07-31
Pessimistic line coverage

listochkin:
`llvm_profparser` looks for all regions within line and if one of them has been hit it would count the whole line as hit.

We only want to count lines that have *all* their regions covered.

Note that the `hits` map does *not* record a region kind, so technically if there's a region with some kind other than `Code` and it has hits we will still count it.

### DIFF
--- a/ferrocene/tools/blanket/src/main.rs
+++ b/ferrocene/tools/blanket/src/main.rs
@@ -259,10 +259,17 @@ fn get_coverage(
     let mut covered = vec![];
 
     for line in source_lines {
+        let lower_hit_bound = func_coverage
+            .hits
+            .iter()
+            .filter(|(location, _)| location.line_start <= line && location.line_end >= line)
+            .map(|(_, hit_count)| *hit_count)
+            .reduce(usize::min);
         // one more thing to do: within a function, some lines will always be uncovered (e.g. }
         // closing braces). so we do have to trust the coverage tool to report those accurately.
-        let status = match func_coverage.hits_for_line(line) {
+        let status = match lower_hit_bound {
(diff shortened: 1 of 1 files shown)


Hoverbear:
bors try

Hoverbear:
Confirming the following test sniped a statement that was impacted by this.

```rust
// library/coretests/tests/ferrocene/iter.rs
//
// <core::iter::adapters::step_by::StepBy<core::ops::range::Range<u16>> as core::iter::adapters::step_by::StepByImpl<core::ops::range::Range<u16>>>::spec_nth
#[test]
fn test_iter_step_by_spec_try_fold_none() {
    let x = 0_u16..100;
    let iter = x.into_iter();
    let mut step_by = iter.step_by(2);
    assert_eq!(step_by.nth(200), None,);
}
```

Hoverbear:
(Noting from discussion this is waiting on counters before we merge)

jyn514:
what does "waiting on counters" mean?

Hoverbear:
We were originally waiting on counters like 

```
L123 (14/15)   $THE_CODEHERE
```

However this has proven not necessary for landing coverage in 1.97.

@listochkin has a draft PR for doing region coloring in https://github.com/ferrocene/ferrocene/pull/2546. I'm unsure if that will supersede this, or what his intent is.

listochkin:
@Hoverbear this one is "accounting change" - we change the rule about which lines are considered partially / fully covered.

https://github.com/ferrocene/ferrocene/pull/2546 is the UI change. In partially covered lines we highlight regions.

listochkin:
superseded by #2546

PR #2522, opened 2026-07-31
pull-upstream / pull-subtrees: Further cleanup

fs-rachel:
Share more code between the two scripts, preferring to take logic from pull-upstream over pull-subtrees where they differ

Improve warning when merge conflicts happen

Pre-emptively make submodule updates recursive to prevent issues if upstream ever adds recursive submodules


fs-rachel:
Note: To test the updates to these two scripts, I developed them based on an earlier commit, mainly so I knew the libc pull would cause a merge conflict that needed resolution. Then rebased up to the main branch once I knew everything worked.

I've uploaded that branch as https://github.com/ferrocene/ferrocene/tree/rachel/further-subtree-tidying-test if you want to test it yourself

ferrocene/tools/fix-merge/fix-merge.sh
```diff
@@ -33,7 +33,7 @@ fi
 
 # We do a `git submodule update` ahead of time to ensure the wrong
 # submodule commits are not accidentally added.
-git submodule update
+git submodule update --recursive
```
ada4a, line 36: This comment suggests that the upstream tries to remove submodules rather than add new ones: https://github.com/ferrocene/ferrocene/blob/c5869c06baafd6c5fda80d149d55d4fa486a298c/ferrocene/tools/pull-upstream/pull.sh#L144-L145

Though I guess that doesn't apply to external projects like llvm, so this might be wise to do.

ada4a (approved):
I won't claim to be able to review bash scripts.. The mergiraf parts seem okay :+1:

ferrocene/tools/fix-merge/fix-merge.sh
```diff
@@ -5,32 +5,96 @@
...
+        echo "Need 'gsed' installed; try 'brew install gnu-sed'"
+        exit 1
+    fi
+    sed=gsed
+else
+    sed=sed
+fi
+
 if ! mergiraf --version >/dev/null 2>&1; then
     echo "Need 'mergiraf' installed to automatically fix merge conflicts"
```
jyn514: hm, it worries me a bit that this is now a hard error, errors in automation are really hard to find because you have to dig into the logs. could we make this an `automation_warning` instead somehow?

jyn514 (commented):
sorry for the long delay. this looks good to me :) one comment (this was a pre-existing bug, but it happens more often now).

ferrocene/tools/fix-merge/fix-merge.sh
jyn514: "update" means that this is related to an *existing* submodule whose commit is being updated. it's not whether a submodule is added or deleted.

ferrocene/tools/fix-merge/fix-merge.sh
fs-rachel: Done, tested the new version both with and without mergiraf installed

jyn514 (approved):
bors merge

PR #2523, opened 2026-07-31
fixes in qualification docs

japaric:
- in QNX8, the contents of the target overview were incorrectly copy pasted from the qualification scope page
- the qualification scope of the RHIVOS manual indicated that the target was bare metal but it actually has a full standard library

### DIFF
--- a/ferrocene/doc/qnx8-manual/src/targets/index.rst
+++ b/ferrocene/doc/qnx8-manual/src/targets/index.rst
@@ -25,17 +25,17 @@ Qualified QNX8 targets
      - Standard library
      - Notes
 
-   * - :target:`x86_64-unknown-linux-gnu`
-     - :target:`aarch64-unknown-qnx`
-     - ``core``
-     - ``alloc``, ``std``, ``test``
-     - ``proc_macro``
+   * - :ref:`aarch64-unknown-qnx`
+     - ``aarch64-unknown-qnx``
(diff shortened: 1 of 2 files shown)


ferrocene/doc/qnx8-manual/src/targets/index.rst
```diff
@@ -25,17 +25,17 @@ Qualified QNX8 targets
...
-
-   * - :target:`x86_64-unknown-linux-gnu`
-     - :target:`x86_64-pc-qnx`
-     - ``core``
-     - ``alloc``, ``std``, ``test``
-     - ``proc_macro``
+   * - :ref:`aarch64-unknown-qnx`
+     - ``aarch64-unknown-qnx``
+     - Cross-compilation
+     - Full
```
jyn514, line 31: are you sure this is right? i thought aarch64 QNX was bare-metal?

jyn514 (changes requested):

ferrocene/doc/qnx8-manual/src/targets/index.rst
japaric: all the QNX targets, both QNX7.1 and QNX8, have *qualified* libstd (+ proc-macro on the host) support even if only a subset of libcore is *certified* at the moment.

jyn514 (approved):
bors merge

PR #2525, opened 2026-08-04
tidy: check for completeness and correctness of qualification docs

japaric:
This PR adds a new tidy check that checks that docs and other manifest are complete and correct for qualified targets. The central location for target information is `ferrocene/targets.toml`; other places will be cross checked against `targets.toml`.

This new tidy check supersedes the `ferrocene_test_outcomes` check as it's more complete; and also lets us make the `:tested_target_with_std:` in the `render-test-outcomes` Sphinx plugin optional again as it'll check for wrong values for bare metal targets.

See individual commits for details. The '[try]' commit is a further potential simplification that has not been integrated into the tidy check as it still needs to be confirmed with a `bors try`.


japaric:
bors try

japaric:
bors try

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+        const OPT_TARGET: &str = ":target:";
+        const OPT_TESTED_TARGET: &str = ":tested_target_with_std:";
+        const OPT_HOST: &str = ":host:";
+
+        let doc_dir = ferrocene_dir.join("doc");
+        let mut found = BTreeMap::new();
+        for report in QUALIFICATION_REPORTS {
+            let rustc_dir = doc_dir.join(report).join("src/rustc");
+            for entry in fs::read_dir(&rustc_dir).unwrap() {
+                let path = entry.unwrap().path();
```
Urhengulas: For me it is more clear to specify what this path is for
```suggestion
                let test_result_page_path = entry.unwrap().path();
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                        if target.is_some() {
+                            check.error(format!(
+                                "duplicate {OPT_TARGET} option in {}",
+                                path.display()
+                            ));
+                            continue;
+                        }
+                        if filename != tuple {
+                            check.error(format!(
+                                "{OPT_TARGET} does not match filename in {}",
```
Urhengulas:
```suggestion
                                "`{OPT_TARGET}{tuple}` does not match filename in {}",
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+            // the first tuple is the host and the second is the target
+            while let Some(line) = lines.next() {
+                // * - :target:`x86_64-unknown-linux-gnu`
+                let Some(rest) = strip_prefixes(line, &["*", "-", ":target:", "`"]) else {
+                    continue;
+                };
+                let Some(host) = rest.strip_suffix('`') else { continue };
+
+                //   - :target:`aarch64-unknown-none`
+                let Some(line) = lines.next() else { continue };
```
Urhengulas, line 277: I am struggling to understand the effect of this line. Could it also be a `break`?

Iiuc it tries to go to the next line, but continues if we are at the last line already. Because of the continue we are jumping back to up to `while let Some(line) = lines.next()` which will also not have a line anymore, therefore the while-loop is over. No?

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+            // look for this pattern:
+            //     * - :target:`x86_64-unknown-linux-gnu`
+            //       - :target:`aarch64-unknown-none`
+            //       - ``core``
+            //       - [``alloc``|``alloc``, ``std``, ``test``]
+            //       - [|``proc_macro``]
+            //
+            // the first tuple is the host and the second is the target
+            while let Some(line) = lines.next() {
+                // * - :target:`x86_64-unknown-linux-gnu`
```
Urhengulas: I was confused first because I thought you only search for this specific target, but then I realised this is an example for how a host target is formatted.

```suggestion
                //  Example: "* - :target:`x86_64-unknown-linux-gnu`"
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                let Some(rest) = strip_prefixes(line, &["*", "-", ":target:", "`"]) else {
+                    continue;
+                };
+                let Some(host) = rest.strip_suffix('`') else { continue };
+
+                //   - :target:`aarch64-unknown-none`
```
Urhengulas:
```suggestion
                // Example: "  - :target:`aarch64-unknown-none`"
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                let Some(line) = lines.next() else { continue };
+                let Some(rest) = strip_prefixes(line, &["-", ":target:", "`"]) else { continue };
+                let Some(target) = rest.strip_suffix('`') else { continue };
+
+                // - ``core``
```
Urhengulas, line 281:
```suggestion
                // Pattern: "  - ``core``"
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                let Some(line) = lines.next() else { continue };
+                let Some(certified) = strip_prefixes(line, &["-"]) else { continue };
+
+                const CERTIFIED: &str = "``core``";
+                if certified != CERTIFIED {
+                    check.error(format!("{target} is missing certified {CERTIFIED} library"));
+                }
+
+                // - [``alloc``|``alloc``, ``std``, ``test``]
```
Urhengulas, line 290:
```suggestion
                // Pattern: "  - [``alloc``|``alloc``, ``std``, ``test``]"
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                let Some(line) = lines.next() else { continue };
+                let Some(uncertified) = strip_prefixes(line, &["-"]) else { continue };
+
+                // - [|``proc_macro``]
```
Urhengulas, line 294:
```suggestion
                // Pattern: "  - [|``proc_macro``]"
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
+use core::fmt;
```
Urhengulas, line 17: Let's add the description from the commit message here as well

```suggestion
//! # Check for completeness and correctness of qualification docs for qualified targets
//!
//! This uses `ferrocene/targets.toml` as the central location that specifies which
//! targets are qualified and their properties, like whether they are bare metal or
//! not.
//!
//! the tidy check analyzes the following files:
//!
//! - compilation target pages in user manuals
//! - compilation target overviews in user manuals
//! - qualification scopes in safety manuals
//! - test results in qualification reports
//! - `ferrocene/packages.toml` which describes criticalup components
//! - `doc/src/target-names.toml` which maps target tuples to more
//!   user friendly target names

use core::fmt;
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                "found the same target `{tuple}` in different groups at {}",
+                path.display()
+            );
+        }
+
+        HostsAndTargets { hosts: hosts.targets, targets }
+    }
+
+    let in_docs = in_docs(ferrocene_dir);
+    // one way comparisons because Quality Managed and Supported targets are
```
Urhengulas, line 486:
```suggestion
    // Check host group
    // one way comparisons because Quality Managed and Supported targets are
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+    for host in &in_docs.hosts {
+        if in_docs.targets.contains(host) {
+            check.error(format!(
+                "host tuple '{host}' must not appear under \
+[groups.cross-compilation] or [groups.qnx] in ferrocene/packages.toml"
+            ));
+        }
+    }
+
+    // compare one way because Quality Managed and Supported targets are
```
Urhengulas, line 512:
```suggestion
    // Check target groups (`cross-compilation` and `qnx`)
    // compare one way because Quality Managed and Supported targets are
```

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+) where
+    T: PartialEq + fmt::Debug,
+{
+    for target in in_toml.keys() {
+        if !in_docs.contains_key(target) {
+            check.error(format!(
+                "target `{target}` was specified in ferrocene/targets.toml but it's not present in {location}"
+            ))
+        }
+    }
```
Urhengulas:
```suggestion
    compare_one_way(
        &in_toml.keys().cloned().collect(),
        &in_docs.keys().cloned().collect(),
        location,
        check,
    );
```

Urhengulas (changes requested):
Looks great! I struggled with adding all the necessary bits for a new target a few times so I am grateful for this impovement ❤️ 

I left a few suggestions. Except adding the docs to the top of `ferrocene_targets.rs` all are optional so feel free to disregard them if you feel they are unecessary.

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+    check: &mut RunningCheck,
+) {
+    fn parse(ferrocene_dir: &Path) -> BTreeSet<String> {
+        let doc_dir = ferrocene_dir.join("doc");
+        let mut found = BTreeMap::new();
+        for manual in USER_MANUALS {
+            let rustc_dir = doc_dir.join(manual).join("src/targets");
+            extend_with_targets_in_dir(&rustc_dir, &mut found)
+        }
+        found.into_keys().collect()
```
tshepang: we only use keys of BTreeMap, so might as well use a BTreeSet

src/tools/tidy/src/ferrocene_targets.rs
```diff
@@ -0,0 +1,622 @@
...
+                    // stop at start of next section
+                    break;
+                }
+
+                // * - :ref:`x86_64-unknown-linux-gnu`
+                let Some(rest) = strip_prefixes(line, &["*", "-", ":ref:", "`"]) else { continue };
+                let Some(tuple) = rest.strip_suffix('`') else { continue };
+
+                // - (..)
+                _ = lines.next();
```
tshepang: may be worth ensuring that the value here matches that in ":ref:" (previous line)... as a sidenote, the repetition in rst is unfortunate

src/tools/tidy/src/ferrocene_targets.rs
japaric: you are correct if this `next` returns `None` then the iterator is exhausted so going back to `while let` will also observe a `None` (+) and finish the loop. Therefore using a `break` here would have the same effect. OTOH, this `else { continue }` pattern is used consistently so using `break` in some lines would break (pun intended) the homogeneity. I'm fine doing it either way.

(+) `Lines` is a "well-behaved" iterator. it will *not* return a `Some` after returning `None`, which the `Iterator` trait allows in principle.

src/tools/tidy/src/ferrocene_targets.rs
japaric: nice catch. `extend_with_targets_in_dir` used to be called in two places, in a previous iteration of this tidy check, but now that this is the only use it makes sense to use a `BTreeSet`

src/tools/tidy/src/ferrocene_targets.rs
japaric: good idea

japaric:
addressed review comments in last 3 commits and rebased against `main` now that #2523 has been merged

Urhengulas (approved):
bors merge

I see Tshepangs comments being reflected as well.

Urhengulas:
@handlebors merge

Urhengulas:
@handlebors r+

tshepang:
@handlebors retry

dotcarmen:
@handlebors retry

Xylakant:
@handlebors retry

dotcarmen:
@handlebors r-

dotcarmen:
bors try

dotcarmen:
bors ping

dotcarmen:
bors merge

PR #2527, opened 2026-08-06
Rewrite internal procedures section on cross-compiling from Mac

fs-rachel:


ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
-      forwardAgent: true
-      localPort: 50123
-    cpus: 8
-    memory: "8GiB"
-    mountTypesUnsupported: ["9p"]
-    EOF
+    container registry login --username <email> --password-stdin harbor.infra.ferrous-systems.net
 
-.. Note::
+and provide the "CLI secret" for your account (which can be found from the User Profile page)
```
Hoverbear: Could probably coalesce this into the sentence above the command to make it feel a bit less jarring.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
+followed by Ctrl-D
+
+Next, configure and create a container using
```
Hoverbear: In other parts of the document you use `:` at the end before code examples, which I think is inline with the rest of the docs. Here you don't, can I ask why?

```suggestion
Next, configure and create a container:
```

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
+    aws sso login --profile ferrocene-ci --use-device-code
+
+    # Check out ferrocene
+    git clone git@github.com:ferrocene/ferrocene.git
+    cd ferrocene
+    ferrocene/ci/scripts/setup-uv.sh
+    [set up bootstrap.toml]
+    ./x test bootstrap # to ensure everything works
+
+Finally, to run cross-platform tests, the following steps need to be done *per target* inside
```
Hoverbear, line 150: Once https://github.com/ferrocene/ferrocene/issues/2446 lands, I wonder if the right structure is to specify these for teach target in the per target docs below? So we have a consistent:

```
Host Setup
  Linux
  Mac
  Windows
Target Setup
  aarch64
  x86_64
  armv7em
  armv7r
  ...
```

This way if someone wants to test armv7r they can always go to that heading and get instructions (for any OS).

... Anyways no change here for now, just thinking for the future.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
-      localPort: 50123
-    cpus: 8
-    memory: "8GiB"
-    mountTypesUnsupported: ["9p"]
-    EOF
+    container registry login --username <email> --password-stdin harbor.infra.ferrous-systems.net
 
-.. Note::
+and provide the "CLI secret" for your account (which can be found from the User Profile page)
+followed by Ctrl-D
```
Hoverbear: It's worth noting this takes a *hot* minute with no output but will eventually output "Login succeeded".

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
+
+Next, configure and create a container using
+
+.. code-block:: bash
 
-    These commands will create a VM named `ferrocene` and a folder at `~/lima/shared` on the host that is added as a writeable shared mount point to the VM.
-    The entire home directory of the host is also mounted as readonly in the VM at `~/host`.
+    container create -t --cpus 10 --memory 16g --name ferrocene-dev <image URL>
```
Hoverbear:
```suggestion
    container create -t --cpus 10 --memory 16g --name ferrocene-dev harbor.infra.ferrous-systems.net/ferrocene-images/ci:ubuntu-20-main
```

We can straight up put in a tag. :) If not, we should show an example with a dummy hash so we know exactly what it should look like.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
 
-Start the guest:
+Adjust the CPU and memory allocations as appropriate to your device, remembering to leave
+some memory free for the host. The above numbers are selected for an M5 Macbook Air.
```
Hoverbear:
```suggestion
some memory free (at least 4gb) for the host. The above numbers are selected for an M5 Macbook Air.
```

Or whatever is a reasonable amount.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
+
+Now start the container and install some packages we will need:
 
 .. code-block:: bash
 
-    limactl start ferrocene
+    container start ferrocene-dev
+    container exec -it --uid 0 ferrocene-dev /bin/bash
+    # This will open a root shell in the container. In that shell, run:
+    apt update
```
Hoverbear, line 104: I kinda thought these would all be not needed on the CI images?

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,143 @@ Lima or WSL2.)
...
 
-Shell into the guest:
+Push information we'll need for logging into things:
 
 .. code-block:: bash
 
-    limactl shell ferrocene
+    container cp ~/.aws/config ferrocene-dev:/home/ci/.aws/config
+    container cp ~/.ssh/id_ed25519 ferrocene-dev:/home/ci/.ssh/id_ed25519
+    container cp ~/.ssh/id_ed25519.pub ferrocene-dev:/home/ci/.ssh/id_ed25519.pub
```
Hoverbear, line 119: I believe most developers use onepassword as their SSH agent on work machines, which is incompatible with this.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: That sounds like a good plan for the future

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Update+upgrade isn't strictly needed, but the extra packages are needed. Some are for quality-of-life, the others are to be able to build our patched qemu. Have asked Felix if this will still be needed after https://github.com/ferrocene/ferrocene/pull/2446 is merged.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Done

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: For variety ;)

(fixed)

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Added a comment to that effect

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Done

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Done

ferrocene/doc/internal-procedures/src/cross-compiling.rst
Hoverbear: This should work for Macs like mine with 1password:

```
container create -t --cpus 10 --memory 16g --name ferrocene-dev -v "$HOME/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock:/tmp/ssh-auth.sock" \
  -e SSH_AUTH_SOCK="/tmp/ssh-auth.sock"  harbor.infra.ferrous-systems.net/ferrocene-images/ci:ubuntu-24-main
```

If folks have their machine configured correctly (unlike, say, me) this should work generally:

```
container create -t --cpus 10 --memory 16g --name ferrocene-dev -v $SSH_AUTH_SOCK:/tmp/ssh-auth.sock \
  -e SSH_AUTH_SOCK="/tmp/ssh-auth.sock"  harbor.infra.ferrous-systems.net/ferrocene-images/ci:ubuntu-24-main
```

This only works if the user is root though, the user needs to manage permissions on `/tmp/ssh-auth.sock` unfortunately. `777` works, but is not ideal. I think the right strategy would be `chown`'ing the socket...

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,145 @@ Lima or WSL2.)
...
+    # Logging in may take a couple of minutes, but will print "Login succeeded" when it's done
 
-.. Note::
+Next, configure and create a container:
+
+.. code-block:: bash
 
-    These commands will create a VM named `ferrocene` and a folder at `~/lima/shared` on the host that is added as a writeable shared mount point to the VM.
-    The entire home directory of the host is also mounted as readonly in the VM at `~/host`.
+    container create -t --cpus 10 --memory 16g --name ferrocene-dev \
```
Hoverbear, line 87: So if the developer doesn't shut down the container, does it keep holding exclusive lock on all this memory/CPUs? Or is it shared? 

Do we need a like...?

```rst
.. warning::
    You must turn off your container when you are done with it to regain your resources!
```

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,145 @@ Lima or WSL2.)
...
-      loadDotSSHPubKeys: true
-      forwardAgent: true
-      localPort: 50123
-    cpus: 8
-    memory: "8GiB"
-    mountTypesUnsupported: ["9p"]
-    EOF
+    container registry login --username <email> --password-stdin harbor.infra.ferrous-systems.net
+    # Paste the "CLI secret" for your harbor account. This can be found on the User Profile
+    # page on harbor. Then press Ctrl-D
```
Hoverbear, line 80:
```suggestion
    # page on harbor. Ensure there is a newline after, then press Ctrl-D
```

This is necessary it seems because if you paste then immediately ctrl+D it never seems to finish.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,145 @@ Lima or WSL2.)
...
-      Hostname 127.0.0.1
-      Port 50123
-
-You may change `User` to your user name and change `lima-vm` to a name that better describes your vm.
-The vm name is displayed in VS Code when trying to connect via ssh.
+    # Login to AWS
+    aws sso login --profile ferrocene-ci --use-device-code
+
+    # Check out ferrocene
+    git clone git@github.com:ferrocene/ferrocene.git
```
Hoverbear, line 144: Above we encourage the user to `cd qemu-build`, and here we have the user clone ferrocene inside that folder, that might confound them later.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
```diff
@@ -49,108 +50,145 @@ Lima or WSL2.)
...
+    QEMU_LD_PREFIX=/usr/x86_64-linux-gnu /opt/qemu-ferrocene/bin/qemu-x86_64 \
+        build/aarch64-unknown-linux-gnu/stage2-tools-bin/remote-test-server -v \
+        --bind 127.0.0.1:12345
+    # This will wait until you do step 3...
+
+Open a second shell and run the tests themselves:
+
+.. code-block:: bash
+
+    TEST_DEVICE_ADDR=127.0.0.1:12345 ./x test library --target x86_64-unknown-linux-gnu
```
Hoverbear, line 177: I get a few failures, is that expected?

```
failures:
    net::tcp::tests::timeouts
    net::udp::tests::timeouts
    os::unix::net::tests::timeouts
    process::tests::smoke_failure
    process::tests::test_process_output_fail_to_start
    sys::process::unix::common::tests::test_process_group_posix_spawn
    sys::process::unix::common::tests::test_setsid_posix_spawn

```

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Does this work for you inside the container, with the setup that you have?

```
    sudo chown ci:ci /tmp/ssh-auth.sock
    sudo chmod 0600 /tmp/ssh-auth.sock
```

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: It looks like both are shared properly. Some quick testing allocating a giant array in Python suggests that the VM only claims memory when it needs it, then holds onto it until the host is under memory pressure, at which point it releases any memory it doesn't need any more.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Oh yeah, I forgot to note that. I also see a handful of test failures: the above ones for `std`, and some each in `ui`, `run-make`, and `compiler`. It looks like they're largely due to the qemu environment being different to what the tests expect (eg, calling `cc` expecting it to target x86_64 and being confused when it targets aarch64). Will have to poke at this some more to figure out how to set up the right environment.

fs-rachel: Thinking about it more, this is likely why Felix's upcoming changes to the CI pipeline builds a custom Docker image per target per run, which is going to be hard to replicate well :\

fs-rachel: Testing on a different axis: `thumbv7em-ferrocene.facade-eabihf` can't build with profiler=true for unclear reasons; disabling that, `TEST_DEVICE_ADDR=[..] ./x test library/core library/alloc library/test --target thumbv7em-ferrocene.facade-eabihf` has one test failure that I'm going to look into

(note that this target doesn't support `std`, so we would *expect* `./x test library` to fail in the `std` tests, and don't run those in CI)

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: I've changed this step to use Ubuntu's `unminimize` script which they include in all their minimal images. That runs apt update + apt upgrade, but also installs a bunch of things which we don't need in CI but which are nice to have in a dev environment

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Fixed by adding explicit `cd ~` steps at the start of the qemu and ferrocene setup instructions

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Done

ferrocene/doc/internal-procedures/src/cross-compiling.rst
fs-rachel: Update: I reran the tests using a newer version of ferrocene and with various fixes to the instructions. There are still some issues, but fewer than before:

On the host (aarch64-unknown-linux-gnu), `./x test --stage 2 bootstrap ui compiler library` work, `run-make` has one failure on a test introduced by https://github.com/rust-lang/rust/pull/118095 which seems to be a linker version issue.

Targetting x86_64-unknown-linux-gnu, `./x test library/core library/alloc` works but `./x test library/std` doesn't

Targetting thumbv7em-ferrocene.facade-eabihf, `./x test library/core library/alloc` works; the test failure I saw last time turned out to be because CI sets RUSTDOCFLAGS="--cfg=ferrocene_facade_secretsauce -Z unstable-options --test-args '--exclude-should-panic'" and I wasn't setting that previously. So I've added that to the instructions.

jyn514:
👋 just a note that for documentation in particular i think it's better to have something than nothing, we can always make fixes in follow-ups.

ferrocene/doc/internal-procedures/src/cross-compiling.rst
Hoverbear: Sure does!

Hoverbear (approved):
bors merge

PR #2528, opened 2026-08-11
fix llvm rebuild

dotcarmen:
since #2518 landed, we've been rebuilding LLVM artifacts more often.

The problem is due to not checking for the existence of only the LLVM targets that will be used in the current workflow. assuming LLVM needs to be built for a given git hash, but we're only testing linux targets:
- `llvm-rebuild` will be rightfully `true` on the first run. LLVM needs to be rebuilt for the linux targets
- on the second run, `llvm-rebuild` will be erroneously `false`, since `calculate_parameters.py` checked for the existence of windows and darwin LLVM as well, even though the linux LLVM (the only one we need) is already built

subsequent runs will have the same results as the second.

this PR updates `calculate_parameters.py` to accept `FULL_BUILD_*` environment variables that are used to determine whether to check for the existence of the corresponding llvm artifact.

this PR also sneaks in a few extra changes that should speed up the `calculate_parameters.py` script a slight bit:
- `llvm_cache.get_llvm_cache_hash()` is now memoized, so the script no longer has to hash the contents of all the files used to calculate the llvm cache hash
- `llvm_cache.get_llvm_cache_hash()` also now avoids re-hashing the `ferrocene/ci/docker-images` directory
- `llvm_cache.get_llvm_cache_hash()` also now follows the same pattern for opening/closing files as from `docker_images.py`
- `setup-awscli.sh` had a syntax error, but none of the invocations resulted in a process failure. the syntax failure has been fixed, and the script was changed to bash (over posix sh) so `set -o pipefail` could be added

additionally, `calculate_llvm_rebuild` now informs via stderr which llvm artifacts are missing, for better debugging in the future.

### DIFF
--- a/ferrocene/ci/docker-images/common/setup-awscli.sh
+++ b/ferrocene/ci/docker-images/common/setup-awscli.sh
@@ -1,10 +1,11 @@
-#!/bin/sh
+#!/bin/bash
 
 # SPDX-License-Identifier: MIT OR Apache-2.0
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
 rm -rf /tmp/awscli /tmp/awscli.zip
 set -xe
+set -o pipefail
 
 mkdir /tmp/awscli && cd /tmp/awscli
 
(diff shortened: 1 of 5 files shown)


fs-rachel (approved):
Looks good to me!

ferrocene/ci/docker-images/common/setup-awscli.sh
```diff
@@ -1,10 +1,11 @@
-#!/bin/sh
+#!/bin/bash
```
dotcarmen, line 1: needed for `set -o pipefail`

ferrocene/ci/docker-images/common/setup-awscli.sh
```diff
@@ -68,6 +69,6 @@ unzip -q -d /tmp/awscli awscliv2.zip
 rm -rf /tmp/awscli awscliv2.zip awscliv2.sig
 
 # some of our environments record all downloaded packages here
-if [ test -x "/ferrocene/packages" ]; then
+if [ -x "/ferrocene/packages" ]; then
```
dotcarmen, line 72: this was a syntax error causing the script to fail

Xylakant:
bors merge

PR #2534, opened 2026-08-12
Build all of our required dependencies from mirrored sources

Xylakant:
* openssl
* zstd
* gcc and clang
* python
* gdb
* musl cross compile
* cross compile toolchain built with crosstool-ng for s390x aarch64-linux, powerpc64le and riscv64

The hashes are not updated, thus no changes to the dockerfile

Newly built dependencies are named with the git commit they've been built with to prevent overwriting previous versions.


jyn514:
@Xylakant is this ready for review?

PR #2537, opened 2026-08-13
fix llvm cache again

dotcarmen:
i admittedly didn't test enough the detection of `FULL_BUILD_*` environment variables in `calculate_parameters.py` from #2528 

~~apparently, when you set an environment variable to the boolean `true` value in yaml, circleci sets the environment variable to `1` (and `false` becomes `0`).~~ (update: the env var can be any of `"true", "false", "1", "0"`) i'd only been checking for the string `true`, which means that the script was only ever checking for linux artifacts, never windows or darwin

this PR changes the logic that checks the `FULL_BUILD_*` environment variables in `calculate_parameters.py` to check the variables for well-known truthy values instead (to be extra absolutely completely safe)

tests:
1. needs rebuild, only builds linux
   - workflow: https://app.circleci.com/pipelines/github/ferrocene/ferrocene/20080
   - commit: https://github.com/ferrocene/ferrocene/commit/b49487c99472e52e5585ca6b2ad9c695037f6d5f
2. linux is built, macos+windows need rebuilding
   - workflow: https://app.circleci.com/pipelines/github/ferrocene/ferrocene/20086
   - commit: https://github.com/ferrocene/ferrocene/commit/8f8c9b9cf6beef2e5a7a3aaa09f0373327687bd7
3. all are built, rebuild not needed
   - workflow: https://app.circleci.com/pipelines/github/ferrocene/ferrocene/20099
   - commit: https://github.com/ferrocene/ferrocene/commit/a8e8a1fa8adc258bb97c856fb76a300f7cc98f40 (but it's identical to https://github.com/ferrocene/ferrocene/commit/8f8c9b9cf6beef2e5a7a3aaa09f0373327687bd7)

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -28,6 +28,9 @@ parameters:
   llvm-rebuild:
     type: boolean
     default: false
+  llvm-hash:
+    type: string
+    default: ""
   targets--x86_64-unknown-linux-gnu--build:
     type: string
     default: ""
@@ -139,7 +142,7 @@ jobs:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
       FERROCENE_TARGETS: << pipeline.parameters.targets--x86_64-unknown-linux-gnu--build >>
       SCRIPT: |
-        ferrocene/ci/scripts/llvm_cache.py download
+        ferrocene/ci/scripts/llvm_cache.py --override-hash << pipeline.parameters.llvm-hash >> download
         ./x.py --stage 2 build cargo library src/tools/rustdoc symbol-report
         ./x.py --stage 1 build compiletest coverage-dump src/tools/rustdoc src/tools/run-make-support
     steps: [ferrocene-job-build]
@@ -279,7 +282,7 @@ jobs:
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
       SCRIPT: |
-        ferrocene/ci/scripts/llvm_cache.py download
+        ferrocene/ci/scripts/llvm_cache.py --override-hash << pipeline.parameters.llvm-hash >> download
         ./x.py test --stage=1 --coverage=library library/core library/alloc --tests
     steps:
       - ferrocene/aws_oidc_auth
@@ -295,7 +298,7 @@ jobs:
     resource_class: ferrocene/k8s-amd64-large
     environment:
       FERROCENE_HOST: x86_64-unknown-linux-gnu
-      SCRIPT: ferrocene/ci/scripts/llvm_cache.py prepare
+      SCRIPT: ferrocene/ci/scripts/llvm_cache.py --override-hash << pipeline.parameters.llvm-hash >> prepare
       # This tries to strike a balance between the highest concurrency possible
       # and not running out of memory during a build. If building LLVM fails
       # due to the OOM killer killing the build, decrease this number.
@@ -414,7 +417,7 @@ jobs:
     resource_class: ferrocene/k8s-arm64-large
     environment:
       FERROCENE_HOST: aarch64-unknown-linux-gnu
-      SCRIPT: ferrocene/ci/scripts/llvm_cache.py prepare
+      SCRIPT: ferrocene/ci/scripts/llvm_cache.py --override-hash << pipeline.parameters.llvm-hash >> prepare
       # This tries to strike a balance between the highest concurrency possible
       # and not running out of memory during a build. If building LLVM fails
       # due to the OOM killer killing the build, decrease this number.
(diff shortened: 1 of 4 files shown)


.circleci/workflows.yml
```diff
@@ -917,6 +920,44 @@ jobs:
 workflows:
   version: 2
 
+  carmen:
```
Xylakant: This looks like leftover test setup

.circleci/config.yml
```diff
@@ -64,4 +64,5 @@ workflows:
         - equal: [<< pipeline.git.branch >>, "staging"]
         - equal: [<< pipeline.git.branch >>, "trying"]
         - equal: [<< pipeline.trigger_source >>, "api"]
+        - equal: [<< pipeline.git.branch >>, carmen/fix-llvm-cache-again]
```
Xylakant: This looks like leftover test harness

Xylakant (changes requested):

ferrocene/ci/scripts/calculate_parameters.py
```diff
@@ -123,10 +123,38 @@
...
+except KeyError:
+    FULL_BUILD_X86_64_WINDOWS_MSVC = "TRUE"
+
+try:
+    FULL_BUILD_AARCH64_DARWIN = os.environ["FULL_BUILD_AARCH64_DARWIN"]
+    print(
+        f"-> FULL_BUILD_AARCH64_DARWIN={FULL_BUILD_AARCH64_DARWIN}",
+        file=sys.stderr,
+    )
+except KeyError:
```
Hoverbear: Hm... Are these values populated by the lines 120-124? If not it seems like these are ~effectively consts and could go above?

Also, it seems like if the `os.environ` lookup above fails, and falls through to here, there is no indication of which branch was taken here other than the lack of the `-> FULL_BUILD_AARCH64_DARWIN` log. I worry about this starting to fail and us not noticing the absence of a log. Could we do like...

```
-> Defaulting FULL_BUILD_AARCH64_DARWIN=true
```

ferrocene/ci/scripts/calculate_parameters.py
```diff
@@ -123,10 +123,38 @@
...
+    FULL_BUILD_AARCH64_DARWIN = "TRUE"
+
+TRUTHY = ["1", "TRUE", "True", "true", "YES", "Yes", "yes", "Y", "y"]
```
Hoverbear: While this is good defensive coding, it may be worth revisiting the idea of using a yaml boolean:

> apparently, when you set an environment variable to the **boolean true value in yaml**, circleci sets the environment variable to 1 (and false becomes 0). i'd only been checking for the string true, which means that the script was only ever checking for linux artifacts, never windows or darwin

This tends to survive various de/ser woodchippers better:

```yaml
thing: "true"
```

dotcarmen:
@handlebors ping

dotcarmen:
testing to make sure handlebors works; i won't let this merge

@handlebors r+

dotcarmen:
@handlebors try

dotcarmen:
@handlebors try cancel

@handlebors r+

dotcarmen:
@handlebors cancel

dotcarmen:
@handlebors cancel

dotcarmen:
@handlebors r-

dotcarmen:
@handlebors try

dotcarmen:
@handlebors try cancel

ferrocene/ci/scripts/calculate_parameters.py
dotcarmen: i can do this, but it would involve changing all of our conditions referencing the boolean params from simple boolean conditions to comparing against the string `"true"` which is more error-prone

ferrocene/ci/scripts/calculate_parameters.py
```diff
@@ -123,29 +123,20 @@
...
 }
+full_build: dict[str, bool] = {}
+for target, envname in FULL_BUILD_TARGET_ENVS:
+    try:
+        val = os.environ[envname]
+        print(f"-> {envname}={val}", file = sys.stderr)
+    except KeyError:
+        print(f"-> {envname} unset, defaulting to '1'", file = sys.stderr)
+        val = "1"
+    full_build[target] = val in ("1", "TRUE", "True", "true", "YES", "Yes", "yes", "Y", "y")
```
dotcarmen: in another test using circleci, i was able to have the env var be set to `true` or `false`. it's just inconsistent 🙃

imo still better than using string booleans in the yaml, since:
1. CircleCI's UI works better for triggering the pipeline with boolean values
2. CircleCI's conditional logic supports booleans nicely. doing string checks is a lot more work + error-prone
3. in python, it's really simple to do string checks against a bunch of possible strings

dotcarmen: i guess we could use an `enum` type parameter, which circleci's ui might like

Hoverbear (approved):

jyn514:
What's the state of this PR? Is it waiting on Carmen? on review?

dotcarmen:
i've been waiting for another sign off by @Xylakant - it's probably fine to merge though.

dotcarmen:
bors merge

dotcarmen:
ugh, that worked when i tried locally

dotcarmen:
bors try

dotcarmen:
now that the llvm artifacts/docker images are built, the merge shouldn't rebuild

bors merge

PR #2542, opened 2026-08-18
remove ec2_runners config from rust-bors.toml

dotcarmen:
this isnt currently handled by our handlebors instance, and in fact breaks it

it's config that only applies to rust-lang/rust anyways...


jyn514 (approved):
long-term our plan is to not use `rust-bors.toml` at all and use `handlebors.toml` instead. this seems ok as a short-term fix.

dotcarmen:
@handlebors ping

PR #2545, opened 2026-08-19
[869d1v4mb] Add `ferrocene::requires_validation` attribute

Urhengulas:


compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -2,14 +2,113 @@
...
+        .emit();
+}
+
+pub(super) fn error_requires_validation_not_a_trait_fn(
+    tcx: TyCtxt<'_>,
+    def_id: LocalDefId,
+    attr_span: Span,
+) {
+    let descr = tcx.def_descr(def_id.to_def_id());
+    let article = descr.starts_with(['a', 'e', 'i', 'o', 'u']).then_some("an").unwrap_or("a");
```
jyn514: nit: this can use [`tcx.article_and_description`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.article_and_description)

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
```diff
@@ -0,0 +1,52 @@
...
+    //~^ ERROR `#[ferrocene::requires_validation]` on a trait method with a default body requires `#[ferrocene::prevalidated]` as well
+    //~| HELP add `#[ferrocene::prevalidated]` to validate the default body
+    fn requires_validation_yes_body(&self) {}
+    //~^ NOTE this default body is not validated
+
+    #[ferrocene::requires_validation]
+    //~^ ERROR `#[ferrocene::requires_validation]` cannot be applied to an associated constant
+    //~| NOTE only trait methods can be marked
+    //~| HELP use `#[ferrocene::prevalidated]`
+    const REQUIRES_VALIDATION_ASSOC_CONSTANT: u8;
```
jyn514: I think this should also allow `requires_validation` as an attribute? correct me if i'm wrong, but my understanding is:
- constants can always be used unless they contain a function pointer.
- constants that contain a function pointer must be annotated with `prevalidated`.
- constants that are annotated with `prevalidated` are checked at definition time to make sure that any function they use is validated.

if all those things are true, i think we should also allow `requires_validation` on constants. consider this example:
```rust
#[ferrocene::prevalidated]
fn uses_const<T: Trait>(x: T) {
  println!("{}", <T as Trait>::REQUIRES_VALIDATION_ASSOC_CONSTANT);
}
```
in order to check this pre-mono, we need to know whether `<T as Trait>::REQUIRES_VALIDATION_ASSOC_CONSTANT` is validated or not.

my suggestion is to do this:
1. if a constant contains a function pointer (see `ferrocene::thir::contains_unknown_fn`), allow `requires_validation`.
2. otherwise, error.

does that seem reasonable?

compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -2,14 +2,113 @@
...
+use rustc_middle::middle::codegen_fn_attrs::ferrocene::has_requires_validation_attribute;
+use rustc_middle::ty::TyCtxt;
 use rustc_span::{STDLIB_STABLE_CRATES, Span};
 use tracing::debug;
 
 use crate::ferrocene::post_mono::InstantiationSite;
 use crate::ferrocene::{LintState, UNVALIDATED, UnvalidatedImplCause, Use, UseKind};
 
+fn ident_span(tcx: TyCtxt<'_>, def_id: DefId) -> Span {
+    tcx.def_ident_span(def_id).unwrap_or_else(|| tcx.def_span(def_id))
```
jyn514, line 16: nit: i would prefer if this took `LocalDefId`, since we only ever use it diagnostics for the local crate.

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -203,18 +205,111 @@ use rustc_middle::{
...
+/// this. That is okay because `check_attribute_placement` guarantees that a
+/// trait method with a default body that is marked with `requires_validation`
+/// is also marked `prevalidated`.
+fn check_impl_of_requires_validation(tcx: TyCtxt<'_>, impl_item: LocalDefId) {
+    // Get the id of the trait method definition being implemented by `impl_item`,
+    // if `impl_item` is an associated function with a body.
+    let Some(trait_method_definition) = matches!(tcx.def_kind(impl_item), DefKind::AssocFn)
+        .then(|| tcx.associated_item(impl_item.to_def_id()))
+        // Inherited trait methods will be filtered out by `trait_item_def_id`
+        .and_then(|assoc| assoc.trait_item_def_id())
```
jyn514: nit:
```suggestion
        // Inherited trait methods will be filtered out by `trait_item_def_id`
        .then(|| tcx.trait_item_of(impl_item.to_def_id()))
```

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -203,18 +205,111 @@ use rustc_middle::{
...
+    else {
+        return;
+    };
+
+    if has_requires_validation_attribute(tcx, trait_method_definition).is_none() {
+        // If the trait method definition has no `requires_validation`
+        // attribute, there is nothing to check.
```
jyn514, line 397: nit:
```suggestion
        // attribute, there is nothing to check.
        return;
```

jyn514 (commented):
this looks awesome <3 one comment about associated consts, and a bunch of nits you can choose to ignore.

have you tested this on any traits in core yet?

compiler/rustc_lint/src/ferrocene/diagnostics.rs
Urhengulas: Correct me if I am wrong, but I think there is exactly one place where we use it for a item which is not from the local crate. The trait method definition may not be in the current crate: https://github.com/ferrocene/ferrocene/blob/6c06c2fe0735ef9cf2be1de8e59dd47b16ed0841/compiler/rustc_lint/src/ferrocene/diagnostics.rs#L94

If you prefer I can change this one case to calling `def_ident_span` manually and then change `ident_span` to take `LocalDefId`.

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
Urhengulas: It does seem reasonable. I totally forgot about the function pointer case.

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: https://github.com/ferrocene/ferrocene/pull/2545/commits/ce31b18ef7d483f5ae37564eb72d24129c77c985

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: https://github.com/ferrocene/ferrocene/pull/2545/commits/ce31b18ef7d483f5ae37564eb72d24129c77c985

compiler/rustc_lint/src/ferrocene/diagnostics.rs
Urhengulas: https://github.com/ferrocene/ferrocene/pull/2545/commits/ce31b18ef7d483f5ae37564eb72d24129c77c985

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
Urhengulas: Work in progress is here: https://github.com/ferrocene/ferrocene/pull/2545/commits/364976a1a40d8d47c3357b08fc95fb8c6ba351ab

tests/ui/ferrocene/lint-prevalidated/requires-validation.rs
```diff
@@ -0,0 +1,67 @@
...
+#![deny(ferrocene::unvalidated)]
+//~^ NOTE lint level
+
+trait Trait {
+    #[ferrocene::requires_validation]
+    //~^ NOTE required to be validated here
+    //~^^ NOTE required to be validated here
+    fn requires_validation_no_default(&self) -> u8;
+    //~^ NOTE all implementations of
+    //~^^ NOTE all implementations of
```
Urhengulas, line 14: The notes have to be duplicated here for the test to pass. I have to investigate why and if I can get rid of it.

jyn514: that's because the UI test suite passes `-Z deduplicate-diagnostics=no`. i wouldn't worry about it, it's fine as-is.

compiler/rustc_lint/src/ferrocene/diagnostics.rs
jyn514: oh, you're right. this is fine then.

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -242,25 +242,33 @@ fn check_attribute_placement(
...
-                TraitFn::Provided(_) => ItemRequiringValidation::MethodYesBody,
+                TraitFn::Required(_) => ItemRequiringValidation::Method { default_body: false },
+                TraitFn::Provided(_) => ItemRequiringValidation::Method { default_body: true },
             },
-            // TODO: handle this case
-            TraitItemKind::Const(_ty, _value) => ItemRequiringValidation::AssocConstFnPtr,
+            // Associated constant
+            TraitItemKind::Const(..) => {
+                let const_ty = tcx.type_of(def_id.to_def_id()).skip_binder();
+                match thir::contains_unknown_fn(const_ty) {
```
jyn514: this doesn't look right.
```suggestion
            TraitItemKind::Const(const_ty, _value) => {
                match thir::contains_unknown_fn(const_ty) {
```

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -242,25 +242,33 @@ fn check_attribute_placement(
...
+                    Some(_) => ItemRequiringValidation::ConstFnPtr,
+                    None => ItemRequiringValidation::None,
+                }
+            }
+            // Associated type (unstable)
             TraitItemKind::Type(..) => ItemRequiringValidation::None,
```
jyn514: not super important because it's still unstable, but it's good to be exhaustive for soundness-critical stuff like this.
```suggestion
            TraitItemKind::Type(_bounds, assoc_ty) if thir::contains_unknown_fn(assoc_ty) => ItemRequiringValidation::ConstFnPtr,            
            TraitItemKind::Type(..) => ItemRequiringValidation::None,
```

jyn514 (commented):
(i know this is still a WIP)

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: Unfortunately that does not work:

```text
error[E0308]: mismatched types
    --> compiler/rustc_lint/src/ferrocene/mod.rs:262:49
     |
 262 |                 match thir::contains_unknown_fn(**const_ty) {
     |                       ------------------------- ^^^^^^^^^^ expected `rustc_middle::ty::Ty<'_>`, found `rustc_hir::Ty<'_>`
     |                       |
     |                       arguments to this function are incorrect
     |
     = note: `rustc_hir::Ty<'_>` and `rustc_middle::ty::Ty<'_>` have similar names, but are actually distinct types
```

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: Same here:

```text
error[E0308]: mismatched types
    --> compiler/rustc_lint/src/ferrocene/mod.rs:267:46
     |
 267 |                 if thir::contains_unknown_fn(**assoc_ty).is_some() =>
     |                    ------------------------- ^^^^^^^^^^ expected `rustc_middle::ty::Ty<'_>`, found `rustc_hir::Ty<'_>`
     |                    |
     |                    arguments to this function are incorrect
```

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -266,7 +266,14 @@ fn check_attribute_placement(
                 }
             }
             // Associated type (unstable)
-            TraitItemKind::Type(..) => ItemRequiringValidation::None,
+            TraitItemKind::Type(_bounds, Some(_assoc_ty)) => {
```
jyn514, line 269:
```suggestion
            TraitItemKind::Type(_bounds, Some(_hir_ty)) => {
```

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -266,7 +266,14 @@ fn check_attribute_placement(
...
+                let assoc_ty = tcx.type_of(def_id.to_def_id()).skip_binder();
+                match thir::contains_unknown_fn(assoc_ty) {
+                    Some(_) => ItemRequiringValidation::ConstFnPtr,
```
jyn514, line 272: could you add a test for this case, please?

compiler/rustc_lint/src/ferrocene/mod.rs
jyn514, line 293: i'm not sure how much this giant match is helping you ... you have a bunch of cases that are marked `// ok`, and a couple cases that say "both these cases should emit an error simultaneously". i think it might be simpler to check the bad cases individually rather than trying to use an exhaustive match?

Urhengulas: You are probably right. I generally like that the match ensures that we consider all possible cases. I will see if it is possible to simplify the match and if not switch back to just checking the bad cases.

compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -4,13 +4,169 @@
...
+
+pub(super) fn error_const_prevalidated_no_requires(
+    tcx: TyCtxt<'_>,
+    def_id: LocalDefId,
+    attr_span: Span,
+) {
+    tcx.dcx()
+        .struct_span_err(
+            attr_span,
+            format!("fn ptr const marked {ATTR_VAL} must be marked {ATTR_REQ} as well"),
```
jyn514, line 78: i don't see why this is true? it should be fine for that particular impl to be validated without requiring other impls to be validated. the difference between them will still be caught by the post-mono check.

compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -4,13 +4,169 @@
...
+        )
+        .with_help(format!("add {ATTR_REQ} to require all consts to be validated"))
+        .emit();
+}
+
+pub(super) fn error_const_no_value(tcx: TyCtxt<'_>, def_id: LocalDefId, attr_span: Span) {
+    tcx.dcx()
+        .struct_span_err(
+            attr_span,
+            format!("fn ptr const without a value cannot be marked {ATTR_VAL}"),
```
jyn514, line 92: this looks the same as error_prevalidated_without_body ?

compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -4,13 +4,169 @@
...
+            ident_span(tcx, def_id.to_def_id()),
+            format!("this const is marked {ATTR_VAL}, but is no function pointer"),
+        )
+        .with_help(format!("remove {ATTR_VAL}"))
+        .emit();
+}
+
+pub(super) fn error_prevalidated_wrong_item(tcx: TyCtxt<'_>, def_id: LocalDefId, attr_span: Span) {
+    tcx.dcx()
+        .struct_span_err(attr_span, "{ATTR_VAL} can only be applied to trait items")
```
jyn514, line 123: i don't think this is true? did you mean to write ATTR_REQ?

compiler/rustc_lint/src/ferrocene/diagnostics.rs
```diff
@@ -4,13 +4,169 @@
...
+        .with_span_label(
+            ident_span(tcx, def_id.to_def_id()),
+            format!("this const is marked {ATTR_VAL}, but has is no function pointer"),
```
jyn514, line 126: this looks like a duplicate of error_const_no_fn_ptr. perhaps both should be combined into a single `error_prevalidated_wrong_item`?

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -203,18 +205,186 @@ use rustc_middle::{
...
+        check_attribute_placement(cx.tcx, item.owner_id.def_id, None);
+        check_impl_of_requires_validation(cx.tcx, item.owner_id.def_id);
         LintThir::check_item(cx.tcx, item.owner_id, item.owner_id.def_id);
     }
 }
 
+/// Reject `#[ferrocene::*]` attributes in positions where they have no defined
+/// semantics, or are invalid.
+///
+/// `trait_item` is `Some` only if `def_id` is a trait item.
```
jyn514, line 230:
```suggestion
/// `trait_item` is `Some` only if `def_id` is an associated trait item.
```

compiler/rustc_middle/src/middle/codegen_fn_attrs/ferrocene.rs
```diff
@@ -160,8 +161,9 @@ pub fn item_is_validated(tcx: TyCtxt<'_>, def_id: DefId) -> ValidatedStatus {
     }
 }
 
-/// Check if this item or any of its parents are validated.
-fn any_parent_is_validated(tcx: TyCtxt<'_>, item: DefId) -> Option<ValidatedStatus> {
+/// Check if this item, or any of its parents, is marked with
+/// `#[ferrocene::prevaidated]`.
+pub fn any_parent_is_validated(tcx: TyCtxt<'_>, item: DefId) -> Option<ValidatedStatus> {
```
jyn514, line 166: this should remain private, it's only a helper and can't be trusted by itself. i'd suggest instead adding a `ValidatedStatus::has_annotation` function.

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -203,18 +205,186 @@ use rustc_middle::{
...
+fn check_attribute_placement(
+    tcx: TyCtxt<'_>,
+    def_id: LocalDefId,
+    trait_item: Option<&TraitItem<'_>>,
+) {
+    let prevalidated = match any_parent_is_validated(tcx, def_id.to_def_id()) {
+        // FIXME: consider if validation is inherited
+        Some(ValidatedStatus::Validated { annotation, inherited: false }) => annotation,
+        _ => None,
+    };
```
jyn514, line 247:
```suggestion
    let marked_validated = item_is_validated(tcx, def_id).has_annotation();
```

compiler/rustc_lint/src/ferrocene/mod.rs
```diff
@@ -203,18 +205,186 @@ use rustc_middle::{
...
+    let requires_validation = has_requires_validation_attribute(tcx, def_id.to_def_id());
+
+    // Return early if it is not a trait item
+    let Some(trait_item) = trait_item else {
+        if let Some(span) = requires_validation {
+            diagnostics::error_requires_validation_wrong_item(tcx, def_id, span)
+        }
+        return;
```
jyn514, line 255: this doesn't look right? why do you assume non-trait items are always valid?

in general i think this is happening at the wrong time. there is existing code to check all this in `LintState::new`. i would expect this to be a modification there rather than completely new code. right now you're rechecking the same items over and over again, once per mono root.

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
```diff
@@ -0,0 +1,66 @@
...
+    //~^ ERROR `#[ferrocene::requires_validation]` on a trait method with a default body requires `#[ferrocene::prevalidated]` as well
+    //~| HELP add `#[ferrocene::prevalidated]` to validate the default body
+    fn requires_validation_yes_body(&self) {}
+    //~^ NOTE this default body is not validated
+
+    #[ferrocene::requires_validation]
+    //~^ ERROR `#[ferrocene::requires_validation]` cannot be applied to an associated constant
+    //~| NOTE only trait methods or associated constants
+    //~| HELP use `#[ferrocene::prevalidated]`
+    const ASSOCIATED_CONST_NO_FN_PTR: u8;
```
jyn514, line 32: hm, this help doesn't seem particularly helpful here.

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
```diff
@@ -0,0 +1,66 @@
...
+    //~^ NOTE not a trait method
+
+    #[ferrocene::requires_validation]
+    const ASSOCIATED_CONST_YES_FN_PTR: fn() -> u8 = || 0;
```
jyn514, line 36: shouldn't this error that it's missing `prevalidated`?

tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
```diff
@@ -0,0 +1,66 @@
...
+    //~| NOTE only trait methods or associated constants containing a fn pointer can be marked
+    //~| HELP use `#[ferrocene::prevalidated]`
+    fn inherent(&self) {}
+    //~^ NOTE not a trait method
+
+    #[ferrocene::requires_validation]
+    //~^ ERROR: `#[ferrocene::requires_validation]` cannot be applied to an associated constant
+    //~| NOTE: only trait methods or associated constants containing a fn pointer can be marked
+    //~| HELP: use `#[ferrocene::prevalidated]`
+    const REQUIRES_VALIDATION_ASSOC_CONSTANT3: u8 = 0;
```
jyn514, line 64: i think this would be a better test if it were a function pointer.

tests/ui/ferrocene/lint-prevalidated/requires-validation.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    #[ferrocene::prevalidated]
+    #[ferrocene::requires_validation]
+    //~^ NOTE required to be validated here
+    fn requires_validation_yes_default(&self) -> u8 {
+        //~^ NOTE all implementations of
+        1
+    }
+}
```
jyn514, line 23: please also test associated constants with function pointers, with and without a default body.

tests/ui/ferrocene/lint-prevalidated/requires-validation.rs
```diff
@@ -0,0 +1,67 @@
...
+// Inheriting the default body needs no annotation: the default is validated.
+struct UsesDefault;
+
+impl Trait for UsesDefault {
+    fn requires_validation_no_default(&self) -> u8 {
+        //~^ ERROR unvalidated method implements a trait method that requires validation
+        //~| NOTE this implementation is unvalidated
+        //~| HELP add `#[ferrocene::prevalidated]`
+        6
+    }
```
jyn514, line 66: this is a duplicate of Bad::requires_validation_no_default above. I think it would be a better test if you *did* mark this with prevalidated, then make sure that no error is emitted for the inherited default for `yes_default`.

jyn514 (commented):
lots of little nits. the main changes i would like to see are:
- move attribute validation to `LintState::new`
- collapse similar branches so it's easier to tell what's happening. i think this is hard to follow right now *because* so many things are special cases, and a more general approach could actually be easier to understand.

happy to chat more if anything is unclear :)

compiler/rustc_lint/src/ferrocene/diagnostics.rs
Urhengulas: That's 100% true

compiler/rustc_lint/src/ferrocene/diagnostics.rs
Urhengulas: Yes

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: > this doesn't look right? why do you assume non-trait items are always valid?

The idea is that that `requires_validation` can only be applied to associated trait items. So if it is not an associated trait item we check if it has the attribute, which would be an error, otherwise we skip this item.

Urhengulas: > in general i think this is happening at the wrong time. there is existing code to check all this in LintState::new. i would expect this to be a modification there rather than completely new code. right now you're rechecking the same items over and over again, once per mono root.

I think I will need your help combining it. I will first finish the functionality as separate functions and then we can see how to combine it.

compiler/rustc_middle/src/middle/codegen_fn_attrs/ferrocene.rs
Urhengulas: I do need the span itself, so matching on the ValidatedStatus works. But I tried to just replace `any_parent_is_validated` with `item_is_validated`, but that results in the lint not detecting a simple method without a body anymore.

```diff
diff --git a/compiler/rustc_lint/src/ferrocene/mod.rs b/compiler/rustc_lint/src/ferrocene/mod.rs
index a1473851075..08d3c289c2b 100644
--- a/compiler/rustc_lint/src/ferrocene/mod.rs
+++ b/compiler/rustc_lint/src/ferrocene/mod.rs
@@ -182,7 +182,7 @@
 use rustc_hir::def::DefKind;
 use rustc_hir::{HirId, Item, TraitFn, TraitItem, TraitItemKind};
 use rustc_middle::middle::codegen_fn_attrs::ferrocene::{
-    ValidatedStatus, any_parent_is_validated, has_requires_validation_attribute, item_is_validated,
+    ValidatedStatus, has_requires_validation_attribute, item_is_validated,
 };
 use rustc_middle::span_bug;
 use rustc_middle::ty::{Instance, Ty, TyCtxt};
@@ -239,9 +239,9 @@ fn check_attribute_placement(
     def_id: LocalDefId,
     trait_item: Option<&TraitItem<'_>>,
 ) {
-    let prevalidated = match any_parent_is_validated(tcx, def_id.to_def_id()) {
+    let prevalidated = match item_is_validated(tcx, def_id.to_def_id()) {
         // FIXME: consider if validation is inherited
-        Some(ValidatedStatus::Validated { annotation, inherited: false }) => annotation,
+        ValidatedStatus::Validated { annotation, inherited: false } => annotation,
         _ => None,
     };
     let requires_validation = has_requires_validation_attribute(tcx, def_id.to_def_id());
diff --git a/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs b/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
index 02ce90df55c..9710caaeb13 100644
--- a/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
+++ b/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
@@ -7,17 +7,17 @@
 
 trait Trait {
     #[ferrocene::prevalidated]
-    //~^ ERROR `#[ferrocene::prevalidated]` cannot be applied to a method without a default body
-    //~| HELP use `#[ferrocene::requires_validation]`
     fn prevalidated_no_body(&self);
-    //~^ NOTE this method has no body to validate
 
     #[ferrocene::requires_validation]
     fn requires_validation_no_body(&self);
```

compiler/rustc_lint/src/ferrocene/mod.rs
Urhengulas: I do need the span itself, so matching on the ValidatedStatus works. But I tried to just replace `any_parent_is_validated` with `item_is_validated`, but that results in the lint not detecting a simple method without a body anymore.

```diff
diff --git a/compiler/rustc_lint/src/ferrocene/mod.rs b/compiler/rustc_lint/src/ferrocene/mod.rs
index a1473851075..08d3c289c2b 100644
--- a/compiler/rustc_lint/src/ferrocene/mod.rs
+++ b/compiler/rustc_lint/src/ferrocene/mod.rs
@@ -182,7 +182,7 @@
 use rustc_hir::def::DefKind;
 use rustc_hir::{HirId, Item, TraitFn, TraitItem, TraitItemKind};
 use rustc_middle::middle::codegen_fn_attrs::ferrocene::{
-    ValidatedStatus, any_parent_is_validated, has_requires_validation_attribute, item_is_validated,
+    ValidatedStatus, has_requires_validation_attribute, item_is_validated,
 };
 use rustc_middle::span_bug;
 use rustc_middle::ty::{Instance, Ty, TyCtxt};
@@ -239,9 +239,9 @@ fn check_attribute_placement(
     def_id: LocalDefId,
     trait_item: Option<&TraitItem<'_>>,
 ) {
-    let prevalidated = match any_parent_is_validated(tcx, def_id.to_def_id()) {
+    let prevalidated = match item_is_validated(tcx, def_id.to_def_id()) {
         // FIXME: consider if validation is inherited
-        Some(ValidatedStatus::Validated { annotation, inherited: false }) => annotation,
+        ValidatedStatus::Validated { annotation, inherited: false } => annotation,
         _ => None,
     };
     let requires_validation = has_requires_validation_attribute(tcx, def_id.to_def_id());
diff --git a/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs b/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
index 02ce90df55c..9710caaeb13 100644
--- a/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
+++ b/tests/ui/ferrocene/lint-prevalidated/requires-validation-placement.rs
@@ -7,17 +7,17 @@
 
 trait Trait {
     #[ferrocene::prevalidated]
-    //~^ ERROR `#[ferrocene::prevalidated]` cannot be applied to a method without a default body
-    //~| HELP use `#[ferrocene::requires_validation]`
     fn prevalidated_no_body(&self);
-    //~^ NOTE this method has no body to validate
 
     #[ferrocene::requires_validation]
     fn requires_validation_no_body(&self);
```

PR #2526, opened 2026-08-05
[1.97] Expanded coverage

Hoverbear:
Related to #2521.

Expands coverage for some try branches we weren't hitting.

PR #2546, opened 2026-08-20
[869f2wya4] show covered regions for partially tested lines

listochkin:
For each line of code I gather regions that overlap this line. A region can be a small expression or a larger block of code spanning several lines: the exact logic of what to consider a region is dictated by LLVM. 

If some of these regions have hit count of 0 then I consider the corresponding part of the line as uncovered. A line that has both covered and uncovered perts is a partially covered line. For these lines I use background color to show the covered portions.

Local testing of this change with @Hoverbear helped us find a few common cases of partially covered lines. One of popular ones were `?` at ends of lines for formatting code. And another were `as` type conversions. The later turned out to be a bug in our coverage metrics aggregation logic, and the fix comes in a separate PR.

### DIFF
--- a/ferrocene/tools/blanket/assets/html_report.css
+++ b/ferrocene/tools/blanket/assets/html_report.css
@@ -5,6 +5,7 @@
     --var-partial: #B3589A;
     --var-annotated: #BE8239;
     --var-ignored: gray;
+    --var-region-tested-bg: color-mix(in srgb, var(--var-tested) 35%, transparent);
     --var-untested-annotated: var(--var-untested);
     --var-partial-annotated: var(--var-partial);
     --var-annotated-text: none;
@@ -156,6 +157,11 @@ details.partial-annotation summary::after {
     padding-right: 0.5rem;
     color: white;
     background-color: var(--var-untested);
+}
+
+.functions .line-untested .region-tested {
+    background-color: var(--var-region-tested-bg);
+    border-radius: 2px;
 }
 
 .functions .line-ignored {
--- a/ferrocene/tools/blanket/src/html_report.rs
+++ b/ferrocene/tools/blanket/src/html_report.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 
 use maud::{DOCTYPE, PreEscaped};
 
-use crate::{Annotated, FunctionCoverage, FunctionCoverageStatus, LineCoverageStatus};
+use crate::{Annotated, ColumnRange, FunctionCoverage, FunctionCoverageStatus, LineCoverageStatus};
 
 const CSS: &str = include_str!("../assets/html_report.css");
 const JS: &str = include_str!("../assets/html_report.js");
@@ -217,7 +217,14 @@ fn generate_function(
                                     (line) "\n"
                                 },
                                 LineCoverageStatus::Untested => span class="line line-untested" data-filename=(filename) data-linenum=(linenum) {
-                                    (line) "\n"
+                                    @for (chunk, tested) in line_runs(line, function.regions.get(linenum)) {
+                                        @if tested {
+                                            span class="region-tested" { (chunk) }
+                                        } @else {
+                                            (chunk)
+                                        }
+                                    }
+                                    "\n"
                                 },
                                 LineCoverageStatus::Annotated => span class="line line-annotated" data-filename=(filename) data-linenum=(linenum) {
                                     (line) "\n"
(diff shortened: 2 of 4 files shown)


jyn514:
@listochkin what's the state of this PR? are you still working on it or are you looking for a review?

listochkin:
Added PR description, please review, @jyn514

ferrocene/tools/blanket/assets/html_report.css
```diff
@@ -5,6 +5,7 @@
     --var-partial: #B3589A;
     --var-annotated: #BE8239;
     --var-ignored: gray;
+    --var-region-tested-bg: color-mix(in srgb, var(--var-tested) 35%, transparent);
```
Hoverbear, line 8: This results in output like: 

Image

This is quite good, I wonder if showing the inverse would be more effective, though?

I experimented with this (ignore the whitespace) and I'm wondering if it communicates things better:

Image

ferrocene/tools/blanket/src/html_report.rs
```diff
@@ -234,3 +241,151 @@ fn generate_function(
...
+
+fn line_runs(line: &str, regions: Option<&Vec<ColumnRange>>) -> Vec<(String, bool)> {
+    let len = line.len();
+    let Some(regions) = regions.filter(|regions| !regions.is_empty() && len > 0) else {
+        return vec![(line.to_string(), false)];
+    };
+
+    let mut tested = vec![false; len];
+
+    let mut by_specificity: Vec<&ColumnRange> = regions.iter().collect();
```
Hoverbear, line 253: I think this particular bit of code earns a comment explaining why we work by specificity so that our future archeological inclined selves can remember.

Hoverbear (commented):
I'm happy with this and it seems to do what we need. Left a few comments.

Hoverbear:
Noting https://github.com/ferrocene/ferrocene/pull/2526 needs to get forward ported once this lands.

Hoverbear:
@listochkin does this PR mean https://github.com/ferrocene/ferrocene/pull/2521 can be closed? What's your intent here?

ferrocene/tools/blanket/assets/html_report.css
listochkin: Yes, I like this more, too. I'll make a change

listochkin:
Yes, I closed #2521

Hoverbear:
@listochkin How do you feel about like...

image
image
image



```patch
diff --git a/ferrocene/tools/blanket/assets/html_report.css b/ferrocene/tools/blanket/assets/html_report.css
index 03175b2f1ac..656838f8fb8 100644
--- a/ferrocene/tools/blanket/assets/html_report.css
+++ b/ferrocene/tools/blanket/assets/html_report.css
@@ -164,6 +164,10 @@ details.partial-annotation summary::after {
     border-radius: 2px;
 }
 
+.functions .line-untested .region-tested {
+    color: var(--var-tested);
+}
+
 .functions .line-ignored {
     color: var(--var-ignored);
     cursor: copy;
diff --git a/ferrocene/tools/blanket/src/html_report.rs b/ferrocene/tools/blanket/src/html_report.rs
index da52027e086..0e5e49aba13 100644
--- a/ferrocene/tools/blanket/src/html_report.rs
+++ b/ferrocene/tools/blanket/src/html_report.rs
@@ -218,10 +218,16 @@ fn generate_function(
                                 },
                                 LineCoverageStatus::Untested => span class="line line-untested" data-filename=(filename) data-linenum=(linenum) {
                                     @for (chunk, tested) in line_runs(line, function.regions.get(linenum)) {
+                                        @let (leading, content) = {
+                                            let content = chunk.trim_start();
+                                            let leading_len = chunk.len() - content.len();
+                                            chunk.split_at(leading_len)
+                                        };
+                                        (leading)
                                         @if tested {
-                                            (chunk)
+                                            span class="region-tested" { (content) }
                                         } @else {
-                                            span class="region-untested" { (chunk) }
+                                            span class="region-untested" { (content) }
                                         }
                                     }
                                     "\n"

```

ferrocene/tools/blanket/src/main.rs
```diff
@@ -253,25 +264,60 @@ fn get_coverage(
...
-        let status = match func_coverage.hits_for_line(line) {
-            None => LineCoverageStatus::Ignored,
-            Some(0) => LineCoverageStatus::Untested,
-            Some(_) => LineCoverageStatus::Tested,
+        let line_regions = column_ranges_for_line(func_coverage, line);
+        let status = if line_regions.is_empty() {
+            LineCoverageStatus::Ignored
+        } else if line_regions.iter().all(|range| range.tested) {
+            LineCoverageStatus::Tested
+        } else {
```
Hoverbear, line 287: (I left a ticket internally that suggests we might want to introduce `LineCoverageStatus::PartiallyTested` like we do with functions. This is fine for now.)

ferrocene/tools/blanket/src/main.rs
```diff
@@ -253,25 +264,60 @@ fn get_coverage(
...
+        source_name,
+        filename,
+        LineCoverage { lines: covered },
+        regions,
+        annotated,
+    ))
+}
+
+fn column_ranges_for_line(func_coverage: &CoverageResult, line: usize) -> Vec<ColumnRange> {
+    let mut ranges: Vec<ColumnRange> = func_coverage
```
Hoverbear, line 309: (Left a ticket that we might want to move to tracking regions as llvm-profparse does the primary result instead of lines, but that's ok as future work.)

Hoverbear (commented):
Suggested a small refinement, suggested a couple follow ups, this looks good otherwise. Excited to get this merged soon! :)

PR #2547, opened 2026-08-20
Improve blanket performance on v8

Hoverbear:
Some users who like blanket (not [Linus](https://peanuts.fandom.com/wiki/Linus%27_security_blanket)) have reported performance issues previously when using the full coverage report.

This was only present in Chrome/v8, not Firefox.

It was, in fact, quite bad: 
image
image

This patch does some naive optimizations that lower the cost:
image

I've tested this on:
* Firefox/Linux
* Chrome/Linux
* Edge/Windows
* Safari/Mac

### DIFF
--- a/ferrocene/tools/blanket/assets/html_report.js
+++ b/ferrocene/tools/blanket/assets/html_report.js
@@ -28,6 +28,7 @@ function main() {
     var annotatedIsTested = false;
 
     function onSearch() {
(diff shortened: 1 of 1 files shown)


jyn514 (approved):
feels a bit scary that we're handling this by looking at `.textContent` instead of structured data, but this seems like a strict improvement over the existing code

bors merge

Hoverbear:
bors merge

PR #2551, opened 2026-08-24
WIP: [869ec9jzq] Fix intrinsic-test

fs-rachel:


fs-rachel:
Current status: Paused until we update LLVM. We currently use LLVM 21, but upstream's CI explicitly disables running the intrinsic test for that LLVM version because it fails to compile.

PR #2553, opened 2026-08-25
[869ec9kc0] Do not try to generate coverage records for "always-const" instrinsics

fs-rachel:
These intrinsics should always be resolved at compile time, so should never exist at runtime. This means runtime coverage cannot possibly work for them. Trying to generate coverage data anyway trips a check which exists to make sure these intrinsics never reach codegen.

Previously we disabled the check when generating coverage; this patch instead disables coverage for these intrinsics, allowing the check to be reinstated.


jyn514 (approved):
this looks clearly correct 👍 thanks for catching this :)

bors merge

fs-rachel:
Network flake at the last moment :(

bors retry

PR #2554, opened 2026-08-26
[869eqbedu] Ensure that upstream pull automation always uses `pull.sh` from the target branch

fs-rachel:
Since I split some of the merge conflict resolution from `pull.sh` out into a separate script, the upstream pulls into our release branches have been broken. This turned out to be caused by the following sequence of events:

* The pull-upstream automation starts on the main branch, and runs `ferrocene/tools/pull-upstream/pull.sh` from that branch.

* During execution of `pull.sh`, it switches the repository to the appropriate release branch

* We're still running the version of `pull.sh` from the main branch, which eventually tries to run `ferrocene/tools/fix-merge/fix-merge.sh`

* This file doesn't exist any more, so things break

Fix this by inserting a wrapper script in between `automation.py` and `pull.sh`, which checks out the target branch first so that we always run the version of `pull.sh` from the target branch.

Add warnings to all relevant scripts to be extremely careful when making changes which could break things in the future.

### DIFF
--- a/ferrocene/tools/pull-upstream/pull.sh
+++ b/ferrocene/tools/pull-upstream/pull.sh
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: MIT OR Apache-2.0
 # SPDX-FileCopyrightText: The Ferrocene Developers
 
+# Note: Regardless of the target branch, the automation will always run
+# the version of this script from the `main` branch.
+
 set -euo pipefail
 IFS=$'\n\t'
 
@@ -17,6 +20,17 @@ X_HELP=src/etc/xhelp
 DIRECTORIES_CONTAINING_LOCKFILES=("" "src/bootstrap/")
 
 LOCAL_REPO_ROOT=$(git rev-parse --show-toplevel)
+
+# Record where we were before we start checking out new things.
+# This is used to re-checkout the original versions of this script and its dependencies,
+# to avoid ending up in a state where we're running pull.sh from one branch but its
+# dependencies from a different branch.
+#
+# When run via Github Actions, this will always be the tip of the main branch; when
+# run locally for testing purposes, it will be whatever commit you had checked out.
+INITIAL_COMMIT=$(git rev-parse HEAD)
+INITIAL_BRANCH=$(git branch --show-current)
+if [ -z "$INITIAL_BRANCH" ]; then INITIAL_BRANCH=$INITIAL_COMMIT; fi
 
 # Set a default max of merges per PR to 30, if it was not overridden in the
 # environment.
@@ -191,7 +205,25 @@ else
     merge_message="pull new changes from upstream"
 fi
 
+# Switch to the target branch, but keep this script and its dependencies
+# pinned to the versions we started with
 git checkout "${current_commit}"
+
+git checkout "${INITIAL_COMMIT}" -- ferrocene/tools/pull-upstream/ ferrocene/tools/fix-merge/
+
+# Sync submodules to the target branch so that the next step doesn't commit a bunch
+# of unintended submodule changes
+git submodule update --recursive
+
+# Commit the updated scripts into the target branch if they're different
+# Note that `git commit` exits with status code 1 if nothing has changed, and we need to suppress
+# that to stop it breaking the script. This could be done with an `|| true` on the git commit line,
+# but that looks weird and unintuitive, so it's nicer to turn off the `-e` flag for a moment.
+set +e
+git commit --quiet -F- <<EOF
+sync pull-upstream and fix-merge scripts from ${INITIAL_BRANCH}
+EOF
+set -e
 
 # - Unset some env vars to make git ignore user configuration. For example, if
 #   mergiraf is set as the merge driver globally, git would attempt to use it
--- a/ferrocene/tools/pull-upstream/pull.sh
+++ b/ferrocene/tools/pull-upstream/pull.sh
@@ -97,7 +97,7 @@ if [[ $# -ge 2 ]]; then
 else
     current_commit="$(git rev-parse HEAD)"
     branch_name="$(git branch --show-current)"
-    if [ -n "$branch_name" ]; then branch_name=$current_commit; fi
+    if [ -z "$branch_name" ]; then branch_name=$current_commit; fi
 fi
 if [[ $# -ge 3 ]]; then
     upstream_commit="$3"
--- a/ferrocene/tools/pull-upstream/automation.py
+++ b/ferrocene/tools/pull-upstream/automation.py
@@ -18,6 +18,9 @@
 # - `GITHUB_TOKEN`: API token with access to the repo contents, issues and RPs
 # - `GITHUB_REPOSITORY`: name of the GitHub repository to run this script on
 
+# Note: Regardless of the target branch, the automation will always run
+# the version of this script from the `main` branch.
+
 from automations_common import AutomatedPR, AutomationResult
 import generate_pr_body
 import os


ferrocene/tools/pull-upstream/wrapper.sh
```diff
@@ -0,0 +1,60 @@
...
+# Regardless of the target branch, the automation starts on the main branch. But at some point we
+# need to check out the target branch to be able to do things. Unfortunately, this can potentially
+# change all the automation scripts under our feet, which can lead to *extremely* weird breakages
+# if not handled carefully.
+#
+# This script deals with the handoff from the main branch to the target branch: The main-branch
+# version of this script is invoked, it checks out the target branch, then calls the target
+# branch's version of `pull.sh`.
+#
+# This could be handled by `automation.py`, but is kept separate to allow for local testing.
```
jyn514: could you please document how to do that local testing? either here or in the internal-procedures book?

jyn514 (changes requested):
the problem makes sense to me, but i don't think this is the right fix. my understanding is that this is trying to smooth over the difference between `pull.sh`-on-stable and `pull.sh`-on-main by copying the differences between them into a separate wrapper, is that correct? if so, i don't think we should do that.

what i suggest instead is what https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/backport/one.py already does: run `git checkout origin/main pull.sh fix-merge.sh` in the automation to make sure we're always using the latest version. then we only have to support one version in `automation.py`, and we don't have to introduce more indirection between python and bash.

fs-rachel:
> the problem makes sense to me, but i don't think this is the right fix. my understanding is that this is trying to smooth over the difference between `pull.sh`-on-stable and `pull.sh`-on-main by copying the differences between them into a separate wrapper, is that correct? if so, i don't think we should do that.
> 
> what i suggest instead is what https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/backport/one.py already does: run `git checkout origin/main pull.sh fix-merge.sh` in the automation to make sure we're always using the latest version. then we only have to support one version in `automation.py`, and we don't have to introduce more indirection between python and bash.

Okay, yes, I can do that instead. Will still add a note about the main-branch version always being used to some relevant places, as that's potentially non-obvious to people working on these files in the future

fs-rachel:
> Okay, yes, I can do that instead. Will still add a note about the main-branch version always being used to some relevant places, as that's potentially non-obvious to people working on these files in the future

The "nice" way to do this fails because re-checking-out the main branch's version of the pull-upstream scripts dirties the working tree, which causes `git merge` to bail out.

I can see a couple of plausible options for how to fix that; the least messy is to commit the updated scripts into the target branch. With this design we'll never actually run the pull-upstream script from the target branch, so that's *probably* fine. So that's what I've done for this rewrite.

ferrocene/tools/pull-upstream/wrapper.sh
fs-rachel: I've dropped the wrapper script in line with your other suggestions, so we're back to the local procedure just being "run pull.sh"

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -97,7 +111,7 @@ if [[ $# -ge 2 ]]; then
 else
     current_commit="$(git rev-parse HEAD)"
     branch_name="$(git branch --show-current)"
-    if [ -n "$branch_name" ]; then branch_name=$current_commit; fi
+    if [ -z "$branch_name" ]; then branch_name=$current_commit; fi
```
jyn514, line 114: oops! good catch

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -191,8 +205,15 @@ else
...
 fi
 
+# Switch to the target branch, but keep this script and its dependencies
+# pinned to the versions we started with
 git checkout "${current_commit}"
 
+git checkout "${INITIAL_COMMIT}" -- ferrocene/tools/pull-upstream/ ferrocene/tools/fix-merge/
+git commit --quiet -F- <<EOF || true # don't fail if the script hasn't changed
+sync pull-upstream and fix-merge scripts from ${INITIAL_BRANCH}
+EOF
```
jyn514: doesn't this commit end up in the final history?

jyn514 (commented):
> The "nice" way to do this fails because re-checking-out the main branch's version of the pull-upstream scripts dirties the working tree, which causes git merge to bail out.

this behavior surprises me, can you post the exact error you're seeing?

fs-rachel:
> > The "nice" way to do this fails because re-checking-out the main branch's version of the pull-upstream scripts dirties the working tree, which causes git merge to bail out.
> 
> this behavior surprises me, can you post the exact error you're seeing?

```
error: Your local changes to the following files would be overwritten by merge:
  ferrocene/tools/fix-merge/README.md ferrocene/tools/fix-merge/fix-merge.sh ferrocene/tools/pull-upstream/automation.py ferrocene/tools/pull-upstream/create-release-branch.sh ferrocene/tools/pull-upstream/pull.sh
Merge with strategy ort failed.
<lots of "automatically resolving conflict for [...] lines>
fatal: There is no merge in progress (MERGE_HEAD missing).
```

ferrocene/tools/pull-upstream/pull.sh
fs-rachel: Yes, I don't really see a great way around that

jyn514:
right ... but i think that only happens because both sides of the merge modify fix-merge.sh? if you look at https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/backport/one.py, it has a comment saying "you cannot backport a change to one.py itself"; i would expect the same here. upstream should never be changing our `ferrocene/` directory.

fs-rachel:
> right ... but i think that only happens because both sides of the merge modify fix-merge.sh? if you look at https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/backport/one.py, it has a comment saying "you cannot backport a change to one.py itself"; i would expect the same here. upstream should never be changing our `ferrocene/` directory.

Yeah, honestly I'm not sure why `git merge` complains here, but it does. Think I'll need to step through the process manually to see what's going on.

ferrocene/tools/pull-upstream/pull.sh
```diff
@@ -191,8 +205,15 @@ else
     merge_message="pull new changes from upstream"
 fi
 
+# Switch to the target branch, but keep this script and its dependencies
+# pinned to the versions we started with
 git checkout "${current_commit}"
 
+git checkout "${INITIAL_COMMIT}" -- ferrocene/tools/pull-upstream/ ferrocene/tools/fix-merge/
+git commit --quiet -F- <<EOF || true # don't fail if the script hasn't changed
+sync pull-upstream and fix-merge scripts from ${INITIAL_BRANCH}
```
Hoverbear, line 224: Is this intended to be a print or comment or something?

fs-rachel: The goal here is to add a commit if the checkout changed things, but to not fail the entire script if it didn't (as `git commit` exits with status 1 if nothing has changed). But heredoc syntax and its interaction with pipelines is *weird*, so... yeah, I should rewrite that to be clearer

fs-rachel: Is it clearer now?

fs-rachel:
> > right ... but i think that only happens because both sides of the merge modify fix-merge.sh? if you look at https://github.com/ferrocene/ferrocene/blob/main/ferrocene/tools/backport/one.py, it has a comment saying "you cannot backport a change to one.py itself"; i would expect the same here. upstream should never be changing our `ferrocene/` directory.
> 
> Yeah, honestly I'm not sure why `git merge` complains here, but it does. Think I'll need to step through the process manually to see what's going on.

@jyn514 It turns out that `git merge` is fine with modified files in the working tree, but it complains if they've been added to the index. Which `git checkout <commit> -- <files>` automatically does.

Thing is, we need to commit the re-synced scripts *anyway*, because otherwise they will get rolled into whatever commit the automation does next, which is just more confusing. And to avoid making this commit weird we need to re-sync the submodules :\

Which lead me to a thought: There is a cleaner way to do this, which avoids potential desync issues without having to commit the updated scripts to the target branch, which is to do everything on a separate worktree while running the scripts from the original (untouched) branch. That would be a much larger change, but would be a lot better IMHO.

ferrocene/tools/pull-upstream/pull.sh
Hoverbear: Much, thanks!

fs-rachel:
Implemented the separate-worktree idea as https://github.com/ferrocene/ferrocene/pull/2560 . We should merge *either* this PR *or* that one, not both

jyn514:
👍 approved https://github.com/ferrocene/ferrocene/pull/2560.

PR #2560, opened 2026-09-04
[869eqbedu] Upstream pull: Set up merge in a separate worktree

fs-rachel:
This allows us to run pull.sh and all of its dependencies from the main branch, without interference from the git shenanigans we need to do to build the merge commit for the upstream pull


jyn514 (approved):
bors merge

PR #2561, opened 2026-09-04
[869enbm57] Synchronize std::fs::set_permissions_nofollow with upstream

fs-rachel:
This function was essentially completely rewritten upstream by https://github.com/rust-lang/rust/pull/160170 .

The previous upstream version of this function was broken on Ubuntu 20.04, and had incorrect test expectations for QNX. So we had our own rewrite of the function to fix these issues. The fixes were rolled into the upstream rewrite, so drop what we had and just use the upstream version.

### DIFF
--- a/library/std/src/path.rs
+++ b/library/std/src/path.rs
@@ -2377,7 +2377,7 @@ pub struct NormalizeError;
 impl Path {
     // The following (private!) function allows construction of a path from a u8
     // slice, which is only safe when it is known to follow the OsStr encoding.
-    unsafe fn from_u8_slice(s: &[u8]) -> &Path {
+    pub(crate) unsafe fn from_u8_slice(s: &[u8]) -> &Path {
         unsafe { Path::new(OsStr::from_encoded_bytes_unchecked(s)) }
     }
     // The following (private!) function reveals the byte encoding used for OsStr.
--- a/library/std/src/sys/fs/unix.rs
+++ b/library/std/src/sys/fs/unix.rs
@@ -1866,61 +1866,90 @@ pub fn set_perm(p: &CStr, perm: FilePermissions) -> io::Result<()> {
     cvt_r(|| unsafe { libc::chmod(p.as_ptr(), perm.mode) }).map(|_| ())
 }
 
+#[cfg(target_os = "android")]
+pub fn set_perm_nofollow(_p: &CStr, _perm: FilePermissions) -> io::Result<()> {
+    // Currently Android seems to be having inconsistent behavior with fchmodat
+    // with `AT_SYMLINK_NOFOLLOW` or openat with `O_NOFOLLOW` + fchmod.
+    // See this issue here mentioning inconsistent behavior on fchmodat:
+    // https://github.com/android/ndk/issues/1258
+    // On the arm-android CI job, using fchmodat with `AT_SYMLINK_NOFOLLOW` +
+    // fallback behavior on a symlink sets the target file's permissions,
+    // which is incorrect behavior.
+    Err(crate::io::ErrorKind::Unsupported.into())
+}
+
+#[cfg(not(target_os = "android"))]
 pub fn set_perm_nofollow(p: &CStr, perm: FilePermissions) -> io::Result<()> {
-    // ESP-IDF and Horizon do not support O_NOFOLLOW, so we skip setting it.
-    // Their filesystems do not have symbolic links, so no special handling is required.
-    cfg_select! {
-        // wasm32-wasip1 targets do not support fchmodat, so we fall down to
-        // open + fchmod
-        target_os = "wasi" => {
-            use crate::fs::OpenOptions;
-            use crate::fs::Permissions;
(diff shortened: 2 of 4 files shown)


fs-rachel:
Test run (minus Windows and Mac) succeeded, so this is ready for review

library/std/src/path.rs
```diff
@@ -2377,7 +2377,7 @@ pub struct NormalizeError;
 impl Path {
     // The following (private!) function allows construction of a path from a u8
     // slice, which is only safe when it is known to follow the OsStr encoding.
-    unsafe fn from_u8_slice(s: &[u8]) -> &Path {
+    pub(crate) unsafe fn from_u8_slice(s: &[u8]) -> &Path {
```
Urhengulas, line 2380: Can you please document this is a change by us?


```suggestion
    // Ferrocene modification: Change visibility from private to `pub(crate)`
    pub(crate) unsafe fn from_u8_slice(s: &[u8]) -> &Path {
```

library/std/src/sys/fs/unix.rs
```diff
@@ -1866,61 +1866,90 @@ pub fn set_perm(p: &CStr, perm: FilePermissions) -> io::Result<()> {
...
+    // with `AT_SYMLINK_NOFOLLOW` or openat with `O_NOFOLLOW` + fchmod.
+    // See this issue here mentioning inconsistent behavior on fchmodat:
+    // https://github.com/android/ndk/issues/1258
+    // On the arm-android CI job, using fchmodat with `AT_SYMLINK_NOFOLLOW` +
+    // fallback behavior on a symlink sets the target file's permissions,
+    // which is incorrect behavior.
+    Err(crate::io::ErrorKind::Unsupported.into())
+}
+
+#[cfg(not(target_os = "android"))]
```
Urhengulas, line 1881: In upstream it is `#[cfg(not(any(target_os = "android", target_os = "vxworks")))]` ([see](https://github.com/rust-lang/rust/blob/745de6eca673de5329ec68f2689629a5ca45ab35/library/std/src/sys/fs/unix.rs#L1906)). Why don't we?

Urhengulas, line 1881: Also upstream has a `#[cfg(target_os = "vxworks")]` version ([see](https://github.com/rust-lang/rust/blob/745de6eca673de5329ec68f2689629a5ca45ab35/library/std/src/sys/fs/unix.rs#L1887)). Why don't we?

library/std/src/sys/fs/unix.rs
```diff
@@ -1866,61 +1866,90 @@ pub fn set_perm(p: &CStr, perm: FilePermissions) -> io::Result<()> {
...
+        // Their filesystems do not have symbolic links, so no special handling is required.
+        #[cfg(not(any(target_os = "espidf", target_os = "horizon")))]
+        {
+            #[cfg(not(target_os = "wasi"))]
             use crate::os::unix::fs::OpenOptionsExt;
-
-            let mut options = OpenOptions::new();
+            #[cfg(target_os = "wasi")]
+            use crate::os::wasi::fs::OpenOptionsExt;
             options.read(true).custom_flags(libc::O_NOFOLLOW);
```
Urhengulas, line 1898: Upstream unconditionally does `options.read` just after creating the `OpenOptions` ([see](https://github.com/rust-lang/rust/blob/745de6eca673de5329ec68f2689629a5ca45ab35/library/std/src/sys/fs/unix.rs#L1914)). Why don't we?

Urhengulas (changes requested):
I found it a bit hard to review since I don't know which are the Ferrocene specific changes and which changes are to match upstream. But I pulled up the upstream code and compared it.

fs-rachel:
Huh, none of that is deliberate. I swear I diff'd the files against upstream after resolving the merge conflicts...

Let me go fix all of that

fs-rachel:
ahhhh I see. All of the things you found are upstream changes from *after* the commit I synced against. I synced from the final commit of the upstream PR: https://github.com/rust-lang/rust/pull/160170/changes/b901613fcab6b1e784be3af9f5f1e84888a0569d

fs-rachel:
Also

> I found it a bit hard to review since I don't know which are the Ferrocene specific changes and which changes are to match upstream.

is expected - this PR is to replace our temporary hack with the upstream fix, so that these problems don't show up in the middle of an already-complicated upstream pull. Doesn't overly matter which commit we sync to (aside from that we need to agree on which one ;) ), just that we synchronize to *some* upstream commit; any later changes will be pulled in by an upstream pull at some point, and should apply relatively cleanly.

And yes, comparing the files against the appropriate upstream commit is the way to go for reviewing this

Urhengulas (approved):
> ahhhh I see. All of the things you found are upstream changes from _after_ the commit I synced against. I synced from the final commit of the upstream PR: [rust-lang/rust@b901613](https://github.com/rust-lang/rust/commit/b901613fcab6b1e784be3af9f5f1e84888a0569d)

That makes sense! Good by me then

Urhengulas:
bors merge

PR #2565, opened 2026-09-08
[869ec9kc0] Sync logic to avoid generating coverage for compile-time-only functions with upstream

fs-rachel:
In https://github.com/ferrocene/ferrocene/pull/2553 I marked compile-time-only functions as ineligible for coverage. I sent this upstream, and after a couple of further iterations it was merged as https://github.com/rust-lang/rust/pull/162354 .

Copy the final version of this logic back to Ferrocene.


Urhengulas (approved):
bors merge

PR #2566, opened 2026-09-08
[869eqbedu] Remove template argument from `mktemp` in pull-upstream script

fs-rachel:
This has completely different semantics on Mac vs. Linux, so there's no good way to use it. Removing this just means that the temporary directory gets a generic name, instead of one which indicates which script created it.


jyn514 (approved):
bors merge

Rachel has tested this manually on Mac and Ana has tested on Linux.

PR #2569, opened 2026-09-09
Use monomorphized function names to track coverage

listochkin:
When a macro is used to generate implementations of the same function for different types LLVM generates separate counters for different implementations. For example, in `core/src/num` a macro `uint_impl!` is used for generating versions of `abs_diff` function for `u8`, `u16`, etc. When the tests run for one of the types, say, `u8`, the counters for a corresponding compiled function `abs_diff` get updated.

However, since Blanket / `llvm_profparser::CoverageResult` tracked coverage by source line-column it does not disambiguate counters for different versions of the same function. When aggregating counters for a specific line or line fragment if an unrelated version of the function was not covered we would get 0 in one of the counters.

Also, while the source for all these versions is the same, when compiling LLVM can use different transformations for different paths. The resulting shape of functions can be different, and the borders of coverage regions don't always match!

To disambiguate counters correctly we need to track coverage over monomorphized versions of functions. I use their mangled names for them, referred as `linkage_name` in code. This gets slightly complicated because the mangled names of functions have randomized segments that get different values between compilation runs. When a tests binary is compiled, Cargo can decide to reuse a version of binary under test from a previous compilation run. For example, if we run `cargo build` followed by `cargo test` the later will reuse the results of the former. Unfortunately, metadata in prof files will have both versions of a mangle name present. For that reason I clean up random segments from mangled names to unify counters.

*Technically*, a truly generic function may not have exact implementations and monomorphized versions. In that case I keep the value for `linkage_name` as empty string. And in theory counters for these functions will still be merged. I'm not sure whether these kinds of functions are actually executable. My gut tells me that *at some point* there should be a monomorphized version of that function that gets called inside tests, and so far this has been the case.

blocked by https://github.com/ferrocene/ferrocene/pull/2546

### DIFF
--- a/ferrocene/tools/Cargo.lock
+++ b/ferrocene/tools/Cargo.lock
@@ -164,21 +164,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "assert_cmd"
-version = "2.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "39bae1d3fa576f7c6519514180a72559268dd7d1fe104070956cb687bc6673bd"
-dependencies = [
- "anstyle",
- "bstr",
- "libc",
- "predicates",
- "predicates-core",
- "predicates-tree",
- "wait-timeout",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


Hoverbear:
@listochkin I don't think vendoring llvm-profparser so we can patch it is my first choice -- Is this something we can upstream to the project?

PR #2577, opened 2026-09-17
[869ezvkep] Make PR title check script more verbose

fs-rachel:
Print inputs, and explicitly print the note about automated branches being exempt from the check.
    
This would have caught my confusion between GITHUB_REF and GITHUB_HEAD_REF much earlier.
    
For future reference, GITHUB_REF can point to various things depending on context. It *can* be the (fully qualified, ie. including "refs/heads/") name of the source branch, but not always. Meanwhile GITHUB_HEAD_REF is always the (unqualified) name of the PR source branch, which is what we want.


tshepang (approved):
bors merge

fs-rachel:
Weird failure which seems unrelated to this patch, so...

bors retry

tshepang:
bors merge

